PR #1, opened 2026-05-29
chore(config): 把 bundle id / team id 从 pbxproj 外部化为 xcconfig 变量

DeveKing7:
## 背景

开源后,`OctoiOS.xcodeproj/project.pbxproj` 里硬编码的私有 **bundle identifier** 与 **DEVELOPMENT_TEAM** 会随仓库泄露给所有人。本 PR 把这两类部署者特有信息外部化为占位变量,真实值只保留在 gitignored 的 `OctoConfig.xcconfig`,入库文件仅含 `$(OCTO_*)` 占位。

## 改动

- `project.pbxproj`
  - `PRODUCT_BUNDLE_IDENTIFIER` → `"$(OCTO_BUNDLE_ID)"`;三个扩展派生为 `.ShareExtension` / `.NotificationService` / `.NotificationContent`
  - `DEVELOPMENT_TEAM` → `"$(APPLE_TEAM_ID)"`
- `OctoConfig.xcconfig.template`
  - 新增 `OCTO_BUNDLE_ID` 占位段及说明(默认 `com.example-app.octo`)

## 变量解析链路

沿用现有 `Podfile` post_install 注入机制,无需新增逻辑:
- **主 App**:走 Pods 聚合 xcconfig 的 `#include? "../../../OctoConfig.xcconfig"`
- **三个扩展**:走 `baseConfigurationReference` 直接绑定 `OctoConfig.xcconfig`

## 验证

`xcodebuild -showBuildSettings` 实测四个 target 均能从占位变量还原真实包名:

- 主 App → `com.<your-org>.octo`
- ShareExtension → `com.<your-org>.octo.ShareExtension`
- `DEVELOPMENT_TEAM` 正常还原

且修复了变量引用需加引号的问题(裸 `$(...)` 会让旧式 plist 解析器报 "project damaged")。

## 说明

- 部署者需在本地 `OctoConfig.xcconfig` 配置 `OCTO_BUNDLE_ID`(参见 template)
- 缺失时 `#include?` 软引用静默失败,占位变量保持空 —— 不会编译崩溃,仅签名需手动指定

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6, opened 2026-05-30
fix(ci): finalize PR notification design (octo-pr-result-notify + delete octo-pr-feed)

### LINKED ISSUE #9: fix(ci): finalize PR notification design — remove octo-pr-feed, fix result-notify
## Problem

The current `octo-pr-result-notify.yml` workflow was added in a prior pass with an outdated design:
- Trigger includes `reopened` (non-terminal state)
- No `feed_group_id` using org-level variable
- A separate `octo-pr-feed.yml` was also added which conflicts with the intended design

## Solution (this PR: #6)

Finalize the PR notification design per product owner direction:
- `octo-pr-result-notify.yml`: terminal state only (`[closed]`), routes to global pr-feed via `vars.OCTO_PR_FEED_GROUP_ID`
- Delete `octo-pr-feed.yml` (not needed)

## Workflow Ownership (final design)

| Workflow | Trigger | Destination |
|----------|---------|-------------|
| `octo-pr-result-notify.yml` | `[closed]` + review submitted | global pr-feed (org variable) |
| `octo-pr-review-feed.yml` | `[ready_for_review, review_requested]` | project group |

lml2468:
Closes #9

## Summary

Updates `octo-pr-result-notify.yml` to align with the finalized PR notification design.

| Workflow | Trigger | Sends to |
|----------|---------|----------|
| `octo-pr-result-notify.yml` | `pull_request_target: [closed]` + `pull_request_review: [submitted]` | global pr-feed (`vars.OCTO_PR_FEED_GROUP_ID` with hardcoded fallback) |
| `octo-pr-review-feed.yml` | `pull_request_target: [ready_for_review, review_requested]` | project group (repo-specific) |

### Design decisions (per product owner direction)

- **`opened`/`reopened` events are intentionally not monitored** — only terminal states (merged/closed) and review outcomes are relevant
- All notification jobs route to the global pr-feed group via org-level variable `OCTO_PR_FEED_GROUP_ID`

### Changes in this PR

- **`octo-pr-result-notify.yml`**:
  - Trigger narrowed from `[closed, reopened]` → `[closed]` only
  - `feed_group_id` now uses `${{ vars.OCTO_PR_FEED_GROUP_ID || '1c303c142e9840f2a9b46c10b0972e8d' }}` — org-level variable with safe fallback to prevent empty-string override of the reusable workflow's built-in default
  - `event_kind` expression simplified to `pr_merged | pr_closed` (removed `pr_reopened` case)

No application code modified.


yujiawei (dismissed):
# Code Review — PR #6 (octo-ios)

Thin caller workflow that wires this repo into the org's reusable `octo-pr-feed.yml`. The change is small, consistent with the existing notification callers in this repo (`octo-pr-result-notify.yml`, `octo-issue-feed.yml`), and the security posture is sound. No blocking issues — approving with a couple of non-blocking observations.

## What's good

- **`pull_request_target` is safe here.** The `zizmor: ignore[dangerous-triggers]` is justified: `permissions: {}` drops the default token to nothing, no PR head code is checked out or executed, and the reusable only reads event metadata. This matches the pattern already used by `octo-pr-result-notify.yml`.
- **Untrusted input is sanitized downstream.** The reusable strips C0/DEL control chars from `pr_title` and `pr_author` (IM-injection guard), validates group IDs (32-char hex), validates `repo_name` against path traversal, and allowlists the API base URL. Passing raw `pull_request.title`/`user.login` from the caller is therefore fine.
- **Draft guard + retry/backoff** in the reusable are reasonable (`if: !draft`, exponential backoff honoring `Retry-After`).

## Observations (non-blocking)

### 1. Duplicate notifications to the project group on `closed`/`reopened` (suggestion)

The reusable `octo-pr-feed.yml` unconditionally sends to **both** `feed_group_id` (default global pr-feed `1c303…`) **and** `project_group_id`. This caller sets `project_group_id: '40dd577b7e55490ab278301065f431a3'`.

That same group `40dd…` is already notified on `closed`/`reopened` by the existing `octo-pr-result-notify.yml` (`feed_group_id: "40dd577b7e55490ab278301065f431a3"`), which carries richer merge/review context. So once this PR lands, the project group will receive **two** messages per `closed`/`reopened` event — the detailed one from result-notify plus a thinner one from pr-feed.

The PR description anticipates this ("`octo-pr-result-notify.yml` continues to handle these for the project group"), but because the reusable always dual-sends, the caller can't actually opt out of hitting the project group. `opened` is the genuinely-new event with no overlap.

Two ways to think about it:
- If the goal is *only* to fill the global pr-feed channel and avoid project-group dupes, consider narrowing this caller to `types: [opened]` (the one event no other workflow covers for the project group). The trade-off: the global feed would then miss close/reopen lifecycle events.
- If full lifecycle in the global feed is the priority (as the title suggests), the current `[opened, closed, reopened]` is correct, and the project-group dup is the accepted cost — in which case it'd be worth a follow-up to let the reusable send to the feed channel only (skip `project_group_id`) so callers can avoid the overlap.

Either is fine; just flagging that the dup is structural, not accidental.

### 2. Draft → ready-for-review PRs never appear in the feed (nit)

A PR opened as a draft is skipped by `if: !draft`, and `ready_for_review` is not in `types`. So a PR that starts as draft and is later marked ready will never produce an `opened`-style feed entry. The reusable's emoji map already handles `ready_for_review`, so adding it to `types` (with the draft guard naturally allowing it through) would close this gap if draft workflows are common in this repo.

### 3. Reusable pinned to `@main` (nit)

`uses: …/octo-pr-feed.yml@main` is a mutable ref. This is consistent with every other caller in the repo, so it's an org convention rather than a regression — noting only as a general hardening option (SHA-pin) should the org ever tighten supply-chain policy.

## Verdict

No P0/P1. Correct, well-scoped, and consistent with existing patterns. Approving. The duplicate-notification point (#1) is worth a conscious decision but does not block merge.

Jerry-Xin (dismissed):
Summary: The PR is relevant to this repository and the workflow changes are thin CI notification wiring with no blocking issues found.

💬 Non-blocking
- 🟡 Warning: `.github/workflows/octo-pr-result-notify.yml:3` removes the previous `pull_request_target` handling for `closed` and `reopened`, while the PR description says that workflow continues to handle those events for the project group. The new `.github/workflows/octo-pr-feed.yml:13` reusable does send to both the global feed and `project_group_id`, so this is not blocking, but the PR description should be corrected or the intended ownership clarified.
- 🔵 Suggestion: `.github/workflows/octo-pr-feed.yml:13` pins the reusable workflow to `@main`, matching existing local patterns. For stronger supply-chain control, consider pinning reusable workflows to a trusted tag or SHA in a future hardening pass.

✅ Highlights
- Project relevance gate passed: CI notification workflows are in scope for `Mininglamp-OSS/octo-ios`.
- `.github/workflows/octo-pr-feed.yml:8` keeps `permissions: {}`, and the reusable workflow does not execute PR code, which keeps the `pull_request_target` usage appropriately constrained.
- The caller inputs match the current reusable `Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml` contract, including the default global pr-feed channel and the explicit project group id.

yujiawei (dismissed):
# Code Review — PR #6 (octo-ios)

## Verdict: APPROVED

A clean, low-risk CI change that adds the final missing caller workflow (`octo-pr-feed.yml`) and removes the now-redundant `pull_request_target` job from `octo-pr-result-notify.yml`. No application code touched. No P0/P1 issues found.

## 1. Verification

- ✅ **`pull_request_target` usage is safe here.** The new caller passes only event *metadata* to the reusable workflow and never checks out or executes PR-supplied code. `permissions: {}` removes all default token scopes, and the reusable (`Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@main`) sanitizes `pr_title`/`pr_author` (strips C0 control chars) and enforces an API base-URL allowlist before using `OCTO_BOT_TOKEN`. The `zizmor: ignore[dangerous-triggers]` annotation with a justification is appropriate. This matches the established pattern in `octo-pr-review-feed.yml` in the same repo.
- ✅ **No notification regression.** Previously `octo-pr-result-notify.yml` handled `closed`/`reopened` via `pull_request_target` and routed them to the project group `40dd…f431a3` (passed as `feed_group_id`). After this change, `octo-pr-feed.yml` handles `opened`/`closed`/`reopened` and routes to the **same** project group via `project_group_id: '40dd577b7e55490ab278301065f431a3'`, plus the default global pr-feed channel (`1c303…`). The old job is removed in the same PR, so there is **no duplicate** delivery and no loss of project-group coverage.
- ✅ **Input wiring is correct.** All required reusable inputs (`repo_name`, `pr_number`, `pr_title`, `pr_url`, `pr_author`, `event_action`, `project_group_id`) are supplied; optional stats/`pr_merged` are provided. `OCTO_BOT_TOKEN` is forwarded via `secrets:`.
- ✅ **`octo-pr-result-notify.yml` after the edit** retains a coherent single `review-result` job triggered on `pull_request_review`, with the same-repo guard (`head.repo.full_name == github.repository`) intact.

## 2. Findings

### P2 — Minor behavior change: `branches: [main]` narrows close/reopen scope
`octo-pr-feed.yml` adds `branches: [main]`, but the removed `pull_request_target` job in `octo-pr-result-notify.yml` had **no** branch filter. As a result, PRs targeting a base branch other than `main` will no longer produce `closed`/`reopened` notifications to the project group. This is almost certainly fine in practice (PRs overwhelmingly target `main`), but it is a behavior change relative to the code being removed — worth a one-line confirmation that it's intentional.

### P2 (nit) — Reusable referenced by mutable `@main`
`uses: …/octo-pr-feed.yml@main` is a mutable ref; pinning to a commit SHA would make the supply chain reproducible/tamper-resistant. Noting this only as an observation: it is **consistent with every other caller in this repo** (`octo-pr-review-feed.yml`, `octo-pr-result-notify.yml`), so changing it here alone would be inconsistent. If SHA-pinning is desired it should be an org-wide sweep, not a blocker for this PR.

## 3. Suggestions
- Confirm the `branches: [main]` restriction for `closed`/`reopened` is intended (vs. the previous all-branches behavior).

## 4. Additional Notes
- `if: ${{ !github.event.pull_request.draft }}` correctly skips draft PRs, matching the other feed workflows.
- No new secret exposure: `OCTO_BOT_TOKEN` is passed through `secrets:` only.

Approving — the change is correct, security-hygienic, and consistent with the repo's existing feed workflows. The P2 items are non-blocking.

Jerry-Xin (changes requested):
This PR is relevant to the repository because it changes the project’s GitHub notification workflow, but it currently breaks the workflow YAML.

🔴 Blocking

- 🔴 Critical — `.github/workflows/octo-pr-result-notify.yml:26` and `.github/workflows/octo-pr-result-notify.yml:43`: `OCTO_BOT_TOKEN: *** secrets.OCTO_BOT_TOKEN }}` is invalid YAML and not a valid GitHub Actions expression. The leading `*` is parsed as a YAML alias and the file fails to parse, so this workflow cannot load or run. Restore the expression as:
  ```yaml
  OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
  ```

💬 Non-blocking

- 🔵 Suggestion — `.github/workflows/octo-pr-result-notify.yml:21`: The simplified `event_kind` expression is consistent with `pull_request_target.types: [closed]`. No issue, but this relies on keeping the trigger limited to `closed`.

✅ Highlights

- The removal of `reopened` from the result notification trigger matches the described ownership split.
- Removing the explicit `feed_group_id` aligns with using the reusable workflow’s default global PR feed, assuming that default is configured in `Mininglamp-OSS/.github`.

yujiawei (changes requested):
# Code Review — PR #6 (octo-ios)

## Summary

The intent of this PR is reasonable: scope `octo-pr-result-notify.yml` to terminal/review events, route to the global pr-feed by dropping the explicit `feed_group_id` override, and delete the now-redundant `octo-pr-feed.yml`. Those three design changes are individually sound.

However, the diff also introduces a **blocking syntax corruption** in the secrets passthrough that breaks the entire workflow file. This must be fixed before merge.

## Findings

### P0 — Corrupted `secrets` expression breaks the workflow (build-break)

`.github/workflows/octo-pr-result-notify.yml` lines 26 and 43 now read:

```yaml
      OCTO_BOT_TOKEN: *** secrets.OCTO_BOT_TOKEN }}
```

The opening `${{` was replaced by `***` (this looks like a GitHub secret-masking artifact that got committed into the source). The base branch had the correct form:

```yaml
      OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
```

Impact:
- This is not valid YAML. A scalar beginning with `*` is parsed as a YAML **alias**, and `***` is not a valid alias name, so the file fails to parse entirely (`while scanning an alias … expected alphabetic or numeric character, but found '*'`).
- GitHub Actions will reject the whole workflow, so **neither** `pr-result` **nor** `review-result` will run — the notification feature this PR is finalizing is fully disabled.
- Brace accounting confirms it: the file has 16 `${{` opens but 18 `}}` closes; the two orphan closes are exactly these two token lines.

**Required fix** — restore both lines to:

```yaml
    secrets:
      OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
```

After fixing, please confirm the file passes `actionlint` (or a YAML parse) locally, since this class of error is caught immediately by either.

## Verified OK

- ✅ `on.pull_request_target.types` narrowed to `[closed]` — consistent with the stated "terminal state" design; merged-vs-closed is correctly derived below.
- ✅ `event_kind` simplification `github.event.pull_request.merged == true && 'pr_merged' || 'pr_closed'` is correct now that `reopened` is no longer a trigger (the old ternary's `pr_reopened` branch is correctly dropped).
- ✅ Removal of the hardcoded `feed_group_id: "40dd577b7e55490ab278301065f431a3"` from both jobs — matches the goal of routing to the global default pr-feed.
- ✅ `octo-pr-feed.yml` deletion — confirmed absent at the head SHA; no remaining references to it in the workflows dir.
- ✅ `permissions: {}` retained (least privilege).

## Verdict

Requesting changes: a single P0 (corrupted `${{ → ***` in the `secrets` block) makes the workflow unparseable. The rest of the diff is good and can merge as-is once the two token lines are restored.

.github/workflows/octo-pr-result-notify.yml
```diff
@@ -18,13 +18,12 @@ jobs:
...
       pr_author: ${{ github.event.pull_request.user.login }}
-      event_kind: ${{ github.event.action == 'closed' && github.event.pull_request.merged == true && 'pr_merged' || github.event.action == 'closed' && 'pr_closed' || 'pr_reopened' }}
+      event_kind: ${{ github.event.pull_request.merged == true && 'pr_merged' || 'pr_closed' }}
       pr_additions: ${{ github.event.pull_request.additions }}
       pr_deletions: ${{ github.event.pull_request.deletions }}
       pr_changed_files: ${{ github.event.pull_request.changed_files }}
-      feed_group_id: "40dd577b7e55490ab278301065f431a3"
     secrets:
-      OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+      OCTO_BOT_TOKEN: *** secrets.OCTO_BOT_TOKEN }}
```
yujiawei: **P0 build-break:** `${{` was replaced by `***` here (a secret-masking artifact committed into source). This is invalid YAML — a scalar starting with `*` is parsed as a YAML alias and `***` fails to scan, so the entire workflow file is rejected and neither job runs. Restore to `OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}`. Same corruption on line 43.

Jerry-Xin (dismissed):
Summary: This PR is in scope for `Mininglamp-OSS/octo-ios` because it modifies repository CI/notification workflow behavior, and I found no blocking issues.

✅ Highlights
- 🔵 Suggestion: `.github/workflows/octo-pr-result-notify.yml:4-5` now only emits terminal PR result notifications on `closed`, which matches the stated ownership split.
- 🔵 Suggestion: `.github/workflows/octo-pr-result-notify.yml:21` correctly simplifies `event_kind` now that the `reopened` trigger is removed.
- 🔵 Suggestion: `.github/workflows/octo-pr-result-notify.yml:25` and `.github/workflows/octo-pr-result-notify.yml:42` rely on the reusable workflow’s default `feed_group_id`, which matches the intended global `pr-feed` routing.
- Security posture remains reasonable: the `pull_request_target` workflow does not check out or execute PR code, and permissions stay empty.

yujiawei (dismissed):
# Code Review — PR #6 (octo-ios)

## Verdict: APPROVED

CI-only change to the PR notification design. No application code touched. The diff is small (+2/-4 in a single workflow file) and the resulting state is correct.

## Verification

✅ **Trigger simplification is sound** (`.github/workflows/octo-pr-result-notify.yml:5`)
`pull_request_target.types` narrowed from `[closed, reopened]` to `[closed]`. Reopened PRs no longer emit a result notification, which matches the stated intent of only notifying on the terminal state. The `pr-result` job still gates on `github.event_name == 'pull_request_target'`, so the narrower trigger is internally consistent.

✅ **`event_kind` expression correctly collapsed** (`octo-pr-result-notify.yml:21`)
Old expression handled three branches (`pr_merged` / `pr_closed` / `pr_reopened`). With `reopened` removed from the trigger, `github.event.action` is always `closed`, so the simplification to `merged == true && 'pr_merged' || 'pr_closed'` is logically equivalent for the remaining event surface — no dead branch, no missing case.

✅ **`feed_group_id` removal is safe** (both jobs)
Confirmed against the reusable workflow `Mininglamp-OSS/.github/.github/workflows/octo-pr-result-notify.yml@main`: `feed_group_id` is `required: false` with `default: '1c303c142e9840f2a9b46c10b0972e8d'`. Dropping the explicit `40dd577b...` override falls back to that global default, which matches the design goal of routing result/review notifications to the global pr-feed while `octo-pr-review-feed.yml` retains the project group `40dd577b...`. No required input is left unset.

✅ **No trigger overlap / double-notification**
`octo-pr-result-notify.yml` fires on `closed` + `pull_request_review[submitted]`; `octo-pr-review-feed.yml` fires on `[ready_for_review, review_requested]`. The event sets are disjoint, so ownership is clean with no duplicate sends.

✅ **YAML parses cleanly** and the `${{ secrets.OCTO_BOT_TOKEN }}` expression on both jobs is well-formed (the head commit fixed a previously malformed token expression).

## Notes (non-blocking)

- **P2 / cosmetic:** The PR title and body describe deleting `octo-pr-feed.yml`, but that file is not present on `main` and is not part of this PR's diff (the GitHub diff shows only `octo-pr-result-notify.yml`). The deletion appears to have already landed earlier. The end state is correct; the description just overstates this PR's scope. Worth tightening the body for an accurate audit trail, but it does not affect behavior.

No P0/P1 issues. Approving.

Jerry-Xin (changes requested):
The PR is in scope, but the new `feed_group_id` wiring can break the intended default notification path.

🔴 Blocking

- 🔴 Critical: `.github/workflows/octo-pr-result-notify.yml:25` and `.github/workflows/octo-pr-result-notify.yml:43` now always pass `feed_group_id: ${{ vars.OCTO_PR_FEED_GROUP_ID }}`. If that repository/org variable is unset, GitHub passes an empty string to the reusable workflow, which prevents the reusable workflow’s own default `feed_group_id` from being used and causes its `FEED_GROUP_ID` validation to fail. This contradicts the PR description’s “global pr-feed (default)” design. If the desired behavior is the reusable default, omit `feed_group_id` entirely; if configurability is required, add an explicit fallback or document/enforce the required variable.

💬 Non-blocking

- 🔵 Suggestion: `.github/workflows/octo-pr-result-notify.yml:21` is now correct only because the trigger is restricted to `types: [closed]`. That is fine, but future trigger changes would make this expression easier to misuse. Keeping the event-kind expression tied to the action, or adding a short comment, would reduce regression risk.

✅ Highlights

- The PR passes the project relevance gate: it updates this repository’s GitHub Actions notification ownership.
- Removing `reopened` from the result notification trigger aligns with the stated terminal-state responsibility split.
- No application code or runtime iOS behavior is affected.

yujiawei (changes requested):
# Code Review — PR #6 (octo-ios)

## Summary

This PR makes a 4-line change to `.github/workflows/octo-pr-result-notify.yml`:

1. Narrows the `pull_request_target` trigger from `[closed, reopened]` to `[closed]`, and simplifies the `event_kind` expression from a 3-way (`pr_merged`/`pr_closed`/`pr_reopened`) to a 2-way (`pr_merged`/`pr_closed`).
2. Replaces the hardcoded `feed_group_id: "40dd577b7e55490ab278301065f431a3"` (twice) with `feed_group_id: ${{ vars.OCTO_PR_FEED_GROUP_ID }}`.

Note: the PR description also claims `octo-pr-feed.yml` was deleted, but that file does not exist on `main` either — so no deletion is actually part of this diff (`changedFiles=1`). The description is slightly out of sync with the diff, but that is cosmetic.

## Verification

- ✅ **Trigger / `event_kind` change is correct.** With `reopened` removed from the triggers, the `pr_reopened` branch is genuinely unreachable, so collapsing the expression to `github.event.pull_request.merged == true && 'pr_merged' || 'pr_closed'` is accurate for the remaining `closed` event. (The consumer reusable workflow still carries a `pr_reopened` branch / `ALLOWED_KINDS` entry — harmless dead code, not this PR's concern.)
- ⚠️ **`feed_group_id` change is risky and likely does not achieve the stated intent** — see P1 below.

## Findings

### P1 — `feed_group_id: ${{ vars.OCTO_PR_FEED_GROUP_ID }}` can break notifications and does not match the stated goal

`.github/workflows/octo-pr-result-notify.yml` (both the `pr-result` and `review-result` jobs):

```yaml
feed_group_id: ${{ vars.OCTO_PR_FEED_GROUP_ID }}
```

Two problems compound here:

1. **The variable appears to be undefined.** Repo-level Actions variables for `Mininglamp-OSS/octo-ios` are empty (`total_count: 0`), so `OCTO_PR_FEED_GROUP_ID` is not set at the repo level. (I could not verify the org level — it requires admin scope.) In GitHub Actions, an undefined `vars.*` expands to an **empty string**.

2. **The consumer reusable workflow is fail-closed on an empty value.** `Mininglamp-OSS/.github/.github/workflows/octo-pr-result-notify.yml` validates the input:

   ```python
   def require_group_id(name):
       val = os.environ.get(name, '').strip()
       if not val:
           print(f'ERROR: required env var {name} is missing')
           sys.exit(2)
       if not re.fullmatch(r'[0-9a-f]{32}', val):
           print(f'ERROR: {name} must be a 32-char lowercase hex group id')
           sys.exit(2)
       return val
   ```

   So if `OCTO_PR_FEED_GROUP_ID` is undefined, the called job exits with code 2 and **the notification fails on every closed PR and every submitted review** — a regression from the current state, where the hardcoded `40dd577b…` value works.

3. **Even if the var is defined, it does not produce the behavior the PR describes.** The description says result-notify should send to the *global pr-feed (default)*. The reusable workflow already declares that default (`feed_group_id` default `'1c303c142e9840f2a9b46c10b0972e8d'`), but a reusable-workflow input default only applies when the input is **omitted**. Because this PR always passes the input (with a possibly-empty value), the default never kicks in. To actually get "the global default", the line should be deleted, not pointed at a variable.

**Recommended fix (pick one):**

- **Most robust — keep it configurable but safe:** add a literal fallback so an unset variable cannot break the job:
  ```yaml
  feed_group_id: ${{ vars.OCTO_PR_FEED_GROUP_ID || '1c303c142e9840f2a9b46c10b0972e8d' }}
  ```
- **Simplest — match the stated "use the global default" intent:** delete the `feed_group_id:` line in both jobs entirely, letting the reusable workflow apply its own default.
- **If the var route is intended:** confirm `OCTO_PR_FEED_GROUP_ID` is actually defined at the org level (set to the desired group id) *before* merge, since the repo level is empty.

## Additional notes (non-blocking)

- The sibling `octo-pr-review-feed.yml` still hardcodes the project group id `'40dd577b7e55490ab278301065f431a3'`. After this PR, result-notify uses a var while review-feed uses a literal — a small inconsistency. If centralizing group ids in `vars.*` is the long-term direction, consider doing both (with the same fallback guard) for consistency; not required for this PR.
- Please reconcile the PR description's "deleted `octo-pr-feed.yml`" line with the actual diff (the file isn't present on `main`), to avoid confusion in the history.

## Verdict

Requesting changes on the `feed_group_id` point (P1): as written, the notification job risks failing on every closed PR / review because the referenced variable is not defined at the repo level and the consumer guard is fail-closed, and the change does not deliver the "global default feed" behavior the description states. Any one of the three fixes above resolves it.

lml2468:
## Update: `feed_group_id` fallback added ✅

Addressed the blocking review concern: passing `${{ vars.OCTO_PR_FEED_GROUP_ID }}` directly would override the reusable workflow's built-in default with an empty string if the variable is unset, causing `require_group_id` validation to fail.

**Fix applied:** changed to `${{ vars.OCTO_PR_FEED_GROUP_ID || '1c303c142e9840f2a9b46c10b0972e8d' }}`

This ensures:
1. ✅ When org variable is set (current state) — uses `OCTO_PR_FEED_GROUP_ID` as configured
2. ✅ When org variable is unset — falls back to the global pr-feed group ID instead of passing empty string
3. ✅ Never bypasses the reusable workflow's safe default

The org-level variable `OCTO_PR_FEED_GROUP_ID` is confirmed to be set at org scope (per product owner). The fallback is a defensive measure only.

Jerry-Xin (changes requested):
The PR is relevant to this repository, but the workflow still allows overriding the destination feed, which contradicts the stated “global pr-feed default” design.

🔴 Blocking

🔴 Critical — `.github/workflows/octo-pr-result-notify.yml:25` and `.github/workflows/octo-pr-result-notify.yml:43` still pass `feed_group_id` explicitly via `${{ vars.OCTO_PR_FEED_GROUP_ID || '1c303c142e9840f2a9b46c10b0972e8d' }}`. The reusable workflow already defines the global pr-feed as its default input, so these caller-side overrides should be removed if the design is truly “send PR result/review result notifications to global pr-feed by default.” As written, any repository variable named `OCTO_PR_FEED_GROUP_ID` will silently reroute terminal PR and review-result notifications away from the global feed, recreating the ownership overlap this PR says it finalizes.

💬 Non-blocking

🟡 Warning — The PR description says `octo-pr-feed.yml` was deleted, but the actual merge diff only modifies `.github/workflows/octo-pr-result-notify.yml`. If the deletion happened in an earlier PR, the description should be clarified to avoid reviewers expecting an absent deletion in this change set.

✅ Highlights

The removal of `reopened` from the `pull_request_target` trigger is consistent with making this workflow terminal-state only.

The `review-result` job remains scoped to approved / changes-requested reviews from same-repository PR branches, which avoids sending unsupported review states and limits secret exposure.

lml2468 (commented):
CHANGES_REQUESTED

Gate check failed, so I am stopping before the diff review per the review process.

Note: the authenticated GitHub account is the PR author, so GitHub does not allow submitting this as a formal "request changes" review from this token.

Blocking status:
- `mergeStateStatus`: `BLOCKED`
- `check-sprint / check-sprint`: failing
- `build`: still pending
- CodeQL `Analyze`: still pending

Please rerun/fix the failing `check-sprint` job and wait for all required checks to complete successfully before requesting another review.

yujiawei (approved):
# Code Review — PR #6 (octo-ios)

## Summary

Small, well-scoped CI change (1 file, +4/-4) to `.github/workflows/octo-pr-result-notify.yml`. The diff:

1. Narrows the `pull_request_target` trigger from `[closed, reopened]` to `[closed]` (terminal-state only).
2. Simplifies the `event_kind` expression accordingly — drops the now-unreachable `pr_reopened` branch, leaving `merged == true ? 'pr_merged' : 'pr_closed'`.
3. Replaces the hardcoded `feed_group_id` in both jobs with `${{ vars.OCTO_PR_FEED_GROUP_ID || '1c303c142e9840f2a9b46c10b0972e8d' }}`.

The net effect matches the stated design: `octo-pr-result-notify.yml` owns terminal/result events → global pr-feed, while `octo-pr-review-feed.yml` owns `ready_for_review` / `review_requested` → project group. Verified clean separation via the job `if:` guards.

## Verification

- ✅ **YAML valid** — file parses cleanly; the earlier `***` secret-masking corruption (which made the file invalid and blocked both jobs) is gone. `OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}` is restored on both jobs.
- ✅ **`feed_group_id` fail-closed risk resolved** — the literal fallback means an unset `vars.OCTO_PR_FEED_GROUP_ID` no longer breaks the called job; it deterministically routes to the global pr-feed default. This is the robust fix.
- ✅ **Trigger narrowing is correct** — removing `reopened` makes this workflow terminal-state only, and the simplified `event_kind` covers both remaining cases.
- ✅ **`review-result` job is safely scoped** — gated to `approved` / `changes_requested` from same-repo head (`github.event.pull_request.head.repo.full_name == github.repository`), avoiding unsupported states and limiting secret exposure on forks.
- ✅ **Job separation is correct** — `pr-result` runs only on `pull_request_target`; `review-result` runs only on `pull_request_review`. No double-firing.
- ⚠️ **`octo-pr-feed.yml` deletion** — the description states this PR deletes `octo-pr-feed.yml`, but the file is not present on `main` and the merge diff only modifies `octo-pr-result-notify.yml`. The deletion happened within this branch's own commit history (added then removed), so the net result is correct, but the description overstates the change set. Non-blocking; consider clarifying.

## Findings

### Non-blocking

- **`OCTO_BOT_TOKEN` is currently empty (infra/secrets gap, not a code defect).** The `Octo PR Result Notify` run on the head SHA fails at the "Send result notification" step with `Missing required env var: OCTO_BOT_TOKEN`. The `feed_group_id` resolved correctly to the fallback, so the YAML change works as intended — but no notification will actually send until the `OCTO_BOT_TOKEN` secret is provisioned at the repo/org level. This is outside the scope of a YAML diff and predates this PR (the same step failed on the prior SHA). Flagging so it can be configured separately; it does not block this change.
- **Consistency:** the sibling `octo-pr-review-feed.yml` still hardcodes its `project_group_id` (`40dd577b...`) while result-notify now uses a `vars.* || literal` pattern. If centralizing group ids in repo/org variables is the long-term direction, applying the same `vars + fallback` pattern there would be consistent. Not required for this PR.

## Verdict

The diff is correct and resolves all previously-blocking findings (the invalid-YAML build break and the fail-closed `feed_group_id`). The remaining CI failures (`check-sprint`, and the notify step's missing `OCTO_BOT_TOKEN`) are unrelated to the correctness of this diff — they are an org-level secrets/config concern, not a defect the PR author can fix in YAML. Approving.

lml2468:
## Response to review feedback

**Re: `feed_group_id` explicit override concern**

The `feed_group_id: ${{ vars.OCTO_PR_FEED_GROUP_ID || '1c303c142e9840f2a9b46c10b0972e8d' }}` is **intentional per product owner direction**.

The product owner explicitly created an org-level Actions variable `OCTO_PR_FEED_GROUP_ID` and instructed that all `octo-pr-result-notify.yml` callers switch to reading `feed_group_id` from this variable. This allows centralized reconfiguration of the target group across all repos by updating a single org variable — without touching individual caller files.

**The fallback `'1c303c142e9840f2a9b46c10b0972e8d'`** is a defensive measure: if the org variable is ever unset, the caller passes the hardcoded default instead of an empty string (which would fail `require_group_id` validation in the reusable workflow).

GitHub Actions variable precedence (`vars.X`: repo → org → env) means a repo-level override would take precedence over the org variable — but this is a known trade-off accepted by the product owner. The design intent is org-wide configuration, not per-repo overrides.

**Regarding `octo-pr-feed.yml` deletion note in the description**: this file was never present on `main` for this repo (it was added and then removed within the branch history, netting to zero). The description references the broader batch of fixes applied across the org.

---

**Remaining blocking items (not in diff):**
1. `OCTO_BOT_TOKEN` org secret not accessible to this repo — needs org admin to add `octo-ios` to the secret's repository access list
2. Linked issue #9 needs Sprint W22 set by a Maintainer on the Octo Board

Jerry-Xin (approved):
The PR is in scope for `Mininglamp-OSS/octo-ios`: it updates repository-owned GitHub Actions notification behavior.

💬 Non-blocking
- 🔵 Suggestion: [.github/workflows/octo-pr-result-notify.yml](/Users/mlamp/.openclaw/tmp/pr-review-qy7x3r2p/repo/.github/workflows/octo-pr-result-notify.yml:21) now depends on the workflow trigger staying limited to `closed`. That is correct for this PR, but if more `pull_request_target` actions are added later, this expression would classify them as `pr_closed`. A short inline comment or restoring the explicit action check would reduce future maintenance risk.

✅ Highlights
- The `reopened` trigger removal matches the stated finalized design.
- The fallback `feed_group_id` is compatible with the reusable workflow’s expected 32-character lowercase hex group ID.
- `pull_request_review` still filters to `approved` and `changes_requested`, and keeps the existing same-repository guard before passing `OCTO_BOT_TOKEN`.

No blocking bugs, security regressions, or test requirements found for this workflow-only change.

PR #8, opened 2026-05-31
fix(voice-mention): 语音输入识别 @所有AI

### LINKED ISSUE #7: fix(voice): 语音 parseMentionMarkers 不识别 @所有AI
## Bug

`WKVoiceInputView` 和 `WKHoldToTalkManager` 的 `parseMentionMarkers` 只识别 @所有人/@all，不识别 @所有AI。

键盘路径已正确支持 @所有AI → uid "__ais__" → `mention.ais=1`（三态 mention 已完整实现），但语音路径缺失。

## Fix

两处 `parseMentionMarkers` 加入 @所有AI / All AIs 的匹配，uid 使用 "__ais__"（与键盘路径对齐）。

## Related
- Web: Mininglamp-OSS/octo-web#175
- Android: Mininglamp-OSS/octo-android#14
- Speech prompt: Mininglamp-OSS/octo-speech#35

yujiawei:
## 背景
语音输入不识别 @所有AI（YUJ-2531）。两处 voice parseMentionMarkers 都需要改。

## 改动
- `Modules/WuKongBase/WuKongBase/Classes/Vendor/CWVoiceClasses/WKVoiceInputView.m`
- `Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKHoldToTalkManager.m`

两处 `parseMentionMarkers` 在 @所有人/@all 之后加入 @所有AI 匹配（LLang 中文 + 英文 "All AIs" 别名），emit sentinel uid `__ais__`，WKConversationContextImpl 已将其映射到 mention.ais=1。英文别名在 @all 之前判断。

@所有人 不变：uid `all` 已正确走 WK_Mentioned_Humans / humans=1。

Fixes #7


Jerry-Xin (changes requested):
This PR is relevant to `octo-ios`, but the implementation misses a required locale-independent voice parsing case.

🔴 Blocking

🔴 Critical — `@所有AI` is not recognized when the app locale is English.
In both [WKHoldToTalkManager.m](/Users/mlamp/.openclaw/tmp/pr-review-4al4datb/repo/Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKHoldToTalkManager.m:1341) and [WKVoiceInputView.m](/Users/mlamp/.openclaw/tmp/pr-review-4al4datb/repo/Modules/WuKongBase/WuKongBase/Classes/Vendor/CWVoiceClasses/WKVoiceInputView.m:1276), `aisName` is set to `LLang(@"所有AI")`. In English locale this resolves to `"All AIs"`, so the Chinese canonical token is no longer matched by the new `hasPrefix:aisName` branch. The downstream mention service already handles broadcast labels locale-independently by including both canonical Chinese labels and English aliases, so the voice parser should do the same. Otherwise a user can speak or input the Chinese canonical token while running the English UI and the parser will leave it as plain text instead of emitting the `__ais__` sentinel.

💬 Non-blocking

🟡 Warning — The new Chinese-token branch has no boundary check.
At [WKHoldToTalkManager.m:1368](/Users/mlamp/.openclaw/tmp/pr-review-4al4datb/repo/Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKHoldToTalkManager.m:1368) and [WKVoiceInputView.m:1303](/Users/mlamp/.openclaw/tmp/pr-review-4al4datb/repo/Modules/WuKongBase/WuKongBase/Classes/Vendor/CWVoiceClasses/WKVoiceInputView.m:1303), any mention beginning with the Chinese canonical AI broadcast token is consumed as the broadcast token before member-name matching runs. That can incorrectly parse a real member or phrase whose name starts with that token. Consider using the same kind of end/space/punctuation boundary policy used for the English alias, or folding broadcast labels into the same longest-prefix strategy with explicit boundaries.

🟡 Warning — No test coverage appears to exercise the voice mention parser paths.
Given this logic is duplicated in two files and has locale-sensitive behavior, add focused coverage or a small shared parsing helper with tests for canonical Chinese token, English alias, legacy `@all`, mixed locale behavior, and prefix cases.

✅ Highlights

🔵 Suggestion — The downstream `__ais__` handling in `WKConversationContextImpl` and `WKInputMentionCache` is already aligned with the intended sentinel flow, and the PR correctly avoids adding the sentinel as a real mention entity.

yujiawei (commented):
# Code Review — PR #8 (octo-ios)

## Summary

This PR adds `@所有AI` (broadcast-to-all-AIs) recognition to the two voice-input mention parsers (`WKHoldToTalkManager.m` and `WKVoiceInputView.m`), emitting the `__ais__` sentinel uid that `WKConversationContextImpl` already maps to `mention.ais=1`. The intent and the sentinel routing are correct, but the matching logic has a locale-dependence bug that defeats the feature's own goal for English-locale users.

## Verification

- ✅ `__ais__` sentinel routing is sound. `WKConversationContextImpl.m:321` / `:389` / `:408` correctly map the sentinel to `mention.ais=1` and exclude it from `mention.entities`, and the SDK (`WKMessageContent.m:330`) does the same. The PR correctly avoids inserting `__ais__` as a real mention entity.
- ✅ The English alias branch is ordered before the legacy `@all` branch, so `"All AIs"` is not pre-empted by `"all"`.
- ❌ **The Chinese canonical token `所有AI` is not recognized when the app runs in English locale** (details below).
- ⚠️ The Chinese-token branch has no trailing boundary check.
- ⚠️ No test coverage for this duplicated, locale-sensitive parsing logic.

## Findings

### 🔴 P1 — `@所有AI` is unrecognized in English locale (functional bug, blocks merge)

`WKHoldToTalkManager.m:1341` and `WKVoiceInputView.m:1276`:

```objc
NSString *aisName = LLang(@"所有AI");   // English locale → "All AIs"
NSString *aisEnName = @"All AIs";
...
if ([rest hasPrefix:aisName]) { ... }   // English locale → hasPrefix:@"All AIs"
```

`LLang` is locale-dependent (`NSString+WKLocalized.m:20` → `LocalizedWithClass:` reads `[WKApp shared].config.langue`), and `en.lproj/Localizable.strings:631` defines `"所有AI"="All AIs"`. So in English locale `aisName` resolves to `"All AIs"`, making the "Chinese canonical" branch identical to the English-alias branch. The literal Chinese string `所有AI` is then never matched.

This matters because the wire format is locale-independent: a message composed on a Chinese-locale device carries the literal `@所有AI`, and a user on an English-locale device can also dictate/input the Chinese token. The existing downstream parser already accounts for this — `WKMentionService.m:37-56` (`broadcastLabels`) hardcodes **both** the Chinese canonical (`@"所有AI"`) **and** the English alias (`@"All AIs"`) plus the `LLang(...)` value, precisely so both directions hit regardless of receiver locale. The voice parsers should mirror that: match the hardcoded Chinese literal `@"所有AI"` in addition to the `LLang`/English-alias value, rather than relying on `LLang` alone.

Suggested fix: add an explicit hardcoded `@"所有AI"` branch (or reuse `WKMentionService.broadcastLabels` / a shared helper) so the canonical Chinese token is always matched independent of locale.

### 🟡 P2 — Chinese-token branch lacks a boundary check

`WKHoldToTalkManager.m:1368` / `WKVoiceInputView.m:1303`:

```objc
if ([rest hasPrefix:aisName]) { ... }
```

Unlike the English-alias branch (which checks end-of-string or a trailing space) and the `@all` branch, the Chinese branch consumes any token *beginning with* the broadcast label. A member or phrase whose name starts with the canonical token (e.g. `@所有AImembers`) would be mis-parsed as the broadcast token before per-member matching runs. Consider applying the same end/space/punctuation boundary policy used for the English alias.

### 🟡 P2 — No test coverage for locale-sensitive, duplicated logic

This parsing logic is duplicated verbatim in two files and is locale-sensitive. The locale bug above would have been caught by a focused test. Recommend extracting a shared parsing helper and adding tests for: canonical Chinese token, English alias, legacy `@all`, mixed-locale behavior, and prefix/boundary cases.

## Verdict

The sentinel routing and ordering are correct, but the locale-dependent matching means the feature does not work for English-locale users — which is the core behavior this PR is meant to deliver. Requesting changes for the P1; the P2 items are non-blocking but worth addressing alongside the fix.

lml2468 (changes requested):
Thanks for the update. The two parser correctness issues from the previous round are addressed:

- `@所有AI` is now matched as a hardcoded canonical Chinese token instead of using `LLang(@"所有AI")`, so English locale no longer loses the Chinese voice-recognition token.
- The Chinese and English AI broadcast branches now require a token boundary after the alias, which prevents prefix false positives such as a member name beginning with `所有AI`.

Blocking issue remains:

The PR still has no test coverage for the locale-sensitive voice mention parser behavior. The diff only changes the two parser implementations (`WKHoldToTalkManager.m` and `WKVoiceInputView.m`); I do not see added or updated tests for canonical Chinese `@所有AI`, English alias `@All AIs`, mixed-locale behavior, or prefix/boundary cases.

This was one of the previous blocking issues, and this logic is duplicated in two production paths, so we still need regression coverage before merge. Please add focused tests for the voice mention parsing cases above, including a negative prefix case like `@所有AI助手`.

Gate note: the branch is currently `MERGEABLE`; build and CodeQL are still pending, and the failed `check-sprint` job is treated as non-blocking per the review instructions.

Jerry-Xin (changes requested):
Project relevance passed; this PR is in scope, but the requested voice broadcast mention still fails when the local member cache is empty.

🔴 Blocking

🔴 Critical: Both voice paths only call `parseMentionMarkers` when `members.count > 0`, but this broadcast sentinel does not depend on member-name matching. If the member DB is empty or not synced yet, the transcribed broadcast text is sent as plain text, no `__ais__` item is added to the mention cache, and `mentionedInfo.ais` is never set. This conflicts with the existing keyboard-path design that exposes the sentinel independently of local member cache state. Fix both call sites by invoking the parser even with an empty member array, letting empty members only disable regular member-name matching.
References: [WKVoiceInputView.m](/Users/mlamp/.openclaw/tmp/pr-review-rrx56548/repo/Modules/WuKongBase/WuKongBase/Classes/Vendor/CWVoiceClasses/WKVoiceInputView.m:395), [WKHoldToTalkManager.m](/Users/mlamp/.openclaw/tmp/pr-review-rrx56548/repo/Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKHoldToTalkManager.m:1179), [WKConversationContextImpl.m](/Users/mlamp/.openclaw/tmp/pr-review-rrx56548/repo/Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/WKConversationContextImpl.m:964)

💬 Non-blocking

🟡 Warning: The new parsing branch always emits the canonical Chinese display label and stores that as `item.name`, even when the English alias was spoken or the current locale would display the localized label. That is functionally routable, but it is inconsistent with the keyboard mention path, which inserts the localized display label for the same sentinel. Consider separating accepted match labels from the emitted display label.
References: [WKHoldToTalkManager.m](/Users/mlamp/.openclaw/tmp/pr-review-rrx56548/repo/Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKHoldToTalkManager.m:1387), [WKVoiceInputView.m](/Users/mlamp/.openclaw/tmp/pr-review-rrx56548/repo/Modules/WuKongBase/WuKongBase/Classes/Vendor/CWVoiceClasses/WKVoiceInputView.m:1322), [WKConversationContextImpl.m](/Users/mlamp/.openclaw/tmp/pr-review-rrx56548/repo/Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/WKConversationContextImpl.m:1093)

🔵 Suggestion: The added token-boundary check rejects ASCII letters, digits, and underscore only. Non-Latin continuation characters after the broadcast label can still be consumed as a valid mention. The existing rich-text mention service has more explicit broadcast-token boundary handling, so a shared helper or focused tests would reduce drift.
References: [WKHoldToTalkManager.m](/Users/mlamp/.openclaw/tmp/pr-review-rrx56548/repo/Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKHoldToTalkManager.m:1368), [WKVoiceInputView.m](/Users/mlamp/.openclaw/tmp/pr-review-rrx56548/repo/Modules/WuKongBase/WuKongBase/Classes/Vendor/CWVoiceClasses/WKVoiceInputView.m:1303), [WKMentionService.m](/Users/mlamp/.openclaw/tmp/pr-review-rrx56548/repo/Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Services/WKMentionService.m:65)

✅ Highlights

The sentinel mapping itself is aligned with the downstream context code: `__ais__` is filtered out of regular message entities and mapped to the broadcast flag. The English multi-word alias is also checked before legacy `all`, which avoids the obvious prefix collision.

yujiawei (commented):
# Code Review — PR #8 (octo-ios)

## Summary

This PR extends `parseMentionMarkers:` in the two voice-input code paths (`WKVoiceInputView.m`, `WKHoldToTalkManager.m`) so that dictated text recognizes the `@所有AI` broadcast token (Chinese canonical + English alias `All AIs`), emitting the sentinel uid `__ais__`. I verified the full cross-file pipeline end-to-end and it is sound: `__ais__` flows through `WKVoicePanel.voiceInputDidTranscribe:mentions:` → `addMentionItems:` → `WKInputMentionCache.allMentionUid:` → `WKConversationContextImpl` where it sets `mentionedInfo.ais = YES` and is filtered out of `realUids`/`mention.entities` so it never leaks to the server as a real uid. The English-before-`@all` ordering is correct, the boundary checks on the new branches are correct, and `@所有人`/`@all` behavior is unchanged.

The mechanism is right, but there is one functional gap on the **primary trigger** that should be fixed before merge, plus several non-blocking items.

## Findings

### P1 — Chinese trigger `@所有AI` is matched case-sensitively; lowercased ASR output `@所有ai` silently fails

`WKHoldToTalkManager.m:1368` / `WKVoiceInputView.m:1303`

```objc
if ([rest hasPrefix:aisName] &&        // aisName = @"所有AI", case-SENSITIVE
    (rest.length == aisName.length || ...)) {
```

The Chinese branch uses a case-sensitive `hasPrefix:` on a token whose Latin substring is uppercase `AI`. The very next branch — the English alias — was deliberately made case-*insensitive* (`rest.lowercaseString hasPrefix: aisEnName.lowercaseString`) with an inline comment acknowledging that ASR casing is unreliable. That same defense was not applied to the Chinese path, which is the **main** trigger for this feature.

Speech-to-text on mixed CJK+Latin frequently lowercases incidental Latin letters, producing `@所有ai`. That string matches neither branch (Chinese fails on case; English fails because `所有ai` has no `all ais` prefix), falls through per-member matching, and is left as literal text — **no mention emitted, no `mention.ais=1`, no error**. For a voice-recognition feature whose entire purpose is recognizing spoken "@所有AI", the primary path silently no-ops on a realistic transcription.

**Suggested fix:** apply the same case-insensitive comparison used by the English branch — e.g. match the `所有` prefix, then compare the next two characters case-insensitively against `ai` (keeping the existing word-boundary guard).

### P2 — English near-misses silently mis-route to `@所有人` (all-humans) instead of all-AIs

`WKHoldToTalkManager.m:1382-1405` / `WKVoiceInputView.m:1317-1340`

Because the new alias branches have a strict alphanumeric word-boundary check but the older `@all` branch only treats a literal trailing space as a boundary, English near-misses of the AIS alias fall through and are swallowed by `@all` → `@所有人` (`mentionedInfo.humans=YES`):

- `@All AIsx` — alias boundary rejects the trailing `x`, then `@all` matches on the space after `All` and consumes it as @所有人, leaving ` AIsx` literal.
- `@All AI` (dropped plural `s`, a very common ASR rendering) — alias prefix `all ais` doesn't match, `@all` matches the space after `All`, consumes as @所有人, leaving ` AI` literal.

The result is not a harmless no-op: a user attempting an all-AIs broadcast instead pings **all humans** — a more disruptive, wrong recipient set. Consider accepting singular `All AI` and/or preventing the `@all` branch from swallowing these AIS near-misses.

### P2 — English-locale chip label inconsistency (voice vs keyboard picker)

`WKHoldToTalkManager.m:1374,1388` / `WKVoiceInputView.m:1309,1323`

Both new branches set `item.name = aisName` (hardcoded `@"所有AI"`) and append the Chinese `@所有AI` chip into the composer, regardless of locale. The keyboard-picker path (`WKConversationContextImpl.m:966`) instead uses `LLang(@"所有AI")`, which resolves to `All AIs` in English locale (`en.lproj/Localizable.strings:631`). So on an English device, dictating "All AIs" shows a Chinese `@所有AI` chip while the picker shows `@All AIs` for the identical action.

This is cosmetic only — `allMentionUid:` keys validation off `item.name` matched against the appended text, and both are `所有AI`, so `mention.ais=1` still fires correctly. The hardcoded comment's concern is about the *match target*, which is correct; the *display label* could be locale-aware (`LLang(@"所有AI")` for `item.name` + the appended text) without affecting routing.

### P2 / nit — ~32 lines of broadcast-token parsing duplicated verbatim across the two files

`parseMentionMarkers:` is byte-for-byte identical in `WKHoldToTalkManager.m` and `WKVoiceInputView.m`. Any future change to broadcast-token semantics (including the P1 fix above) must be applied in both places by hand; the receiving/wire side is already centralized in `WKMentionService.broadcastLabels`, so the parse side would benefit from the same treatment (a shared helper). No correctness impact today since the copies agree.

### nit — debug `NSLog` of full transcript + mentions left unguarded in `WKVoiceInputView.m:399-403`

These log the user's full dictated text and resolved mention uid/name on every transcription, with no `#if DEBUG` guard, in release builds. The same file already guards a separate length-only log with a "Critical privacy: 不打用户 context 内容" comment, so this is both a minor PII concern and an internal inconsistency. The sibling `WKHoldToTalkManager.m` has no such block. Recommend wrapping in `#if DEBUG` or removing.

## Verdict

The cross-file plumbing is correct and the feature works for clean input. However, the P1 case-sensitivity gap means the **primary Chinese trigger** can silently fail on a realistic ASR rendering, and the P2 near-miss mis-routing can broadcast to the wrong group. Both should be addressed before merge. Requesting changes; the P2/nit display and dedup items are non-blocking follow-ups.

Jerry-Xin (approved):
This PR is in scope for `octo-ios` and correctly extends the existing voice-input mention path to emit the `__ais__` sentinel used by the conversation send pipeline.

💬 Non-blocking

🟡 Warning: `WKVoiceInputView.m:1270` and `WKHoldToTalkManager.m:1337` add non-English code comments in shared OSS source. The implementation is understandable, but this project appears to target open-source review standards; keeping comments in English would make this easier to maintain for broader contributors.

🟡 Warning: `WKVoiceInputView.m:1272` and `WKHoldToTalkManager.m:1339` normalize the English spoken alias to the Chinese canonical visible token. Routing works because the mention cache checks the same emitted label, but English-locale users may see their dictated `@All AIs` rewritten as `@所有AI`. Consider using a shared locale-independent parser while preserving the matched display label when possible.

🔵 Suggestion: The two `parseMentionMarkers` implementations remain duplicated across `WKVoiceInputView.m` and `WKHoldToTalkManager.m`. Since this PR now needs the same sentinel and alias behavior in both places, a small shared helper would reduce future drift.

✅ Highlights

The PR correctly parses broadcast mentions even when the channel member list is empty, which is necessary for sentinel-only mentions.

The English alias is checked before the legacy `@all` branch, avoiding the obvious prefix collision.

lml2468 (approved):
Round 3 incremental re-review.

Verified at PR head `212fddb5d9f9d12214c43c041682869e0a698d5f` / `multica/2531-voice-ais`:

- The previous blocking `members.count > 0` issue is fixed in both voice paths. `WKHoldToTalkManager.m` and `WKVoiceInputView.m` now call `parseMentionMarkers` unconditionally with `(members ?: @[])`, so the `@所有AI` sentinel path no longer depends on the local member cache being populated.
- I did not find new blocking regressions in the incremental diff.

Remaining non-blocking follow-ups:

- No voice mention parsing tests were added in this update. Coverage is still missing for Chinese canonical, English alias, mixed-locale, and prefix-boundary cases.
- The Chinese `@所有AI` match remains case-sensitive, so dictated `@所有ai` is still not recognized.
- The existing `@all` fallback can still catch near-miss English phrases such as `@All AI` as `@everyone` rather than the AI broadcast sentinel.

CI gate: `check-sprint` is failing, which I am treating as non-blocking per review instructions. `build` and `Analyze` were pending at review time, with no blocking failure observed.

yujiawei (commented):
# Code Review — PR #8 (octo-ios)

## Summary

This PR teaches the two voice-input mention parsers to recognize the `@所有AI` broadcast token, bringing the voice path to parity with the keyboard path. It modifies `parseMentionMarkers:` in both `WKVoiceInputView.m` and `WKHoldToTalkManager.m` to match the Chinese canonical token `所有AI` plus the English alias `All AIs`, emitting the `__ais__` sentinel uid that the conversation pipeline already maps to `mention.ais=1`. It also makes both call sites invoke `parseMentionMarkers:` unconditionally instead of only when `members.count > 0`.

## Verification

- **Sentinel mapping is correct.** `__ais__` is the established sentinel across the codebase: `WKConversationContextImpl.m:389/406-408` sets `mentionedInfo.ais = YES` when `__ais__` is present, and `:321` / `WKMessageContent.m:331` correctly exclude the sentinel from `mention.entities` so it is never sent as a real user uid. The voice mentions reach this path via `WKVoicePanel.m:350-352` (`addMentionItems:`) → `WKInputMentionCache.allMentionUid:` → `mentionedInfo:`. ✅
- **Empty-member-cache fix is correct and necessary.** A broadcast token does not depend on member-name matching, so gating the parse on `members.count > 0` previously dropped `@所有AI` in channels where the local member cache was empty. Calling `parseMentionMarkers:members:(members ?: @[])` unconditionally fixes this without affecting name-based matching (the name loop simply iterates an empty list). ✅
- **English-locale correctness.** Hardcoding `aisName = @"所有AI"` rather than `LLang(@"所有AI")` is the right call — the voice recognizer emits the Chinese token regardless of UI locale, so an `LLang` lookup would break matching under an English locale. ✅
- **Ordering of the English alias.** The `All AIs` branch is correctly placed before the generic `@all` branch (`WKVoiceInputView.m:1313` before `:1326`). Without this, `@all` (`hasPrefix:"all"` with a trailing space before `AIs`) would consume `All` and misclassify the broadcast as `@所有人`. The inline comment documents this. ✅
- **Word-boundary handling.** Both new branches guard the trailing boundary (`rest.length == name.length` or next char not alphanumeric and not `_`), so `@所有AI的事` matches while `@所有AIsomething` does not. Consistent with intent. ✅
- **`WKInputMentionCache.allMentionUid:` ghost-mention guard.** The sentinel is re-validated against the visible `@所有AI` label still being present in the send text, so deleting the token after dictation does not leak a phantom mention. ✅

## Findings

No P0 (correctness/security/data-loss/build-break) or P1 (functional blocker) issues found.

### P2 / Nits (non-blocking)

- **Lowercase Chinese-token edge case.** The Chinese branch uses a case-sensitive `hasPrefix:aisName` where `aisName` contains uppercase `AI`. If the speech-to-text engine ever transcribes the token as `所有ai` (lowercase), it will fall through to the English/`@all`/name branches and miss. Low likelihood for Chinese STT output, but a case-insensitive compare on the Latin suffix would be more robust. Not blocking.
- **Non-English inline comments in shared OSS source.** The added comments are in Chinese in files under a public OSS module. Consistent with surrounding code in these files, but English comments would be friendlier for external contributors. Stylistic only.
- **Duplication across the two parsers.** The `所有AI` / `All AIs` matching logic is copy-pasted in both `parseMentionMarkers:` implementations. Acceptable for this fix given the two methods are already near-duplicates; a future refactor could extract a shared helper. Not in scope here.

## Verdict

The change is focused, correct, and addresses the two blocking issues raised in earlier review rounds (locale-independence and the empty-member-cache gate). The remaining items are P2/nits and do not block merge.

PR #12, opened 2026-06-01
fix(typing): 后台回前台 typing 卡死 — 落库清 typing + 回前台 reset (#10)

yujiawei:
## 问题

octo-ios#10：群聊中 App 切后台、bot 回复完成后切回前台，typing indicator 永久卡死、消息看似不刷新。

## 根因

typing 清除的唯一可靠路径在 `onRecvMessages`（push 路径）。后台期间 bot 回复经 conversation-sync 直写 DB，绕过该路径 → typing 永不清。叠加回前台无任何 typing reset 机制。

### Fix-1：conversation-sync 落库后清 typing
- `WKConversationManager handleSyncConversation:` 写库后，对有 `recents` 的 channel 发 `WKConversationSyncedNewMessages` 通知。
- `WKTypingManager` 监听该通知 → `removeTypingByChannel:`（统一切到主线程执行）。
- `WKMessageListView onConnectStatus:WKConnected` 的 pullup 回调里，若当前 channel 仍有 typing 显式清掉。

### Fix-2：回前台 reset 全部 typing
- `WKApp appDidBecomeActive:` 遍历 `getAllTypingMessages` 清除所有 typing。若对方仍在输入，下一个 typing CMD 到达后自然恢复。

## 副作用

回前台瞬间 typing 会被清掉，若对方确实还在输入会闪一下，可接受。改动均在主线程，不引入新的线程安全问题。Fix-3/4 留后续独立 PR。

## 验证

1. 群聊让 bot 回复，App 切后台等回复完成 → 切回前台：typing 立即消失，消息正常显示。
2. 正常前台聊天：typing 显示/消失不受影响。

Fixes #10

PR #13, opened 2026-06-01
fix(typing): Fix-3 background timer + dict locking, Fix-4 syncCompleted refresh hook

### LINKED ISSUE #10: 🐛 [iOS] 群聊中龙虾已回复但切回 App 后仍显示「工作中」，消息不刷新
## 问题描述

**反馈人**：lejun（DMWork uid:f145574e162a4fa1a056b511be03aa77）、辉哥（DMWork uid:eca0702f83e048c7b6151b21b1a3b9de）
**端**：iOS
**版本**：未提供

## 复现步骤

1. 在群聊中与龙虾对话（发送消息给龙虾）
2. 龙虾开始处理时，切出 Octo App（切到其他应用）
3. 龙虾在后台完成回复
4. 切回 Octo App

## 期望行为

切回 App 后，能看到龙虾的回复消息，typing indicator（工作中状态）消失

## 实际行为

切回 App 后，界面仍显示龙虾「工作中」（三个点 typing indicator），新消息不刷新。需要退出当前 session 再重新进入，消息才显示

## 截图

iOS typing indicator stuck

## 📢 反馈记录

- 2026-05-30 user:lejun (uid:f145574e162a4fa1a056b511be03aa77) via Bug 反馈群 (gid:30ac52c2ccba4f8d9f152311f1f785b5)
- 2026-05-30 user:辉哥 (uid:eca0702f83e048c7b6151b21b1a3b9de) via Bug 反馈群 (gid:30ac52c2ccba4f8d9f152311f1f785b5)
- 2026-05-30 user:刘沛 (uid:5ebd42f85c1445c4bbd0c6d677b85056) via Bug 反馈群 (gid:30ac52c2ccba4f8d9f152311f1f785b5)

## 🔑 闭环关键词

工作中, typing, 正在输入, 三个点, 消息不刷新, 切回, 前台, 后台, resume, foreground

---

**原始来源**：DMWork group:Bug 反馈群 (gid:30ac52c2ccba4f8d9f152311f1f785b5) user:lejun (uid:f145574e162a4fa1a056b511be03aa77) + user:辉哥 (uid:eca0702f83e048c7b6151b21b1a3b9de)

yujiawei:
Second batch of fixes for #10 (PR #12 shipped Fix-1 + Fix-2). This PR does Fix-3 and Fix-4.

## Fix-3 — typing 8s timeout timer → background queue + dict locking

`WKTypingManager` previously scheduled the 8s typing-timeout `dispatch_after` on the **main queue**. When the app is backgrounded the main queue is suspended, so the timer freezes and stale typing indicators stick around after returning to the foreground.

- Guard `channelTypingMessageDict` / `cancelTypingBlockDict` with a dedicated `@synchronized` lock (prerequisite — the background timer callback must not touch an unlocked `NSMutableDictionary`, which crashes under concurrent multi-channel typing).
- Move the 8s timer to `dispatch_get_global_queue(QOS_CLASS_UTILITY, 0)` so it keeps firing while backgrounded.
- Timer callback re-checks the message timestamp (`now - timestamp >= 8.0`) before clearing, so a refreshed typing event isn't wrongly removed.
- `removeTypingByChannel` now also cancels and drops the pending timer block.
- `+shared` singleton switched from `if(_instance==nil)` to `dispatch_once` (per YUJ-2595).

`removeTypingByChannel`'s delegate callback (`callTypingRemoveDelegate`) already dispatches to the main thread, so firing it from the background timer is safe for UI.

## Fix-4 — iOS syncCompleted full-refresh hook

iOS lacked Android's `syncCompleted → getData(true)` equivalent. Messages received while backgrounded arrive via conversation-sync writing straight to the DB, bypassing `onRecvMessages`' incremental insert, so the open chat could miss them.

- `WKMessageListView` now adopts `WKConversationManagerDelegate` and registers/unregisters it.
- `onConversationSyncFinished` checks whether the current channel's DB last message is newer than the UI's; if so it forces a refresh (pullup when at bottom, last-message ref update otherwise), so new messages show up even when pullup's drift fallback would swallow the refresh.

## Verification
1. Background > 8s → foreground → typing auto-clears within 8s (Fix-3 background timer).
2. Receive a message while backgrounded → foreground → message list auto-refreshes (Fix-4).
3. High concurrency (many groups typing at once) does not crash (Fix-3 locking).

Part of #10

Closes #10

### DIFF
--- a/Modules/WuKongBase/WuKongBase/Classes/Sections/Common/WKTypingManager.m
+++ b/Modules/WuKongBase/WuKongBase/Classes/Sections/Common/WKTypingManager.m
@@ -23,6 +23,13 @@
 
 @property(nonatomic,strong) NSMutableDictionary<WKChannel*,dispatch_block_t> *cancelTypingBlockDict; // 取消输入中状态的的block
 
+/**
+ *  dictLock 用于保护 channelTypingMessageDict / cancelTypingBlockDict 的读写。
+ *  超时 timer 现已挂在后台队列(见 addTypingByMessage)，回调会从后台线程访问这两个
+ *  dict，无锁访问 NSMutableDictionary 在多群并发 typing 时会崩溃，故统一加锁。
+ */
+@property(nonatomic,strong) NSObject *dictLock;
+
 @property(nonatomic,assign) BOOL offTyping; // 是否关闭typing
 @end
 
@@ -39,15 +46,20 @@ static WKTypingManager *_instance = nil;
 }
 
 +(instancetype) shared{
-    if (_instance == nil) {
-        _instance = [[super alloc]init];
-    }
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        _instance = [[super alloc] init];
+    });
     return _instance;
 }
 
 - (instancetype)init {
     self = [super init];
     if (self) {
+        // 提前初始化共享状态，保证后台 timer 回调与主线程访问看到的是同一份 dict / lock。
+        _channelTypingMessageDict = [[NSMutableDictionary alloc] init];
+        _cancelTypingBlockDict = [[NSMutableDictionary alloc] init];
+        _dictLock = [[NSObject alloc] init];
         // 监听 conversation-sync 落库通知：后台期间 bot 回复经同步直写 DB 绕过
         // onRecvMessages，导致 typing 卡死，这里收到通知后清除对应 channel 的 typing。
         [[NSNotificationCenter defaultCenter] addObserver:self
@@ -90,6 +102,13 @@ static WKTypingManager *_instance = nil;
     return _cancelTypingBlockDict;
 }
 
+- (NSObject *)dictLock {
+    if (_dictLock == nil) {
+        _dictLock = [[NSObject alloc] init];
+    }
+    return _dictLock;
+}
+
 - (NSLock *)delegateLock {
     if (_delegateLock == nil) {
         _delegateLock = [[NSLock alloc] init];
@@ -116,7 +135,10 @@ static WKTypingManager *_instance = nil;
 }
 
 -(BOOL) hasTyping:(WKChannel*)channel {
-    WKMessage *message = self.channelTypingMessageDict[channel];
+    WKMessage *message = nil;
+    @synchronized (self.dictLock) {
+        message = self.channelTypingMessageDict[channel];
+    }
     if(message) {
         return true;
     }
@@ -124,31 +146,42 @@ static WKTypingManager *_instance = nil;
 }
 
 - (void)addTypingByMessage:(WKMessage *)typingMessage {
-    
+
 //    WKMessage *typingMessage = [self convertMessageToTypingMessage:message];
     if( [typingMessage.fromUid isEqualToString:[WKApp shared].loginInfo.uid]) {
         return;
     }
-    
+
     WKChannel *channel = typingMessage.channel;
-    WKMessage *oldTypingMessage = self.channelTypingMessageDict[channel];
-    self.channelTypingMessageDict[channel] = typingMessage;
-    
-    dispatch_block_t cancelTypingBlock = self.cancelTypingBlockDict[channel];
-     if(cancelTypingBlock) {
-         dispatch_block_cancel(cancelTypingBlock);
-     }
-     __weak typeof(self) weakSelf = self;
-     cancelTypingBlock = dispatch_block_create(0, ^{
-         [weakSelf removeTypingByChannel:channel newMessage:nil];
-     });
-     self.cancelTypingBlockDict[channel] = cancelTypingBlock;
-     dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(8 * NSEC_PER_SEC)), dispatch_get_main_queue(),cancelTypingBlock);
-    if(!oldTypingMessage) {
+    __weak typeof(self) weakSelf = self;
+    BOOL isNewTyping = NO;
+    @synchronized (self.dictLock) {
+        WKMessage *oldTypingMessage = self.channelTypingMessageDict[channel];
+        self.channelTypingMessageDict[channel] = typingMessage;
+        isNewTyping = (oldTypingMessage == nil);
+
+        dispatch_block_t oldCancelBlock = self.cancelTypingBlockDict[channel];
+        if(oldCancelBlock) {
+            dispatch_block_cancel(oldCancelBlock);
+        }
+        // timer 改挂后台队列，避免 App 进后台时 main queue 挂起导致 8s 超时冻结，
+        // 切回前台后 typing 卡死不消失。回调内按时间戳判定是否过期再清除。
+        dispatch_block_t cancelTypingBlock = dispatch_block_create(0, ^{
+            WKMessage *cur = nil;
+            @synchronized (weakSelf.dictLock) {
+                cur = weakSelf.channelTypingMessageDict[channel];
+            }
+            if (cur && ([[NSDate date] timeIntervalSince1970] - cur.timestamp >= 8.0)) {
+                [weakSelf removeTypingByChannel:channel newMessage:nil];
+            }
+        });
+        self.cancelTypingBlockDict[channel] = cancelTypingBlock;
+        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(8 * NSEC_PER_SEC)),
+                       dispatch_get_global_queue(QOS_CLASS_UTILITY, 0), cancelTypingBlock);
+    }
+    if(isNewTyping) {
         [self callTypingAddDelegate:typingMessage];
     }
-     
-    
 }
 
 -(WKMessage*) convertParamToTypingMessage:(NSDictionary*)param {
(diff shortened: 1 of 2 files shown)


Jerry-Xin (approved):
## ✅ APPROVED

Fix-3 (background timer + dict locking) and Fix-4 (syncCompleted refresh hook) — both well-executed. Completes the typing-stuck fix series from #12.

**What I verified:**

### Fix-3: Background timer + dict locking

1. **Timer moved to `QOS_CLASS_UTILITY` queue:** Correct fix — main queue suspends in background, utility queue keeps firing. The timestamp re-check (`now - timestamp >= 8.0`) before clearing prevents a race where a fresh typing event arrives just as the old timer fires. Clean.

2. **`@synchronized(self.dictLock)` locking:** All dict read/write paths now hold the lock — `addTypingByMessage`, `hasTyping`, `getAllTypingMessages`, `getTypingMessage`, `removeTypingByChannel`, and the timer callback. Consistent coverage.

3. **`removeTypingByChannel` now cancels+drops the timer block:** Critical — prevents stale timers from firing after explicit removal. Previously missing; this would have caused double-remove or ghost callbacks.

4. **Eager init of dicts + lock in `init`:** Eliminates the lazy-getter race window for background timer callbacks. The remaining lazy getters for `channelTypingMessageDict`, `cancelTypingBlockDict`, and `dictLock` are now dead code (init sets the ivars directly via `_ivar = ...`). Not a bug — they're harmless fallbacks — but could be cleaned up in a follow-up.

5. **Singleton `shared` → `dispatch_once`:** Proper fix. The old `if (_instance == nil)` pattern was technically racy (though practically fine for a main-thread-only call path). Belt-and-suspenders with `allocWithZone:`'s existing `dispatch_once`.

6. **Delegate callbacks stay outside `@synchronized`:** `callTypingAddDelegate` and `callTypingRemoveDelegate` are called outside the lock — correct pattern to avoid holding the lock during potentially long delegate dispatch chains.

### Fix-4: syncCompleted refresh hook

7. **`WKConversationManagerDelegate` adoption:** Clean integration — `addDelegates`/`removeDelegates` properly paired. `onConversationSyncFinished` is dispatched to main thread by the SDK's `callOnConversationSyncFinishedDelegates` — verified in source.

8. **DB vs UI comparison logic:** `clientMsgNo` equality check to determine if UI is stale, then branches on `pullupHasMore` — if pagination in progress, only updates the last-message ref without disrupting the scroll anchor. If at bottom, does a full pullup. Handles both cases correctly.

9. **Typing cleanup in the same path:** Clearing residual typing when new messages arrive via sync — consistent with the Fix-1/Fix-2 defense-in-depth from #12.

**💬 Non-blocking:**

- The lazy getters for `channelTypingMessageDict`, `cancelTypingBlockDict`, and `dictLock` are now unreachable dead code since `init` eagerly sets the ivars. Consider removing them to avoid confusion about which initialization path is active. Low priority — they're harmless.

lml2468 (approved):
## Verdict: APPROVED ✅

Solid follow-up to PR #12. Fix-3 and Fix-4 each address a real gap; locking and dispatch are correct.

### Fix-3: Background timer + dict locking

**Timer fix is correct.** The root cause: `dispatch_after` on `dispatch_get_main_queue()` freezes when the app is backgrounded (main queue suspended). Moving to `dispatch_get_global_queue(QOS_CLASS_UTILITY, 0)` keeps the timer firing. The callback re-reads `channelTypingMessageDict[channel]` under lock and re-checks `now - timestamp >= 8.0` before clearing — prevents wrongly removing a refreshed typing event that arrived after the timer was scheduled. Good.

**Locking pattern is sound.** `@synchronized(self.dictLock)` consistently guards all dict reads/writes. Delegate callbacks (`callTypingAddDelegate`, `callTypingRemoveDelegate`) are correctly called **outside** the lock, avoiding deadlocks with UI code that might re-enter the typing manager.

**`removeTypingByChannel:` now cancels the pending timer block** inside the synchronized section — prevents the background timer from firing after an explicit removal. The `callTypingRemoveDelegate` dispatch-to-main-thread in existing code ensures UI safety when called from the background timer callback.

**`dispatch_once` for `+shared`** — good hardening over the previous `if(_instance==nil)` pattern.

**TOCTOU in timer callback is benign:** Between releasing the lock (after reading `cur`) and calling `removeTypingByChannel:`, another thread could remove the typing first. `removeTypingByChannel:` re-checks under lock and no-ops if already gone. No crash, no double-fire of delegate.

### Fix-4: syncCompleted refresh hook

**Correct gap fill.** iOS lacked Android's `syncCompleted → getData(true)`. The `WKConversationManagerDelegate` protocol already declares `onConversationSyncFinished` and the SDK dispatches it to the main thread via `callOnConversationSyncFinishedDelegates`. Delegate add/remove lifecycle in `addDelegates`/`removeDelegates` is properly paired.

**Pagination-aware refresh:** `pullupHasMore` branch avoids full reload when the user hasn't scrolled to bottom — just updates `lastMessage` reference, letting pagination fill in naturally. At-bottom branch does a pullup then updates. Correct.

**clientMsgNo comparison** for identity check is the right approach — avoids false positives from content edits that don't change identity.

### Non-blocking

1. **Redundant `dictLock` lazy getter:** `init` now eagerly sets `_dictLock`, so the lazy getter `if(_dictLock == nil)` never triggers. The getter is dead code. Not harmful, but slightly misleading since `@synchronized(nil)` is a silent no-op in ObjC — if someone later removes the eager init without noticing, the lock would silently stop protecting. Consider removing the lazy getter and letting the ivar speak for itself.

2. **`onConversationSyncFinished` fires for all channels:** The callback does a `getLastMessage:` DB lookup on every sync completion, even if the sync was for a different channel. Single read is fast so no performance concern, but if a channel filter is available from the SDK in the future, it would be a nice optimization.

yujiawei (commented):
# Code Review — PR #13 (octo-ios)

**Verdict: APPROVED** — no P0/P1 blockers. Two well-scoped fixes (background typing timer + dict locking; `syncCompleted` full-refresh hook). The locking is comprehensive and thread-safety reasoning is sound. A few P2/nit observations below; none block merge.

## 1. Verification

**Fix-3 — typing 8s timeout timer → background queue + dict locking**

- ✅ All accesses to `channelTypingMessageDict` / `cancelTypingBlockDict` are now guarded by `@synchronized(self.dictLock)`: `hasTyping` (`WKTypingManager.m:139`), `addTypingByMessage` (`:158`), `getAllTypingMessages` (`:217`), `getTypingMessage` (`:223`), `removeTypingByChannel` (`:231`), and the timer block (`:171`). No unlocked path remains.
- ✅ Dicts and `dictLock` are pre-initialized in `init` (`WKTypingManager.m:60-62`), so the background timer callback and main-thread accessors observe the same instances — the lazy getters never race to create a second instance.
- ✅ UI safety of the background-fired callback is preserved: `removeTypingByChannel` calls `callTypingRemoveDelegate` (`:243`), which marshals every delegate call onto the main thread (`:296-302`).
- ✅ The timestamp re-check (`:174`, `now - cur.timestamp >= 8.0`) is the real correctness guarantee: even if `dispatch_block_cancel` is a no-op because the block already started, a refreshed typing event (newer `timestamp`) prevents an erroneous clear. Belt-and-suspenders with the per-event cancel/recreate.
- ✅ No deadlock: the timer block releases `dictLock` before calling `removeTypingByChannel` (which re-acquires it); `@synchronized` is recursive regardless.
- ✅ `+shared` → `dispatch_once` (`:48-54`) composes correctly with the existing `allocWithZone:` `dispatch_once` (`:40-46`).
- ✅ `isNewTyping` computed under lock, `callTypingAddDelegate` invoked outside the lock (`:182-184`) — same add-only semantics as before, and not holding the lock across the delegate fan-out is an improvement.

**Fix-4 — `syncCompleted` full-refresh hook**

- ✅ `WKConversationManagerDelegate` is adopted (`WKMessageListView.m:25`) and registered/unregistered symmetrically (`:121` add, `:135` remove). The manager holds delegates weakly (`WKConversationManager.m:425`), so no retain cycle.
- ✅ `onConversationSyncFinished` is dispatched to the main thread by the SDK (`WKConversationManager.m:520-526`), so the UITableView work in the handler is main-thread-safe — the "no `dispatch_async` needed" comment is accurate.
- ✅ The handler mirrors the established `onConnectStatus:WKConnected` refresh path (`WKMessageListView.m:2408-2431`): guards `channel`/DB-last-msg, short-circuits when UI and DB last msg match, splits on `pullupHasMore` (update last-msg ref vs. full pullup). Consistent with existing patterns.
- ✅ `getLastMessage` returns a persisted message (typing messages are `noPersist=YES`, `WKTypingManager.m:196`), so the DB-vs-UI comparison can't be polluted by a typing placeholder; `pullupHasMore` also handles a `WK_TYPING` tail (`:840-846`).

## 2. Issues

None at P0/P1.

## 3. Suggestions (P2 / nits — non-blocking)

- **P2 — Queue-move rationale is imprecise (behavior is still correct).** The PR states moving the timer off the main queue keeps it firing "while backgrounded." In practice, once iOS suspends the app the *global* `QOS_CLASS_UTILITY` queue is suspended too, so the timer does not fire during background on either queue — it fires on resume. The actual robustness comes from the timestamp re-check (`WKTypingManager.m:174`) plus the `WKConversationSyncedNewMessages` observer clear (`:77-89`). The queue change is benign, but consider tightening the comment at `:167-168` so future readers don't rely on a background-fire that doesn't happen.
- **P2 — Foreground produces two overlapping refresh paths.** On foreground/reconnect, both `onConnectStatus:WKConnected` (`:2409`) and `onConversationSyncFinished` (`:2441`) run a `pullup`. They are complementary (the connect-time pullup may run before synced rows are written to DB; the sync-finished one runs after), and `updateLastMsgIfNeed`'s `orderSeq` guard (`:1813`) makes the second a no-op when there's nothing new — so this is acceptable. Worth being aware of in case `pullup` redundancy ever shows up in profiling.
- **nit — `dictLock` lazy getter TOCTOU.** `- (NSObject *)dictLock` (`:105-110`) has a check-then-create that two threads could theoretically both enter. It never triggers because `init` pre-populates `_dictLock`, and it matches the existing `delegateLock` pattern — leaving as-is is fine.

## 4. Additional observations

- The `clientMsgNo`-based equality used for the DB-vs-UI last-message comparison (`:2450`, `:2472`) is consistent with the existing reconnect path and relies on received messages carrying a stable `clientMsgNo`. No change needed; flagging only as the assumption both fixes share.

Overall this is a clean, defensive change with correct thread-safety. Approving.

PR #19, opened 2026-06-03
feat(richtext): iOS send RichText=14 mixed text+image (Phase 1)

### LINKED ISSUE #18: feat(richtext): iOS 发送图文混排 RichText=14 (Phase 1)
## What

iOS **发送侧** 图文混排 RichText(=14) Phase 1（文本+图片混排）。承接已合并的接收侧 PR #16（WKRichTextContent + WKRichTextCell），与 web 发送侧 #227 / Android 发送侧 #29 对称，凑齐四端收发对齐。

## 对称实现要求（参照 web #227）

1. **聚合成单条 type=14 消息**：聊天输入区同时含文本+图片且无非图片文件时，构造 **一条** `type=14` payload（content 块按交错顺序），而不是逐块发独立消息。纯文本/纯图/含文件路径**保持不动**，零回归。
2. **复用接收侧权威 schema 不另造**：用 #16 的 `WKRichTextContent.h/.m` 做 block 构造/编解码，wire-format 与 octo-lib 权威 schema 字节对齐；补 encode↔decode 往返单测。
3. **plain 非权威**：client 填本地占位（image → 占位 wire token，非本地化文案），server #232 Finalize 重算覆盖。
4. **安全（与接收侧对称）**：每个图片 URL 先上传再校验 http/https，阻断 javascript:/data:/file: 注入；不安全或失败的图片跳过并提示。
5. **mention 合并**：文本块的 all/uids/humans/ais 合并到单条消息，群 @ 通知（含 @所有AI）不丢。

## Out of scope (Phase 1 锁定)

富文本格式（粗体/标题等）—— Phase 1 只做文本+图片。不动其他仓，不回滚 #16/#218/#232。

## 验证

- Xcode build 通过
- RichText 单测（含发送侧新用例：block 构造 / size=0 不注入 / plain 占位 / encode↔decode 往返）

## 一起合并

发送侧 PR 满足 same-HEAD 2-reviewer 门槛后合并到 main（接收侧 #16 已在 main）。

yujiawei:
## What

iOS **send-side** for 图文混排 RichText(=14), Phase 1 (text + image mixed). Completes the four-end loop after receive-side **#16** (`WKRichTextContent` + `WKRichTextCell`, already on main), symmetric with web **#227**.

## Entry point

The iOS chat input area is **text-only** — the album picker sends images immediately and ignores any typed text. The only iOS surface that submits **text + images together in one user action** is the **Share Extension confirm panel** (`WKApp -sendShareFiles:toChannel:extraText:`: an attached-files list + an accompanying-text field). That is the faithful iOS analog of web #227's interleaved editor, so the aggregation lives there.

## Changes

1. **Aggregate into a single `type=14` payload** (`WKApp.m`): when **every** share item is an image **and** accompanying text is present, `sendRichTextMixedImages:` builds **one** RichText payload (image blocks in order, then the text block). Pure text, pure image, and any path containing files/links/video stay on the existing per-item send path — **zero regression**.
2. **Reuse the receive-side schema, no fork** (`WKRichTextContent.h/.m`): added sender-side `+textBlock:` / `+imageBlock:...` / `+contentWithBlocks:` builders and the optional `size` / `name` fields, wired through both `decodeWithJSON:` and `encodeWithJSON`. `size`/`name` serialize **only when set**, keeping wire-format byte-aligned with the octo-lib authoritative schema. Round-trip `encodeWithJSON`↔`decodeWithJSON` covered by new tests.
3. **plain stays non-authoritative**: client fills a local placeholder using the **non-localized** `[图片]` wire token (distinct from the receive-side localized display placeholder `LLang(@"[图片]")`); server #232 `Finalize` recomputes and overwrites.
4. **Security (symmetric with receive side)**: each image is uploaded first, then its URL is validated `http`/`https` with a non-empty host before going into a block — blocks `javascript:`/`data:`/`file:` injection.
5. **Atomic** (contract §5.1): any image failure (missing file / unreadable dimensions / upload error / unsafe URL) aborts the **whole** message via a single terminal-state guard — never "sends half". Share dir is cleaned up on both success and failure.

## Notes

- Image dimensions read via **ImageIO metadata** and byte size via **file attributes** — no full-image decode, avoiding memory spikes that could get the share extension killed on multi-image shares.
- `mention`: the Share Extension has no @-mention entry, so there is nothing to merge here; the payload construction path itself carries no mention (correct for this entry point). The schema is unchanged for other producers.

## Out of scope (Phase 1 locked)

Rich-text formatting (bold/headings/etc.). No changes to other repos; no rollback of #16 / #218 / #232.

## 审查状态

- **改动规模**: +444 / -6, 6 文件
- **/review** (self-audit): PASS ✅ — every changed line traces to the send-path / schema / security / atomicity asks; existing pure-text/pure-image/file/link paths untouched; image URLs scheme+host gated before wire; plain treated as non-authoritative.
- **/codex**: PASS ✅ — independent review across two rounds; all flagged issues fixed (terminal-state race guard, ImageIO metadata instead of full decode, host check in URL validation, raw file path to upload task avoiding percent-encoding breakage, share-dir cleanup on failure).

## 验证

- Xcode build / RichText 单测：本机为 Linux，无 `xcodebuild`，未能本地跑测。新增 `WKRichTextSendTests`（block 构造 / size=0&name 空不注入 wire / plain 非本地化占位 / encode↔decode 往返 4 类 8 用例）已注册进 `WuKongBase/Example` 测试 target，待 CI（macOS）执行。

Fixes #18

### DIFF
--- a/Modules/WuKongBase/Example/LiMaoBase.xcodeproj/project.pbxproj
+++ b/Modules/WuKongBase/Example/LiMaoBase.xcodeproj/project.pbxproj
@@ -26,6 +26,7 @@
 		873B8AEB1B1F5CCA007FD442 /* Main.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 873B8AEA1B1F5CCA007FD442 /* Main.storyboard */; };
 		93FBC126DB4B04BC75AF04E2 /* WKExternalViewerResolverTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 85EB8C00E7C49A151B670678 /* WKExternalViewerResolverTests.m */; };
 		A1410000000000000000F001 /* WKJoinGroupSuccessHelperTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A1410000000000000000F002 /* WKJoinGroupSuccessHelperTests.m */; };
+		A1990000000000000000F001 /* WKRichTextSendTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A1990000000000000000F002 /* WKRichTextSendTests.m */; };
 		A2130000000000000000F001 /* WKGroupScanJoinResponseParsingTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A2130000000000000000F002 /* WKGroupScanJoinResponseParsingTests.m */; };
 		A35414ED98745ED690EC26AF /* Pods_LiMaoBase_Example.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4F569B6CE3F4D594F10B3833 /* Pods_LiMaoBase_Example.framework */; };
 		A97C0DE197001001E7B8C3A1 /* WKGroupQRCodeInfoModelTests.m in Sources */ = {isa = PBXBuildFile; fileRef = A97C0DE197001001E7B8C3A2 /* WKGroupQRCodeInfoModelTests.m */; };
@@ -87,6 +88,7 @@
 		873B8AEA1B1F5CCA007FD442 /* Main.storyboard */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = file.storyboard; name = Main.storyboard; path = Base.lproj/Main.storyboard; sourceTree = "<group>"; };
 		996488C1EDA937D964C11C1B /* README.md */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = net.daringfireball.markdown; name = README.md; path = ../README.md; sourceTree = "<group>"; };
 		A1410000000000000000F002 /* WKJoinGroupSuccessHelperTests.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = WKJoinGroupSuccessHelperTests.m; sourceTree = "<group>"; };
+		A1990000000000000000F002 /* WKRichTextSendTests.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = WKRichTextSendTests.m; sourceTree = "<group>"; };
 		A2130000000000000000F002 /* WKGroupScanJoinResponseParsingTests.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = WKGroupScanJoinResponseParsingTests.m; sourceTree = "<group>"; };
 		A97C0DE197001001E7B8C3A2 /* WKGroupQRCodeInfoModelTests.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = WKGroupQRCodeInfoModelTests.m; sourceTree = "<group>"; };
 		B2180000000000000000F002 /* WKConversationListVMSystemBotsTests.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = WKConversationListVMSystemBotsTests.m; sourceTree = "<group>"; };
@@ -209,6 +211,7 @@
 				F1900190F19001000190F202 /* WKUserInfoVMExternalTests.m */,
 				F1900190F19001000190F302 /* WKUserInfoExternalGateTests.m */,
 				A1410000000000000000F002 /* WKJoinGroupSuccessHelperTests.m */,
+				A1990000000000000000F002 /* WKRichTextSendTests.m */,
 				A2130000000000000000F002 /* WKGroupScanJoinResponseParsingTests.m */,
 				B2180000000000000000F002 /* WKConversationListVMSystemBotsTests.m */,
 				B2360000000000000000F002 /* WKSpaceConvSyncCacheRemovalTests.m */,
@@ -551,6 +554,7 @@
 				F1900190F19001000190F201 /* WKUserInfoVMExternalTests.m in Sources */,
 				F1900190F19001000190F301 /* WKUserInfoExternalGateTests.m in Sources */,
 				A1410000000000000000F001 /* WKJoinGroupSuccessHelperTests.m in Sources */,
+				A1990000000000000000F001 /* WKRichTextSendTests.m in Sources */,
 				A2130000000000000000F001 /* WKGroupScanJoinResponseParsingTests.m in Sources */,
 				B2180000000000000000F001 /* WKConversationListVMSystemBotsTests.m in Sources */,
 				B2360000000000000000F001 /* WKSpaceConvSyncCacheRemovalTests.m in Sources */,
--- a/Modules/WuKongBase/Example/Tests/WKRichTextSendTests.m
+++ b/Modules/WuKongBase/Example/Tests/WKRichTextSendTests.m
@@ -1,0 +1,126 @@
+// Copyright 2026 MININGLAMP Technology and the OCTO contributors
+// SPDX-License-Identifier: Apache-2.0
+//
+//  WKRichTextSendTests.m
+//  LiMaoBase_Tests
+//
+//  图文混排 RichText(=14) 发送侧 schema 单测（与接收侧 #16 共用同一份定义）。
+//  对称 web #227 的发送侧用例：
+//   - block 构造（textBlock / imageBlock）
+//   - image block size=0 / name 空 不注入 wire（防 byte-match 污染）
+//   - contentWithBlocks 本地填 plain 占位（image → 非本地化 [图片] wire token）
+//   - encode↔decode 往返：发送构造的 payload 可被接收侧解析回同一 blocks
+//
+
+@import XCTest;
+#import "WKRichTextContent.h"
+
+@interface WKRichTextSendTests : XCTestCase
+@end
+
+@implementation WKRichTextSendTests
+
+#pragma mark - block 构造
+
+- (void)testTextBlock_producesTextBlock {
+    WKRichTextBlock *b = [WKRichTextContent textBlock:@"hi"];
+    XCTAssertEqual(b.type, WKRichTextBlockTypeText);
+    XCTAssertEqualObjects(b.text, @"hi");
+}
+
+- (void)testTextBlock_nilTextFallsBackToEmpty {
+    WKRichTextBlock *b = [WKRichTextContent textBlock:nil];
+    XCTAssertEqual(b.type, WKRichTextBlockTypeText);
+    XCTAssertEqualObjects(b.text, @"");
+}
+
+- (void)testImageBlock_requiredFieldsSet {
+    WKRichTextBlock *b = [WKRichTextContent imageBlock:@"https://x/a.png"
+                                                 width:10
+                                                height:20
+                                                  size:123
+                                                  name:@"a.png"];
+    XCTAssertEqual(b.type, WKRichTextBlockTypeImage);
+    XCTAssertEqualObjects(b.url, @"https://x/a.png");
+    XCTAssertEqual(b.width, 10);
+    XCTAssertEqual(b.height, 20);
+    XCTAssertEqual(b.size, 123);
+    XCTAssertEqualObjects(b.name, @"a.png");
+}
+
+#pragma mark - encode：size=0 / name 空 不注入 wire
+
+- (void)testEncode_imageWithSizeAndName_included {
+    WKRichTextBlock *b = [WKRichTextContent imageBlock:@"https://x/a.png"
+                                                 width:10
+                                                height:20
+                                                  size:123
+                                                  name:@"a.png"];
+    WKRichTextContent *c = [WKRichTextContent contentWithBlocks:@[b]];
+    NSDictionary *wire = [c encodeWithJSON];
+    NSDictionary *img = [wire[@"content"] firstObject];
+    XCTAssertEqualObjects(img[@"type"], @"image");
+    XCTAssertEqualObjects(img[@"url"], @"https://x/a.png");
+    XCTAssertEqualObjects(img[@"width"], @10);
+    XCTAssertEqualObjects(img[@"height"], @20);
+    XCTAssertEqualObjects(img[@"size"], @123);
+    XCTAssertEqualObjects(img[@"name"], @"a.png");
+}
+
+- (void)testEncode_imageSizeZeroAndNameEmpty_notInjected {
+    WKRichTextBlock *b = [WKRichTextContent imageBlock:@"https://x/a.png"
+                                                 width:1
+                                                height:1
+                                                  size:0
+                                                  name:nil];
+    WKRichTextContent *c = [WKRichTextContent contentWithBlocks:@[b]];
+    NSDictionary *wire = [c encodeWithJSON];
+    NSDictionary *img = [wire[@"content"] firstObject];
+    XCTAssertNil(img[@"size"], @"size=0 不应注入 wire");
+    XCTAssertNil(img[@"name"], @"name 空 不应注入 wire");
+    // 必填字段仍在。
+    XCTAssertEqualObjects(img[@"url"], @"https://x/a.png");
+    XCTAssertEqualObjects(img[@"width"], @1);
+    XCTAssertEqualObjects(img[@"height"], @1);
+}
+
+#pragma mark - plain 本地占位（非本地化 wire token）
+
+- (void)testContentWithBlocks_fillsWirePlainPlaceholder {
+    NSArray *blocks = @[
+        [WKRichTextContent textBlock:@"看图："],
+        [WKRichTextContent imageBlock:@"https://x/a.png" width:10 height:20 size:0 name:nil],
+        [WKRichTextContent textBlock:@"怎么样？"],
+    ];
+    WKRichTextContent *c = [WKRichTextContent contentWithBlocks:blocks];
+    // image → 非本地化 [图片] wire token（跨端一致，server #232 Finalize 会覆盖）。
+    XCTAssertEqualObjects(c.plain, @"看图：[图片]怎么样？");
+}
+
+#pragma mark - encode↔decode 往返
+
+- (void)testRoundTrip_encodeThenDecode_preservesBlocks {
+    NSArray *sentBlocks = @[
+        [WKRichTextContent textBlock:@"hello"],
+        [WKRichTextContent imageBlock:@"https://x/a.png" width:10 height:20 size:81920 name:@"a.png"],
+    ];
+    WKRichTextContent *sent = [WKRichTextContent contentWithBlocks:sentBlocks];
+    NSDictionary *wire = [sent encodeWithJSON];
+
+    WKRichTextContent *received = [WKRichTextContent new];
+    [received decodeWithJSON:wire];
+
+    XCTAssertEqual(received.content.count, 2);
+    WKRichTextBlock *t = received.content[0];
+    XCTAssertEqual(t.type, WKRichTextBlockTypeText);
+    XCTAssertEqualObjects(t.text, @"hello");
+    WKRichTextBlock *i = received.content[1];
+    XCTAssertEqual(i.type, WKRichTextBlockTypeImage);
+    XCTAssertEqualObjects(i.url, @"https://x/a.png");
+    XCTAssertEqual(i.width, 10);
+    XCTAssertEqual(i.height, 20);
+    XCTAssertEqual(i.size, 81920);
+    XCTAssertEqualObjects(i.name, @"a.png");
+}
+
+@end
--- a/Modules/WuKongBase/WuKongBase/Assets/Lang/en.lproj/Localizable.strings
+++ b/Modules/WuKongBase/WuKongBase/Assets/Lang/en.lproj/Localizable.strings
@@ -48,6 +48,8 @@
 "清空聊天记录"="Clear Chat History";
 "投诉"="Report";
 "发送成功"="Sent Successfully";
+"发送中"="Sending";
+"发送失败"="Failed to send";
 "删除"="Delete";
 "取消"="Cancel";
 "上午"="AM";
--- a/Modules/WuKongBase/WuKongBase/Classes/Sections/Messages/WKRichTextContent.h
+++ b/Modules/WuKongBase/WuKongBase/Classes/Sections/Messages/WKRichTextContent.h
@@ -5,7 +5,7 @@
 //  图文混排消息（RichText, ContentType=14）。
 //  跨端契约见 octo-lib docs/richtext-blocks-contract.md：content 为有序 block
 //  数组（text / image），数组顺序即图文穿插顺序；plain 为 server 生成的冗余纯
-//  文本，供复制/搜索/摘要复用。Phase 1 只做接收渲染，发送端留 Phase 2。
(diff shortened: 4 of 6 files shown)


lml2468 (dismissed):
This PR is blocked at the review gate because CI is red, so I stopped before code review.

Blocking items:
1. GitHub checks: check-sprint / check-sprint is failing on head 7f12936fad9cce0bdcbf5148197c10216efdc3be, and build / Analyze are still pending. Please fix or rerun CI and request review again after the gate is green.

Non-blocking notes:
- I did not inspect the diff because the requested process says to stop on red CI.

Highlights:
- The branch is currently mergeable, so the blocker is CI state rather than conflicts.

Jerry-Xin (dismissed):
The PR is in scope for `octo-ios` and the implementation is aligned with the existing share-extension and rich-text message architecture. I did not find blocking correctness or security issues.

💬 Non-blocking

🟡 Warning: The new async send orchestration is not covered by tests. `WKRichTextSendTests.m` covers schema construction and encode/decode behavior, but `WKApp.m:2223`-`2367` has the highest risk: image-only+text routing, upload failure aborts, unsafe URL rejection, cleanup, and notification emission. A focused unit/integration seam around this method would catch regressions in the “all or nothing” contract.

🔵 Suggestion: `richTextMimeForExtension:` in `WKApp.m:2371` duplicates MIME logic already present in `WKFileUploadTask.m`, and it omits some image variants such as `heif`. Consider sharing a small MIME helper or extending the mapping so future upload paths do not drift.

✅ Highlights

🔵 The project relevance gate passes: this PR modifies the iOS rich-text send path, schema, localization, and tests within the existing `octo-ios` messaging stack.

🔵 The rich-text schema changes are conservative: `size` and `name` round-trip on decode but serialize only when meaningful in `WKRichTextContent.m:119`-`124`.

🔵 The send path preserves existing behavior unless the share payload is image-only plus non-blank text, as guarded in `WKApp.m:2066`-`2085`.

🔵 The new image URL validation in `WKApp.m:2186`-`2191` correctly restricts generated blocks to `http`/`https` URLs with a host.

Verification note: I could not run Xcode tests because this environment only has Command Line Tools, not full Xcode. I did run `plutil -lint` for the touched localization files, and both passed.

lml2468 (dismissed):
This should be very close, but I found one blocking nullability issue that can break strict warning-as-error builds.

Blocking items:
1. Modules/WuKongBase/WuKongBase/Classes/Sections/Messages/WKRichTextContent.h:53 declares +textBlock: as nonnull under NS_ASSUME_NONNULL_BEGIN, while Modules/WuKongBase/Example/Tests/WKRichTextSendTests.m:32 intentionally calls [WKRichTextContent textBlock:nil] and the implementation handles nil by falling back to an empty string. Clang reports that call as -Wnonnull, and with -Werror it becomes a compile failure. Please either mark the parameter nullable in the public header, for example + (WKRichTextBlock *)textBlock:(nullable NSString *)text;, to match the implementation/test contract, or remove the nil fallback/test if nil is not part of the API.

Non-blocking notes:
- No additional non-blocking notes beyond the existing coverage/MIME drift comments already raised by Jerry-Xin.

Highlights:
- The rich-text schema constructors are now public in WKRichTextContent.h, and size/name encode-decode round-trip cleanly.
- The share-extension path only switches to RichText=14 for image-only shares with non-blank attached text, preserving the existing pure-image and mixed-file flows.
- The send path preserves message atomicity: failures during local metadata read, credentials, PUT upload, or generated URL validation all abort before creating the RichText message.
- Image dimensions are read through ImageIO metadata without full image decode, and the CGImageSourceRef is released correctly.

lml2468 (approved):
Well-structured send-side complement to #16. The code quality is high, and the blocking finding from the first review attempt (nonnull vs nil in textBlock:) does not actually block CI — the build check passed, confirming the project does not enable -Werror for -Wnonnull.

Non-blocking notes:

1. **Nullability contract mismatch (WKRichTextContent.h:53 + WKRichTextSendTests.m:32):** `+textBlock:` is declared nonnull (under `NS_ASSUME_NONNULL_BEGIN`), but the test intentionally passes nil and the implementation handles it with a `?: @""` fallback. This works today because the project does not appear to use `-Werror -Wnonnull`, but if strict warnings are ever enabled, this test will fail to compile. Recommend marking the parameter `nullable` in the header to match the implementation contract, or removing the nil test if nil is not an intended input.

2. Jerry-Xin's notes on MIME dedup and async send test coverage are worth tracking for Phase 2.

Highlights:

- All new public methods (`textBlock:`, `imageBlock:width:height:size:name:`, `contentWithBlocks:`) are properly declared in the `.h` header — the PR #16 header-visibility lesson was applied.
- The `size`/`name` fields encode only when meaningful (size>0, name non-empty), preventing wire-format pollution.
- Atomicity is enforced: any failure in the serial upload chain (missing file, dimension read, upload error, URL validation) aborts the entire message via a one-shot `settled` guard — no partial sends.
- ImageIO metadata reading avoids full image decode, appropriate for share extension memory constraints.
- URL validation (`_WKRichTextIsSafeImageURL`) is symmetric with the receive-side contract.

lml2468 (dismissed):
This should be very close, but I found one blocking nullability issue that can break strict warning-as-error builds.

Blocking items:
1. Modules/WuKongBase/WuKongBase/Classes/Sections/Messages/WKRichTextContent.h:53 declares +textBlock: as nonnull under NS_ASSUME_NONNULL_BEGIN, while Modules/WuKongBase/Example/Tests/WKRichTextSendTests.m:32 intentionally calls [WKRichTextContent textBlock:nil] and the implementation handles nil by falling back to an empty string. Clang reports that call as -Wnonnull, and with -Werror it becomes a compile failure. Please either mark the parameter nullable in the public header, for example + (WKRichTextBlock *)textBlock:(nullable NSString *)text;, to match the implementation/test contract, or remove the nil fallback/test if nil is not part of the API.

Non-blocking notes:
- No additional non-blocking notes beyond the existing coverage/MIME drift comments already raised by Jerry-Xin.

Highlights:
- The rich-text schema constructors are now public in WKRichTextContent.h, and size/name encode-decode round-trip cleanly.
- The share-extension path only switches to RichText=14 for image-only shares with non-blank attached text, preserving the existing pure-image and mixed-file flows.
- The send path preserves message atomicity: failures during local metadata read, credentials, PUT upload, or generated URL validation all abort before creating the RichText message.
- Image dimensions are read through ImageIO metadata without full image decode, and the CGImageSourceRef is released correctly.

Jerry-Xin (dismissed):
Updating my earlier verdict to **Request Changes**: the test target does not compile, so the build is broken.

🔴 Blocking

1. **Compile error: `nil` passed to a `nonnull` parameter.**
   - `WKRichTextContent.h:213` declares `+ (WKRichTextBlock *)textBlock:(NSString *)text;`. This header is inside the `NS_ASSUME_NONNULL_BEGIN` / `NS_ASSUME_NONNULL_END` region, so `text` is implicitly `nonnull`.
   - `WKRichTextSendTests.m` (`testTextBlock_nilTextFallsBackToEmpty`) calls `[WKRichTextContent textBlock:nil];`, which triggers `-Wnonnull` and fails the build under the project's warnings-as-errors setting. This is the CI failure reported on this PR.
   - The implementation in `WKRichTextContent.m:290` is clearly designed to tolerate `nil` ("falls back to empty"), so the intent is a nullable contract.
   - Fix: change the declaration to `+ (WKRichTextBlock *)textBlock:(nullable NSString *)text;` (and the implementation signature to match). This keeps the documented "nil → empty" behavior the test asserts.

Once the test target compiles and CI is green, this is otherwise good to go — see non-blocking items below.

💬 Non-blocking

🟡 The new async send orchestration (`WKApp.m:2223`-`2367`) is the highest-risk logic (image-only+text routing, upload-failure abort, unsafe-URL rejection, cleanup, notification emission) and has no test coverage; `WKRichTextSendTests.m` only covers schema encode/decode. A focused seam around this method would catch regressions in the all-or-nothing contract.

🔵 `richTextMimeForExtension:` (`WKApp.m:2371`) duplicates MIME logic already in `WKFileUploadTask.m` and omits some variants (e.g. `heif`). Consider extracting a shared helper.

✅ Highlights

🔵 Schema changes are conservative: `size`/`name` round-trip on decode but serialize only when meaningful (`WKRichTextContent.m:119`-`124`).
🔵 Send path preserves existing behavior unless the payload is image-only plus non-blank text (`WKApp.m:2066`-`2085`).
🔵 New image URL validation (`WKApp.m:2186`-`2191`) correctly restricts generated blocks to `http`/`https` with a valid host — this closes the SSRF/unsafe-scheme concern raised against the receive side in #16.

yujiawei (commented):
# Code Review — PR #19 (octo-ios)

**Verdict: APPROVED** — no P0/P1 blockers. The send-side closes the four-end RichText(=14) loop cleanly, reuses the receive-side schema without forking, and gets atomicity + URL safety right. Findings below are P2/nits and human-verify items; none block merge.

## 1. Verification summary

| Area | Result |
|---|---|
| Cross-file symbol resolution (no undeclared-method build break) | ✅ verified |
| Schema reuse / wire byte-alignment | ✅ |
| Atomicity (no "send half") | ✅ |
| Image-URL safety gate | ✅ |
| Retain-cycle / race handling | ✅ |
| i18n | ✅ (one nit) |
| Cross-platform image rendering | ⚠️ see P2-1 (human verify) |

### Cross-file resolution (the historical landmine on this module)

The receive-side PR introduced a build break by calling category methods that were never declared in any header. This PR does **not** repeat that — every cross-file call resolves with a matching signature:

- `WKApp.m` forward-declares `sendRichTextMixedImages:extraText:toChannel:` and `richTextMimeForExtension:` inside the class extension (`WKApp.m:121-123`), so the new private methods are visible at their call sites.
- `getUploadCredentialsForPath:type:filename:contentType:fileSize:` — declared `WKAPIClient.h:126`, matches the call at `WKApp.m:2318`.
- `createFileUploadPutTask:fileURL:contentType:contentDisposition:progress:completeCallback:` — declared `WKAPIClient.h:105`, matches `WKApp.m:2333`. The raw absolute path is handled by the `hasPrefix:@"/"` branch (`WKAPIClient.m:321`), so passing `path` directly (no `file://` prefix) is correct.
- `forwardMessage:channel:` (`WKChatManager.h:118`) does save + send of the protocol packet (no media upload) — correct for RichText, whose images already carry URLs (unlike the pure-image path which uses `sendMessage` to trigger upload).
- HUD methods (`showHUD:` / `hideHud` / `showHUDWithHide:`), `_OctoShareAppGroupId()`, `kShareDirName`, and the new `WKRichTextContent` builders all resolve.
- `<ImageIO/ImageIO.h>` is imported and the framework is already linked in this module.
- `AnyPromise` / `WKAPIClient` were already used in `WKApp.m` before this PR, so no new link/import surface.

### Schema & wire format
`encodeWithJSON` emits `content` blocks + `plain`, with `size`/`name` injected only when set (`WKRichTextContent.m`), and the SDK `encode` (`WKMessageContent.m:86`) layers in `type=14`, mention, reply, entities. `encode`↔`decode` round-trips correctly, and the new unit tests cover the optional-field omission and the non-localized wire placeholder. The `[图片]` wire token is correctly kept distinct from the receive-side localized `LLang(@"[图片]")` display placeholder.

### Atomicity — and why it avoids the web counterpart's data-loss bug
The single `settled` flag gates `finish`/`fail` to fire exactly once; `uploadNext` self-captures to stay alive across async callbacks and is nil'd on terminal state to break the retain cycle. Any failure (missing file, unreadable dimensions, upload error, unsafe URL, nil task) routes to `fail` → nothing is sent. There is no editor/draft cleared ahead of an async result here (this is a one-shot share-confirm flow, not a live editor), so the tick-ordering data-loss class that affected the web send path does not apply. On failure the user sees a "Failed to send" HUD and re-shares — same behavior as the existing per-item path.

### Security
`_WKRichTextIsSafeImageURL` enforces `http`/`https` + non-empty host on the post-upload download URL before it enters a block, symmetric with the receive side. Dimensions are read via ImageIO metadata (no full decode) and byte size via file attributes — memory-conservative. Good for a security-sensitive change.

## 2. Findings

### P2-1 (human verify) — raw image bytes are uploaded as-is; HEIC may not render on web/Android receivers
The pure-image share path re-encodes through `WKImageContent` to JPEG/PNG (`WKImageContent.m:132`). This mixed path instead uploads the **raw** shared file with its original extension (`WKApp.m`: `objectKey` and the upload use `path`/`ext` directly). The share extension, for a file-URL source, copies the original file preserving `lastPathComponent` (`ShareExtension/ShareViewController.m:215`), so an Apple-default `.heic` photo stays HEIC. `richTextMimeForExtension:` maps `heic → image/heic`, so the upload itself is correct — but web receivers render images via `` and browsers do not decode HEIC, so a HEIC image shared **with text** could show as broken where the pure-image path (re-encoded to JPEG) would have worked.

This is the one item I'd want a human to confirm before relying on it in production: does the COS/CDN pipeline transcode HEIC server-side, or do the web/Android receive renderers handle it? If neither, consider re-encoding (or at least transcoding HEIC→JPEG) on the mixed path for parity with the pure-image path. Not marked blocking because it depends on deployment-side transcoding I can't observe from the diff.

### P2-2 — no per-message image count / total-size cap
Images upload serially with no upper bound on count or aggregate bytes. Mirrors a note already raised on the receive side; not a regression introduced here, but worth a cap eventually.

### P2-3 (nit) — MIME fallback can mismatch raw bytes for uncommon image types
`richTextMimeForExtension:` returns `image/jpeg` for any unmapped extension. A shared `.bmp`/`.tiff` (both conform to `UTTypeImage`) would be PUT with an `image/jpeg` signed content-type while the bytes are bmp/tiff. The upload succeeds (COS doesn't validate the body), but the stored content-type is wrong. The pure-image path avoids this by re-encoding. Low impact; consider deriving MIME from the actual file or widening the table.

### Nit — comment wording about "share extension killed"
`sendRichTextMixedImages:` runs in the **main app** (`handleShareExtensionData` opens the main app, then `WKForwardSelectVC`'s confirm fires `sendShareFiles:`), not in the extension process. The memory-conservation rationale is still good; the "share extension process killed on multi-image shares" comment is slightly mislabeled.

## 3. Suggestions
- Resolve P2-1 by confirming the transcode/render path for HEIC, or transcode on the mixed path to match the pure-image path.
- Add a modest image-count cap on the aggregate path.
- Tighten the MIME fallback to avoid content-type/body mismatch on uncommon image formats.

## 4. Additional notes
- `hasText` is computed from trimmed text but the text block stores the raw (untrimmed) `extraText` — intentional and harmless (preserves user spacing); just noting the asymmetry.
- The `name` field carries the original filename on the wire for file-URL sources (e.g. `IMG_1234.HEIC`); consistent with the web send path, no action needed.
- i18n: `"发送中"` / `"发送失败"` added to `en.lproj` only; other locales fall back to the key (Chinese) via `LLang`, consistent with this module's existing pattern. No duplicate keys.

Jerry-Xin (approved):
Re-approving after verification. My earlier Request Changes was incorrect — I assumed `-Wnonnull` would fail the build under `-Werror`, but the CI `build` job is green and this project treats `-Wnonnull` as a warning (many existing files already pass `nil` to nonnull params without breaking the build). The reported red check is `check-sprint` (Sprint Board), not code compilation.

In scope for `octo-ios`; implementation aligns with the existing share-extension and rich-text architecture. No blocking correctness or security issues.

💬 Non-blocking

🟡 Nullability annotation mismatch: `WKRichTextContent.h:213` declares `+ (WKRichTextBlock *)textBlock:(NSString *)text;` inside the `NS_ASSUME_NONNULL` region (implicitly nonnull), while the implementation and `testTextBlock_nilTextFallsBackToEmpty` intentionally allow `nil`. Recommend marking the param `nullable` to align the contract and silence `-Wnonnull`. Not a build blocker, but worth fixing.

🟡 The new async send orchestration (`WKApp.m:2223`-`2367`) is the highest-risk logic (image-only+text routing, upload-failure abort, unsafe-URL rejection, cleanup, notification emission) and has no test coverage; `WKRichTextSendTests.m` only covers schema encode/decode. A focused seam would catch regressions in the all-or-nothing contract.

🔵 `richTextMimeForExtension:` (`WKApp.m:2371`) duplicates MIME logic already in `WKFileUploadTask.m` and omits some variants (e.g. `heif`). Consider extracting a shared helper.

✅ Highlights

🔵 Atomicity: any failure in the upload chain aborts the whole message via a one-shot settled guard — no half-sent messages.
🔵 Image URL validation (`WKApp.m:2186`-`2191`) restricts generated blocks to `http`/`https` with a valid host, symmetric with the receive side and closing the SSRF/unsafe-scheme concern from #16.
🔵 Schema changes are conservative: `size`/`name` round-trip on decode but serialize only when meaningful (`WKRichTextContent.m:119`-`124`).
🔵 Memory: ImageIO metadata read instead of full-image decode — appropriate for the share extension's memory limits.

yujiawei (commented):
# Code Review — PR #19 (octo-ios) — Updated

**Updated verdict: CHANGES_REQUESTED.** My earlier comment approved this; I'm revising after a closer pass on nullability annotations, which surfaces a real public-API contract defect (independently flagged by two other reviewers). One blocking fix; the rest is good.

## P1 (blocking) — `+textBlock:` declares `nonnull` but is contractually nullable

`WKRichTextContent.h:53` declares `+ (WKRichTextBlock *)textBlock:(NSString *)text;` inside the `NS_ASSUME_NONNULL_BEGIN`/`END` region (lines 14/69), so `text` is implicitly `nonnull`. But:

- the implementation (`WKRichTextContent.m:140`) deliberately tolerates nil: `block.text = text ?: @"";`
- the shipped test `WKRichTextSendTests.m:32` (`testTextBlock_nilTextFallsBackToEmpty`) calls `[WKRichTextContent textBlock:nil]`.

Passing a literal `nil` to a `nonnull` parameter triggers `-Wnonnull`. The contract is self-contradictory: header says non-null, impl + test assert nil→empty.

**Fix:** make the contract nullable in both header and impl:
```objc
+ (WKRichTextBlock *)textBlock:(nullable NSString *)text;
```
(or drop the nil-tolerance + nil test if nil is not intended to be part of the API). This is a public header other producers import, so the annotation should match real behavior regardless of build flags.

### Severity / CI note (correction to the record)
This is a genuine contract defect worth fixing before merge, but for accuracy: it is **not** the current CI failure on this PR. The `build` job **passed** (it builds the `OctoiOS` scheme, which does not include the `WuKongBase/Example` test target where this test lives), and the `WuKongBase/Example` project does not set `GCC_TREAT_WARNINGS_AS_ERRORS`, so `-Wnonnull` is a warning there, not a hard error. The only red check is `check-sprint` (a Sprint-field process check, unrelated to code). So treat this as a latent contract bug to fix, not a build-is-broken emergency. It still warrants CHANGES_REQUESTED because the public API contract is wrong.

## What's correct (verified)

- **Cross-file resolution** — every cross-file call resolves with a matching signature: `getUploadCredentialsForPath:type:filename:contentType:fileSize:` (`WKAPIClient.h:126`), `createFileUploadPutTask:fileURL:contentType:contentDisposition:progress:completeCallback:` (`WKAPIClient.h:105`, raw absolute path handled by the `hasPrefix:@"/"` branch at `WKAPIClient.m:321`), `forwardMessage:channel:` (`WKChatManager.h:118`, save+send with no media upload — correct for RichText whose images already carry URLs), HUD methods, `_OctoShareAppGroupId()`, `kShareDirName`. Forward decls for the new private methods are present (`WKApp.m:121-123`). `<ImageIO/ImageIO.h>` imported + linked; `AnyPromise`/`WKAPIClient` already used in `WKApp.m`. No recurrence of the receive-side undeclared-method build break.
- **Atomicity** — single `settled` flag gates `finish`/`fail` once; `uploadNext` self-captures across async callbacks and is nil'd on terminal state to break the retain cycle; every failure path (missing file, unreadable dims, cred/PUT error, unsafe URL, nil task) aborts before any message is created. No editor-clear-before-async (this is a one-shot share-confirm flow), so the tick-ordering data-loss class seen on the web send path does not apply here.
- **Security** — `_WKRichTextIsSafeImageURL` gates the post-upload URL to `http`/`https` + non-empty host before it enters a block, symmetric with the receive side; closes the unsafe-scheme concern. Dimensions read via ImageIO metadata (no full decode), byte size via file attributes — memory-conservative; `CGImageSourceRef` released correctly.
- **Schema** — `size`/`name` round-trip on decode but serialize only when meaningful; `encode`↔`decode` clean; non-localized `[图片]` wire token correctly distinct from the receive-side localized display placeholder. SDK `encode` (`WKMessageContent.m:86`) layers `type=14`/mention/reply over `encodeWithJSON`.
- **Routing** — only switches to RichText=14 for image-only shares with non-blank attached text; pure-image / pure-text / any-file paths untouched (zero regression).

## P2 (non-blocking)

- **P2-1 (human verify)** — raw image bytes are uploaded as-is. The pure-image path re-encodes via `WKImageContent` to JPEG/PNG (`WKImageContent.m:132`); this path uploads the original file, and the share extension preserves the original `lastPathComponent` for file-URL sources (`ShareExtension/ShareViewController.m:215`), so an Apple-default `.heic` photo stays HEIC. `richTextMimeForExtension:` maps `heic→image/heic`, so the upload is correct — but web receivers render via `` and browsers don't decode HEIC. A HEIC image shared **with text** could render broken on web/Android where the pure-image path (JPEG) would have worked. Please confirm whether the COS/CDN or the receive renderers transcode HEIC; if not, transcode HEIC→JPEG on this path for parity.
- **P2-2** — no per-message image count / aggregate-size cap on the serial upload loop (mirrors a receive-side note; not a regression here).
- **P2-3 (nit)** — `richTextMimeForExtension:` (`WKApp.m`) falls back to `image/jpeg` for unmapped extensions and omits variants (e.g. `heif`); a shared `.bmp`/`.tiff` would PUT with an `image/jpeg` signed content-type mismatching the bytes. Duplicates MIME logic already in `WKFileUploadTask.m` — consider a shared helper.
- **Nit** — `sendRichTextMixedImages:` runs in the **main app** (`handleShareExtensionData` opens the main app, then the confirm panel fires `sendShareFiles:`), not the extension process; the memory rationale is sound but the "share extension killed" comment is mislabeled.
- **Nit** — the async send orchestration (the highest-risk logic: routing, abort, URL rejection, cleanup, notification) has no test coverage; only schema encode/decode is tested. A seam around the all-or-nothing contract would catch regressions.

## Summary
Solid, conservative send-side that gets the hard parts (cross-file wiring, atomicity, URL safety, schema reuse) right. Blocking only on the `+textBlock:` nullability contract; once that annotation matches behavior, this is good to merge (P2s can follow).

lml2468 (approved):
Re-approving to refresh the stale code-review commit status. No new commits since 7f12936fad9c — original review findings unchanged.

Verdict: APPROVED. The nullable annotation on textBlock: is a valid non-blocking suggestion (CI build passes as-is). All other aspects — atomicity, URL validation, ImageIO metadata, wire-format round-trip — are solid.

PR #24, opened 2026-06-04
feat(richtext): iOS RichText=14 Phase 2 — caption confirm page in main chat (Fixes #20)

### LINKED ISSUE #20: iOS 图文混排 Phase 2：把发送入口从 Share Extension 搬进主聊天输入框（含相册选图丢字 footgun 修复）
## 背景

图文混排 RichText=14 Phase 1 已三端齐活并入 main（iOS 接收 #16、发送 #19；Android 接收；web #227）。Phase 1 目标是把 schema / 原子性 / wire 格式管道验通，这点没问题。本 issue 针对 **iOS 发送侧入口的体验落点** 提 Phase 2 改进。

## 问题

PR #19 把"文字+图片一次发出"的唯一入口放在 **Share Extension 确认面板**，理由是它已有"附件列表 + 附带文字框"、是 web #227 内联编辑器的 iOS 类比。但从体验看这是个死胡同：

1. **主入口不能混排**：99% 的发消息发生在主聊天输入框，而它恰恰不支持配文+附图。用户为了"配张图说句话"得跳出 app 到系统分享面板再选 octo 回来——日常场景没人走这条路。
2. **与同类产品背离**：微信 / iMessage / Telegram / WhatsApp 都是在输入框直接配文+附图一次发出。Share Extension 是"从别的 app 分享进来"的边缘入口，拿它当混排主入口=把边缘当主路。
3. **真正忠实的 web 类比** 不是 Share Extension，而是让聊天输入框本身具备配文+附图能力。选 Share Extension 是工程阻力最小，不是体验最优。

## 附带 footgun（建议一并修，与混排独立）

PR #19 自述："相册选图会立即发送、忽略已输入的文字"。即用户先打了一句话、再去相册选图 → 图发出、**那句话被静默丢弃**。静默丢字体验上比"不能混排"更糟，应单独修复（至少不该丢草稿）。

## 建议方案

**方案 1（最小高价值，建议优先）**：相册选图后、发送前弹一个带文字框的 caption 确认页（微信/TG 标准款），图 + caption 直接打成一条 RichText=14。改动可控，覆盖最高频场景。

**方案 2（最强，Phase 2 正解）**：输入框附件托盘——打字不清空，选中图以缩略图挂在输入框上方，点发送整体打包。真正的"穿插编辑"，天然保留草稿。

## 范围

- Phase 1 维持现状（Share Extension 验管道），不回滚 #16 / #19。
- 本 issue = Phase 2：把混排入口搬进主聊天输入框，优先方案 1；同时修复"相册选图丢已输入文字"的 footgun。

yujiawei:
## What

图文混排 RichText=14 **Phase 2（方案1）**：把发送入口从 Share Extension 确认面板搬进**主聊天输入框**路径，对齐 web #227 基线 / 硬约束 octo-web#237。

相册选图后、发送前弹一个带文字框的 **caption 确认页**（微信/TG 标准款）：用户可补一段描述，点「发送」把图 + caption 直接打成一条 **RichText=14**（复用 #19 的 `sendRichTextMixedImageDatas:extraText:` 发送能力，走 `[context sendMessage:]` 保留全部会话语义）。不写 caption 则纯图发送，wire 零回归。

## How

- **新增 `WKRichTextCaptionViewController`**：横向缩略图条展示已选图 + caption 文字框，取消/发送按钮，键盘自适应底栏。Send/Cancel 回调经 `settled` 守卫只触发一次、dismiss 后回调。
- **`WKMoreItemClickEvent` 相册路径**：选中全为图片时弹 caption 确认页（唯一权威混排入口）；含视频/其它仍走原逐条发送路径，零回归。
- **草稿不丢（硬约束 2）**：进确认页前把输入框已有文本预填进 caption 框并清空输入框；取消 / 发送失败都把这段文本恢复回输入框（与期间新输入合并），文字绝不静默丢。
- **不碰 schema（硬约束 3）**：type=14 wire 复用 #16/#19 的 `WKRichTextContent` blocks 与 #22 的 `shouldAggregateAlbumImagesWithText:` 决策闸门，Phase 2 只动入口/UX。
- caption 页文案本地化（zh-Hans/en）。

## 范围说明

本单交付**方案1**（最小高价值，覆盖最高频场景）。方案2（主聊天输入框附件托盘，打字不清空、缩略图挂输入框上方、整体按真实顺序打包）为终态，留后续单。

Fixes #20


Jerry-Xin (changes requested):
The PR is in scope for `octo-ios`, but the new empty-caption image path can silently send only part of the user’s selected album batch.

**🔴 Blocking**
- 🔴 Critical: [WKMoreItemClickEvent.m](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-z1bd7_bh/repo/Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Panel/MoreItem/WKMoreItemClickEvent.m:149) routes empty captions to `sendAlbumImageDatas:context:`, and [WKMoreItemClickEvent.m](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-z1bd7_bh/repo/Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Panel/MoreItem/WKMoreItemClickEvent.m:249) sends whatever compressed `imageDatas` arrived without validating it against `assetCount`. `WKPhotoBrowser` can return fewer image data entries than selected assets when compression returns nil, while the non-empty caption RichText path correctly rejects `imageDatas.count != assetCount`. With an empty caption, the user can select N images and send fewer than N with no error. Please apply the same asset-count atomicity check to the pure-image branch before sending, and fail visibly instead of partially sending.

**💬 Non-blocking**
- 🟡 Warning: [WKRichTextCaptionViewController.m](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-z1bd7_bh/repo/Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Panel/MoreItem/WKRichTextCaptionViewController.m:220) only invokes cancel/send callbacks from button actions. That is fine with full-screen presentation, but it conflicts with the header contract claiming dismissal gestures invoke cancel. Either implement dismissal-delegate handling or narrow the documented contract.

**✅ Highlights**
- The PR is project-relevant and targets the main chat album-send flow.
- The RichText caption path reuses `context sendMessage:` and preserves chat-send semantics.
- String files lint successfully with `plutil`.

Verification note: I attempted an `xcodebuild` workspace build, but this checkout is missing `OctoConfig.xcconfig` and generated Pods support files, so the build could not reach compile validation.

lml2468 (changes requested):
🔴 P0 blocking issues
- `Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Panel/MoreItem/WKMoreItemClickEvent.m:127` / `:128` / `:157`: the new all-image caption branch performs UIKit and input-panel work directly inside `selectCompressImageBlock` (`inputText`, `inputSetText:@""`, and `presentViewController:`). That callback is not safe to treat as main-thread-only: `WKPhotoBrowser.m:164-173` invokes the selection callback from the GIF compression completion handler, and the existing RichText send path already documents that this callback may arrive from a background compression thread. Selecting a GIF/image batch can therefore mutate the text view and present the caption controller off the main thread, which is undefined UIKit behavior and can crash or corrupt UI state. Please hop to the main queue before reading/updating the conversation input and before presenting the caption controller; keep the send/upload work itself off-main as appropriate.

🟡 P1 non-blocking suggestions
- No additional P1 findings from this pass. I am intentionally not repeating the existing review's pure-image partial-send atomicity finding on the same SHA.

🔵 Highlights / good patterns
- The caption send path reuses the existing RichText upload/send implementation and still sends through the conversation context, preserving the main chat send semantics.
- The new files are included by the `WuKongBase/Classes/**/*` podspec glob, so the controller is wired into the module build rather than being dead code.
- The added localization files pass `plutil -lint`.

📋 Summary verdict
Changes requested. CI was still pending during review, but the main-thread safety issue above is a blocking correctness problem independent of CI.

yujiawei (commented):
# Code Review — PR #24 (octo-ios)

Scope: data-loss / regression review of the RichText=14 Phase 2 caption confirm page in main chat. Focus on the hard constraint that the main-chat text draft must never be silently lost, plus the cross-file producer atomicity and the pure-image send path.

## Verdict

**CHANGES_REQUESTED** — there is a P0 silent data-loss path on the pure-image branch.

## P0 — Original draft silently lost on "clear caption then Send" (pure-image path)

`WKMoreItemClickEvent.m`

On entry to the caption page the existing main-chat draft is captured and the input is cleared:

```objc
NSString *pendingText = [weakContext inputText] ?: @"";   // line 127
[weakContext inputSetText:@""];                            // line 128  draft moved into caption box
```

The draft is only ever restored in two places:
- `onCancel` → `restoreDraft(pendingText)` (line 155)
- the aggregate `onFailure` → `restoreDraft(caption)` (line 147)

The pure-image Send branch restores nothing:

```objc
captionVC.onSend = ^(NSString *caption) {
    if ([WKApp shouldAggregateAlbumImagesWithText:YES assetCount:assetCount pendingText:caption]) {
        [[WKApp shared] sendRichTextMixedImageDatas:images ... onFailure:^{ restoreDraft(caption); }];
    } else {
        // 无 caption：纯图发送
        [weakSelf sendAlbumImageDatas:images context:weakContext];   // line 151 — no restore
    }
};
```

Repro:
1. User has draft "看这个" in the main input.
2. Open album, select images → caption page opens with "看这个" prefilled, main input cleared.
3. User deletes the prefilled caption, taps Send.
4. `caption == ""` → `shouldAggregateAlbumImagesWithText` returns NO → pure-image send runs.
5. Images are sent; the original draft "看这个" is gone and never restored.

This is exactly the footgun the feature was built to prevent. The earlier (pre-Phase-2) code never cleared the input unless it was about to send the aggregated message, so this clear-then-pure-image window is a new regression introduced by moving the clear to the top of the `allImages` block.

### Fix direction (one of)

- Treat the pure-image branch as "draft was not consumed" and restore it: call `restoreDraft(pendingText)` before/after `sendAlbumImageDatas:`. (Caption page already let the user discard their own typing; the *original* draft must not vanish.) Decide product-intent: if a non-empty prefilled caption that the user blanks should be treated as an intentional discard, that needs an explicit confirm — silent loss is not acceptable.
- Or do not clear the main input until the send path is actually known to consume the text; clear only inside the aggregate branch.

Either way, every terminal branch of `onSend`/`onCancel` must account for `pendingText`. Right now 1 of the 3 terminal outcomes drops it.

## Verified — these paths are safe

- **Cancel**: `onCancel` → `restoreDraft(pendingText)`; original draft prepended to any new input. OK.
- **Send with caption, success**: aggregate path consumes `caption` as the message body; draft already lives in caption → becomes the sent message. OK.
- **Send with caption, failure**: `sendRichTextMixedImageDatas` funnels every failure (empty text / no images / no context / `imageDatas.count != assetCount` / temp-file write failure) through `failEarly` → `onFailure` → `restoreDraft(caption)`. Atomic: it never sends a partial subset. OK.
- **Compression drops images (`images.count != assets.count`)**:
  - Aggregate branch: producer's `imageDatas.count != assetCount` gate (`WKApp.m:2450`) → whole message fails → draft restored. Atomicity holds. OK.
  - Decision is made on `assetCount` (user-selected count), not the compressed `images.count`, so a dropped-image compression cannot silently demote to the lossy original loop. OK.
- **`weakContext` nil during restore**: `[weakContext inputText]` returns nil → guarded by `?: @""`; `[weakContext inputSetText:]` to nil is a no-op. No crash. (Draft is lost, but only because the conversation itself is gone — acceptable.)
- **`sendAlbumImageDatas` vs the original per-index loop / `full` flag**: the new method iterates `imageDatas` directly instead of indexing `images[i]` by `assets` index, so there is no out-of-bounds crash if compression dropped images. The `full:NO` hardcode (line 254) vs the original `full:isOriginal` (line 173) is a **no-op on the wire**: `sendImageMessageOfData:` (lines 241–245) never reads `full` — it only calls `WKImageContent initWithData:width:height:` then `sendMessage:`. So the original-photo flag was already dead in both paths; no wire regression. (If you intend to honor original-photo someday, note the flag is dropped here — but that is pre-existing dead behavior, not a Phase-2 regression.)
- **Re-entrancy**: `settled` guard makes `onSend`/`onCancel` mutually exclusive and single-shot; callbacks are copied to a local strong ref before `dismiss`, so they still fire after the VC deallocates. OK.

## P2 / nits

- **Doc vs behavior mismatch (latent data-loss if modal style changes)**: `WKRichTextCaptionViewController.h` documents "下滑关闭回调 onCancel" (swipe-down triggers onCancel), but the VC is `UIModalPresentationFullScreen` (`.m` line 35), which is not interactively dismissable, and there is no `UIAdaptivePresentationControllerDelegate` / `presentationControllerDidDismiss:`. Today that means swipe-down does nothing (safe). But if anyone later switches to `pageSheet`/`formSheet` (the iOS 13+ default), swipe-down would dismiss the page without firing `onCancel` → the cleared draft would be lost with no restore. Either implement `presentationControllerDidDismiss:` to route to the cancel path now, or remove the swipe-down claim from the header doc.
- **`sendAlbumImageDatas` skips empty data** (`data.length == 0 → continue`, line 252) whereas the original loop sent unconditionally. Harmless/defensive, but a behavior delta worth a comment.

## Summary

One blocking issue: the pure-image Send branch silently discards the original main-chat draft when the user blanks the prefilled caption. All other traced paths (cancel, caption-success, caption-failure, compression-drop atomicity, nil context, full-flag, re-entrancy) are safe. Please make every terminal branch of the caption flow account for the captured `pendingText`.

yujiawei (commented):
# Code Review — PR #24 (octo-ios)

RichText=14 Phase 2: a fullscreen caption confirm page presented after an all-image album selection. The user can add a caption (→ single RichText=14 message) or send empty (→ per-image plain send). Scope is tight (+332/-19, 5 files) and the cross-file wiring is clean.

**Verified clean**
- All cross-file producers resolve with matching signatures: `-sendRichTextMixedImageDatas:assetCount:extraText:inContext:onFailure:` (WKApp.h:164), `+shouldAggregateAlbumImagesWithText:assetCount:pendingText:` (WKApp.h:187), `-inputText`/`-inputSetText:`/`-sendMessage:` (WKConversationContext.h), `-topViewController` (WKNavigationManager.h:34). No `#16`-style "no visible @interface" build break.
- Localization keys `添加描述` / `说点什么…` added to both zh-Hans and en; reused `取消`/`发送` already exist.
- `settled` flag correctly guards onSend/onCancel to fire exactly once; callbacks are copied to a local strong ref before `dismissViewControllerAnimated:completion:` so they survive VC deallocation.
- The new `sendAlbumImageDatas:` iterates `imageDatas` directly (not by `assets` index), which removes the out-of-bounds crash risk the old per-index loop would have had if compression dropped images. `full:NO` vs the old `full:isOriginal` is a wire no-op — `sendImageMessageOfData:` never reads the `full` parameter.
- The aggregate send path keeps `[context sendMessage:]` (not `forwardMessage:`), preserving full conversation semantics, and enforces `imageDatas.count == assetCount` atomicity.

## Blocking issues

### P1 — Draft permanently lost when user clears the prefilled caption then sends image-only
`WKMoreItemClickEvent.m:127-128, 142-152`

On entry the main-chat draft is captured into `pendingText` and the input is cleared unconditionally (`inputSetText:@""`, :128). The caption box becomes the *only* carrier of that text. Three terminal outcomes exist, but only two restore the draft:
- `onCancel` → `restoreDraft(pendingText)` ✅ (:155)
- aggregate send failure → `restoreDraft(caption)` ✅ (:147)
- **pure-image send (empty caption) → restores nothing** ❌ (:149-151)

Repro: type a draft `Hello` → pick an all-image album → caption page opens prefilled with `Hello`, input cleared → user deletes the prefilled text → taps **Send** → `caption == ""` → `shouldAggregateAlbumImagesWithText` returns `NO` → `sendAlbumImageDatas:` sends the images and the original `Hello` is gone, no restore, no prompt. This is a regression introduced by hoisting the input-clear to the top of the `allImages` block (on `main` the clear lived only inside the aggregate branch), and it directly violates the feature's stated hard invariant (code comment :119-120) and the header contract (`WKRichTextCaptionViewController.h`: "文字绝不静默丢").

**Fix:** the pure-image else-branch must restore the draft, e.g. call `restoreDraft(pendingText)` before/after `sendAlbumImageDatas:` — pure-image send should carry the same draft-preservation semantics as cancel. (Note `restoreDraft(caption)` would be a no-op here since `caption` is empty; `pendingText` is the correct value.) Alternatively, only clear the input inside the branch that actually consumes the text.

### P1 — Asymmetric atomicity: compression-dropped images silently partial-send on the no-caption path
`WKMoreItemClickEvent.m:140-152, 249-256`

`images.count != assets.count` is reachable: in the non-GIF compression branch, when `jl_compressImageSize:` returns nil nothing is appended to the data array (unlike the GIF branch, which falls back to the original), so a selected image can be silently dropped. The caption page is then built from the (short) `images` array — the user selects 3, sees 2 thumbnails, with no warning. From there the two send paths diverge:
- **with caption** → `sendRichTextMixedImageDatas:` hits its atomic gate (`imageDatas.count != assetCount`) → `failEarly` shows "发送失败" and restores the draft (loud, correct).
- **without caption** → `sendAlbumImageDatas:` has no count check → silently sends the partial set, breaking the "never send only some images" atomicity guarantee that Phase 1 repeatedly asserts.

**Fix:** unify the atomicity check across both paths — validate `images.count == assets.count` before presenting the caption page (fail loudly + preserve draft if not), or add the same count gate inside `sendAlbumImageDatas:`.

## Non-blocking (P2 / nits — fine as follow-ups)

- **P2 — Keyboard frame math can push the input bar off-screen** (`WKRichTextCaptionViewController.m:194`). `constant = overlap > 0 ? -(overlap - safeBottom) : 0` yields a *positive* constant when `0 < overlap < safeBottom` (e.g. iPad hardware-keyboard accessory bar, floating keyboard, split view: overlap=10, safeBottom=34 → +24), moving the bar below the safe area. The common iPhone software-keyboard path (overlap ≫ safeBottom) is fine. Fix: `CGFloat lift = MAX(0, overlap - safeBottom); self.bottomBarConstraint.constant = -lift;` and ideally drive the animation with the keyboard's curve.
- **P2 — Thumbnails eagerly decode all images to full bitmaps** (`WKRichTextCaptionViewController.m:114-124`). `[UIImage imageWithData:]` per image, no downsampling, all resident at once. Bounded by `maxSelectCount = 9` and the ~500KB byte cap (so realistic peak is tens-to-~150MB, not "full-resolution OOM"), but still avoidable — prefer `CGImageSourceCreateThumbnailAtIndex` with `kCGImageSourceThumbnailMaxPixelSize` decoded off-main.
- **P2 — Empty `images` still presents an empty caption page** (`WKMoreItemClickEvent.m:126`). Entry gates only on `assets.count`; if every image fails compression, the page shows a blank thumbnail strip and the no-caption path sends nothing with no feedback. Rare; guard `images.count == 0` before presenting. (Overlaps the P1 atomicity fix.)
- **P2 — Header doc promises swipe-to-dismiss → onCancel, but it's unimplemented** (`WKRichTextCaptionViewController.h:22,36` vs `.m:35`). The VC is `UIModalPresentationFullScreen` (not interactively dismissable) and implements no `presentationControllerDidDismiss:`, so the documented swipe path is currently a no-op — safe today, but a latent draft-loss trap if the modal style ever changes to a sheet. Either drop the doc claim or route `presentationControllerDidDismiss:` through the settled-guarded cancel path.
- **nit** — Placeholder color `lightGrayColor` (`.m:149`) is the only color in the file not dark-mode adapted; rest use `labelColor`/`secondarySystemBackgroundColor`. Prefer `placeholderTextColor`/`tertiaryLabelColor` with a fallback.
- **nit** — Thumbnail strip and caption field lack accessibility labels/hints (buttons & title are auto-labeled by UIButton/UILabel, so VoiceOver isn't zero, but the image strip is unannounced).

## Summary

The cross-file wiring, atomicity gate on the captioned path, callback-once guard, and zero-wire-regression claim all check out. But two confirmed P1s break the feature's own central invariant ("the main-chat draft must never be silently lost" / "never send only some images"): the pure-image Send branch drops the draft, and the no-caption path partial-sends on compression loss while the captioned path correctly fails. Both fixes are small and local. Requesting changes.

lml2468 (approved):
## Incremental Re-review — 369c6df1e1e9 (prev: cc6245e6822f)

All three P0 blockers from the previous round are resolved in this commit:

✅ **P0 main-thread safety (my finding):** The entire UIKit/input-panel block is now wrapped in a `dispatch_block_t presentCaption` and dispatched via `[NSThread isMainThread]` guard + `dispatch_async(dispatch_get_main_queue(), ...)`. GIF compression callbacks that fire off-main now safely hop to the main queue before touching `inputText`, `inputSetText:`, or `presentViewController:`. Correct.

✅ **P0 pure-image atomicity (Jerry-Xin):** `sendAlbumImageDatas:` now takes `assetCount:` and validates `imageDatas.count != assetCount` with a visible HUD failure ("发送失败"). Empty data entries are also caught. Symmetric with the captioned path's gate in `sendRichTextMixedImageDatas:`. Correct.

✅ **P0 draft restore in pure-image path (yujiawei):** The empty-caption branch now calls `restoreDraft(pendingText)` before `sendAlbumImageDatas:…`, so the original draft is always restored to the input box when the user chose not to include text. Aligned with `onCancel` and `onFailure` behavior. Correct.

💬 **Remaining non-blocking (not blocking merge):**
- 🟡 `WKRichTextCaptionViewController.h` header contract still claims dismiss gestures invoke `onCancel`, but the implementation only handles button taps (full-screen presentation makes this moot in practice). Documentation fix can follow.

No new issues found in the incremental diff.

yujiawei:
@codex 三个 CR blocker 已在 commit 8 修复（草稿还原 / count 原子性对称 / UIKit hop main queue，全在 WKMoreItemClickEvent.m 纯图分支），请在新 SHA 上复审。

Jerry-Xin (approved):
Summary: The PR is in scope for `octo-ios` and the implementation looks mergeable; I found no blocking correctness, security, or build-integration issue in the reviewed code.

💬 Non-blocking

- 🟡 Warning: The automated coverage still only tests the old aggregation decision helper, not the new caption-confirm flow. Add tests or a small seam around `WKMoreItemClickEvent.m:126`-`172` for draft restore on cancel, caption send failure, and the no-caption pure-image branch. Existing tests at `Modules/WuKongBase/Example/Tests/WKRichTextSendTests.m:145`-`149` do not exercise the new Phase 2 path.

- 🔵 Suggestion: `WKRichTextCaptionViewController.m:100`-`125` builds thumbnail views by decoding each selected image data directly. With the current max selection count this is probably acceptable, but using generated thumbnail images would reduce peak memory for large originals or GIF-derived data.

✅ Highlights

- The PR correctly keeps UIKit/input access on the main thread in `WKMoreItemClickEvent.m:131`-`171`.
- Draft restoration handles both cancel and failure paths, and avoids overwriting text typed after dismissal by prepending the restored text in `WKMoreItemClickEvent.m:137`-`142`.
- The no-caption path adds an atomic count/data gate before pure-image sends in `WKMoreItemClickEvent.m:266`-`284`, avoiding partial sends when compression drops images.

Verification note: I read the full PR diff with `gh pr diff`. I also attempted an iOS simulator build, but this checkout is missing generated Pods config files and `OctoConfig.xcconfig`, so the build could not proceed far enough to validate compilation.

yujiawei (commented):
# Code Review — PR #24 (octo-ios)

**Verdict: APPROVED** (no P0/P1 blockers). Phase 2 caption confirm page is well-scoped, correctly reuses the #19/#22 send + decision primitives, and the three follow-up fixes (main-thread hop, pure-image atomicity, empty-caption draft restore) are all sound. Comments below are P2/nits only — none block merge.

## 1. Verification results

| Item | Status | Evidence |
|---|---|---|
| Reused send API signature matches | ✅ | `sendRichTextMixedImageDatas:assetCount:extraText:inContext:onFailure:` declared `WKApp.h:164-168`, called `WKMoreItemClickEvent.m:150` |
| Decision gate is pure-function reuse | ✅ | `shouldAggregateAlbumImagesWithText:assetCount:pendingText:` `WKApp.m:2418-2429`, called `WKMoreItemClickEvent.m:148` with `caption` (correct — gate re-evaluated on the post-edit caption, not stale `pendingText`) |
| Context protocol methods exist | ✅ | `inputText` / `inputSetText:` / `channel` / `targetVC` all in `WKConversationContext.h` (113, 98, 30, 212) |
| New VC compiled into target | ✅ | podspec glob `source_files = 'WuKongBase/Classes/**/*'` (`WuKongBase.podspec:44`) covers the new `MoreItem/` files; no pbxproj manual entry needed |
| Localization complete | ✅ | `添加描述` / `说点什么…` added to both `en.lproj` (973-974) and `zh-Hans.lproj` (5-6); reused keys `取消`/`发送`/`发送失败` already present in en (45/52/54) |
| Thread safety | ✅ | `selectCompressImageBlock` may fire on the GIF-compression background thread (`WKPhotoBrowser.m:164-176`); all UIKit access (`inputText`, `inputSetText:`, `presentViewController:`) is hopped to main via the `presentCaption` block + `[NSThread isMainThread]` guard (`WKMoreItemClickEvent.m:167-171`) |
| Pure-image atomicity | ✅ | `sendAlbumImageDatas:assetCount:context:` rejects `count==0 / count!=assetCount / empty NSData` with a visible "发送失败" HUD (`WKMoreItemClickEvent.m:145-151`) — symmetric with the captioned path's gate in `WKApp.m:2450` |
| Draft never silently dropped | ✅ | Draft restored on cancel (162), on empty-caption pure-image send (156), and on captioned send failure (151). Empty-caption case correctly restores the original `pendingText` (which was never sent), not the cleared caption |
| Re-entrancy on dismiss | ✅ | `settled` guard + callback copied to a local strong ref before `dismiss…completion:` so it still fires after the VC deallocs (`WKRichTextCaptionViewController.m:425-433`) |
| Wire / schema untouched | ✅ | type=14 path unchanged; mixed video/other still routes to the original per-item loop (`allImages=NO`, line 173 onward), zero regression |

## 2. Findings

### P2-1 — Presenting the caption VC on `topViewController` while the ZL picker may still be presented
`WKMoreItemClickEvent.m:165` presents on `[WKNavigationManager shared].topViewController`, which returns the **nav stack's** top VC (`WKNavigationManager.m:65-67`), not the currently-presented `ZLPhotoPreviewSheet` modal. If ZL has not finished dismissing its sheet by the time `selectCompressImageBlock` fires (the GIF branch is async — `WKPhotoBrowser.m:164`), `presentViewController:` from a VC that already has a `presentedViewController` will log *"whose view is not in the window hierarchy / which is already presenting"* and the caption page may not appear. The pre-existing code presented only a transient HUD from the same callback, so this path is newly load-bearing. **Recommendation:** confirm in QA that the picker is fully dismissed before the caption page presents (mixed image + GIF selection is the case to test); if not, present from the actual top-most presented controller instead of the nav top.

### P2-2 — Thumbnail strip uses fixed 140pt frames with no empty/oversized-count handling
`buildThumbnails` (`WKRichTextCaptionViewController.m:325-337`) lays out thumbnails with manual frames at a fixed 140pt side. With `maxPreviewCount = 50` (`WKPhotoBrowser.m:80`) the horizontal strip is fine (scrollable), and `imageDatas` is guaranteed non-empty on this branch. No bug — just noting the layout is frame-based inside an Auto-Layout scrollview; acceptable since scrollview content subviews don't require constraints.

### Nit-1 — Header doc mentions "下滑关闭" but modal is full-screen
`WKRichTextCaptionViewController.h:186` documents swipe-to-dismiss, but `modalPresentationStyle = UIModalPresentationFullScreen` (`.m:35`) disables interactive swipe-down. `onCancel` only fires via the Cancel button. Harmless doc/behavior mismatch — either drop the "下滑关闭" wording or switch to a non-fullscreen style if swipe-dismiss is desired (in which case `onCancel` must also be wired via `presentationControllerDidDismiss:` so the draft is still restored on swipe).

### Nit-2 — `full:NO` hardcoded on the pure-image path
`sendAlbumImageDatas:` calls `sendImageMessageOfData:full:NO` (`.m:283`), dropping the `isOriginal` flag the original loop forwarded (line 187). This is benign today because `full:` is currently ignored inside `sendImageMessageOfData:` (`.m:255-258` — the parameter is unused), but if original-image fidelity is ever wired up, this branch would silently always compress. Consider threading `isOriginal` through for future-proofing.

## 3. Recommendations
- Verify P2-1 in QA with a mixed selection that includes a GIF (forces the async compression branch) to confirm present timing.
- Reconcile Nit-1 (doc vs. full-screen) — trivial.

## 4. Additional observations
- The empty-caption decision correctly calls `shouldAggregateAlbumImagesWithText:…pendingText:caption` with the *current caption* rather than the original draft, so clearing the caption and sending yields a clean pure-image send while still restoring the untouched original draft. This asymmetry between "what is sent" (caption) and "what is restored" (pendingText) is the right call and is well-commented (`.m:154-156`).
- Temp-file lifecycle for the captioned path is handled in `WKApp.m:2455-2495` (per-send UUID dir, 10s deferred cleanup) — unchanged by this PR and not a regression surface.

PR #45, opened 2026-06-08
ci: pass TRIAGE_WEBHOOK_URL to octo-issue-notify

lml2468:
The issue notification caller only forwarded OCTO_BOT_TOKEN, so the triage webhook mode was silently skipped and the triage agent was never invoked. Forward the org-level TRIAGE_WEBHOOK_URL secret so triage runs on new issues.

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -18,3 +18,4 @@ jobs:
       event_action: ${{ github.event.action }}
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+      TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}


Jerry-Xin (approved):
This PR is in scope for `Mininglamp-OSS/octo-ios`: it updates repository CI automation for issue handling and fixes the missing secret forwarding needed by the shared issue notification workflow.

✅ Highlights
- `TRIAGE_WEBHOOK_URL` is passed correctly to the reusable workflow in [.github/workflows/octo-issue-feed.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-_f7u_2lr/repo/.github/workflows/octo-issue-feed.yml:21).
- The called reusable workflow declares `TRIAGE_WEBHOOK_URL` as an optional `workflow_call` secret, so this should not break workflow validation.
- No sensitive value is hardcoded; the webhook URL remains sourced from GitHub Actions secrets.
- No tests are required for this narrow workflow wiring change.

No blocking or non-blocking issues found.

yujiawei (approved):
# Code Review — PR #45 (octo-ios)

## Summary

This one-line change forwards the org-level `TRIAGE_WEBHOOK_URL` secret from the `octo-issue-feed.yml` caller into the reusable `octo-issue-notify.yml@v1` workflow. Without it, the reusable workflow received an empty `webhook-url`, so the triage webhook mode was silently skipped — exactly the bug the PR describes. The change is correct, minimal, and safe.

## Verification

I traced the full secret chain end-to-end (caller → reusable workflow → composite action):

- ✅ **Cross-file contract holds.** The called reusable workflow `Mininglamp-OSS/.github/.github/workflows/octo-issue-notify.yml@v1` explicitly declares the secret under `workflow_call.secrets`:
  ```yaml
  secrets:
    OCTO_BOT_TOKEN:
      required: true
    TRIAGE_WEBHOOK_URL:
      required: false
  ```
  So the newly-forwarded secret is a recognized input, not an unused/typo'd name.

- ✅ **The secret is actually consumed.** The reusable workflow passes it to the shared `octo-notify` composite action as `webhook-url: ${{ secrets.TRIAGE_WEBHOOK_URL }}`, which enables webhook mode when non-empty.

- ✅ **Safe when the secret is absent.** `TRIAGE_WEBHOOK_URL` is `required: false` with default `''`. A missing org/repo secret resolves to an empty string, and `octo-notify` treats an empty `webhook-url` as "webhook mode disabled." The change therefore cannot break existing runs even if the secret is not configured — it strictly adds the capability.

- ✅ **Least-privilege secret passing.** Secrets are forwarded explicitly (not `secrets: inherit`), so the reusable workflow only receives the two secrets it needs.

## Security review (PR flagged security-sensitive)

- ✅ **No secret exposure.** The webhook URL is passed as a secret reference and is never interpolated into a `run:` block. In the composite action it is consumed via `env:` (`WEBHOOK_URL`) and POSTed by Python — never built into a shell command or JSON string via `${{ }}`. GitHub will also mask the value in logs.
- ✅ **Capability-secret handling is correct.** The URL (which may itself carry the auth capability) is sourced from a secret, not hardcoded — appropriate given the repo is public.
- ✅ **Minimal egress.** The webhook payload contains only `repo`, `number`, `title`, `url`, `author`, `action`; the issue body is not forwarded. `title`/`author` are control-char sanitized and length-capped, and `repo` is regex-validated before use.

## Notes / non-blocking

- **(nit, pre-existing)** The reusable workflow is pinned to the mutable tag `@v1` rather than a commit SHA. This is consistent with the existing `OCTO_BOT_TOKEN` line and is not introduced by this PR, so it's out of scope here — but org-wide SHA-pinning of third-party/shared workflows would be a worthwhile hardening follow-up.

## For a human to confirm (security-sensitive merge)

- Confirm that `TRIAGE_WEBHOOK_URL` is configured as an org/repo **secret** (not a variable) at the level visible to this repo. If it is unset, the change is a harmless no-op; once set, triage will begin firing on newly-opened issues.

## Verdict

No P0/P1 issues. The change does exactly what it claims, with correct cross-file wiring and safe secret handling. **Approved.**

PR #57, opened 2026-06-27
fix: 未读/搜索/会话列表/@提醒 多项修复 + build 72

DeveKing7:
## 概述
dev_0626 一轮 bug 修复,覆盖未读同步、搜索空间隔离与预览、会话列表、@我 提醒、历史会话拉取、文件预览。build 号 → 72。

### 未读
- **部分读不上报 + 退出后未读莫名清空**:恢复 build 62 的双路径(读完走 store / 部分读走 PUT+本地 DB 双写),修「滚动看一部分退出后角标回弹」「点进不看退出角标消失」。

### 历史会话
- **历史会话拉不到**:对齐 Android `coldStartSyncDone`,进程内首次 conversation/sync 强制 `version=0` 全量拉历史,成功后转增量,logout 复位。

### 搜索
- 聊天记录/文件搜索按**当前空间**过滤(私聊下钻消息级 `space_id`、Bot 走 `WKSpaceBotRegistry`、群聊 `decideChannel`+白名单),修「切空间能搜到别空间内容、点开空白」。
- 会话内「N 条相关聊天记录」详情页逐条预览改为**关键词居中 snippet**。
- 预览行高 15→20,修中文底部被遮。
- 诊断日志开关默认关闭。

### 会话列表
- **去掉群聊 3 天不活跃就隐藏**的逻辑,群不再从列表消失(列表/未读/@提醒口径一致)。

### @我 提醒
- **跨端已读后清 @我 badge**:多端已读 CMD(`WKCMDUnreadClear`,unread=0)时本地把该频道待办 reminder 标 done(只清不创建,无复活风险);reminder sync 的 channel_ids 对齐 web(群+子区)。
- 「直接 @uid / 离线历史 @」的本地**创建**补偿经确认是服务端问题,已**还原**(避免已读重装复活),交服务端解决。

### 文件预览
- 视频/音频走 AVPlayerViewController + 文件打开失败暴露诊断日志(分支既有)。

## 测试
移动端 UI 类改动以真机手测为准(未写单测)。

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lml2468:
## Dispatch

**Assigned to:** qa-engineer
**Stage:** review (QA)
**Reason:** PR #57 opened — fan-out triad review (1/3: QA)

### Scope
PR `dev_0626 → main`: fix: 未读/搜索/会话列表/@提醒 多项修复 + build 72

Changes span: unread sync (partial read / badge rollback), history sync (coldStartSyncDone), search space isolation (per-space filter), conversation list (hidden groups removed), @mention badge clearing (cross-device read), file preview (AVPlayer).

### Deliverables
- Test plan covering each fix area
- Edge case analysis (cross-device sync, space switching, offline scenarios)
- Verdict: CLEARED / CLEARED_WITH_RISK / BLOCKED

Waiting on: your review completion.

lml2468:
## Dispatch

**Assigned to:** security-engineer
**Stage:** review (Security)
**Reason:** PR #57 opened — fan-out triad review (2/3: Security)

### Scope
PR `dev_0626 → main`: fix: 未读/搜索/会话列表/@提醒 多项修复 + build 72

Changes span: unread sync (partial read / badge rollback), history sync (coldStartSyncDone), search space isolation (per-space filter), conversation list (hidden groups removed), @mention badge clearing (cross-device read), file preview (AVPlayer).

### Security focus areas
- Data leakage across spaces (search isolation correctness)
- Sync protocol: partial read PUT+local DB dual-write — race conditions?
- @mention badge clearing via CMD — can it be spoofed to clear others' badges?
- File preview: AVPlayer — any untrusted content handling risks?

### Deliverables
- Security threat analysis
- Verdict: CLEARED / CLEARED_WITH_RISK / BLOCKED

Waiting on: your review completion.

lml2468:
## Dispatch — PR Review Fan-out

| Reviewer | Dispatch ID | Task |
|----------|-------------|------|
| [@qa-engineer](mention://agent/1275f630-47fb-4839-8dea-eee46411e000) | `octo-ios#57\|review\|qa` | QA review — functionality, edge cases, test coverage |
| [@security-engineer](mention://agent/318fd1be-4d36-473c-9adf-bd9cff491149) | `octo-ios#57\|review\|security` | Security review — auth flows, data handling, secrets |
| [@code-reviewer](mention://agent/230fdcb2-adf5-4dad-b3fd-91c2135a7ca2) | `octo-ios#57\|review\|code` | Code review — correctness, patterns, maintainability |

**Waiting on**: All three verdicts required before summary. Post verdict to this PR with `CLEARED` / `BLOCKED` / `CLEARED_WITH_RISK`.

lml2468:
## Dispatch

- **Assigned to:** qa-engineer
- **Stage:** review (QA)
- **Reason:** PR opened — fan-out review三件套之QA评审。检查功能覆盖、边界条件、回归风险。
- **PR:** https://github.com/Mininglamp-OSS/octo-ios/pull/57
- **Sub-issue:** OCTO-203
- **Waiting on:** qa-engineer verdict (CLEARED / CLEARED_WITH_RISK / BLOCKED)

lml2468:
## Dispatch

**Assigned to:** code-reviewer
**Stage:** review (Code)
**Reason:** PR #57 opened — fan-out triad review (3/3: Code)

### Scope
PR `dev_0626 → main`: fix: 未读/搜索/会话列表/@提醒 多项修复 + build 72 (size: XL, +677 lines)

Changes span: unread sync (partial read / badge rollback), history sync (coldStartSyncDone), search space isolation (per-space filter), conversation list (hidden groups removed), @mention badge clearing (cross-device read), file preview (AVPlayer).

### Review focus
- Code quality, naming, Swift conventions
- Correctness of dual-write pattern (store + DB)
- Space isolation logic — boundary cases
- Test coverage assessment (PR notes: manual testing only, no unit tests)
- Dead code / leftover diagnostic logging

### Deliverables
- Inline code review comments
- Verdict: CLEARED / CLEARED_WITH_RISK / BLOCKED

Waiting on: your review completion.

lml2468:
## Dispatch

- **Assigned to:** security-engineer
- **Stage:** review (Security)
- **Reason:** PR opened — fan-out review三件套之安全评审。检查空间隔离越权、认证授权、信息泄露、注入风险。
- **PR:** https://github.com/Mininglamp-OSS/octo-ios/pull/57
- **Sub-issue:** OCTO-205
- **Waiting on:** security-engineer verdict (CLEARED / CLEARED_WITH_RISK / BLOCKED)

lml2468:
## Dispatch

- **Assigned to:** code-reviewer
- **Stage:** review (Code Quality)
- **Reason:** PR opened — fan-out review三件套之代码评审。检查Objective-C代码质量、双路径未读逻辑正确性、搜索隔离一致性、资源管理。
- **PR:** https://github.com/Mininglamp-OSS/octo-ios/pull/57
- **Sub-issue:** OCTO-206
- **Waiting on:** code-reviewer verdict (CLEARED / CLEARED_WITH_RISK / BLOCKED)

Jerry-Xin (dismissed):
Well-scoped, well-documented bug-fix pass covering unread sync, cold-start history sync, space-aware search, conversation-list visibility, @-mention badge clearing, and file-preview diagnostics. Each finding was verified against this PR's actual delta (merge-base..head). Two items initially looked blocking but resolve to non-blocking after delta verification, so this is an approve with should-fix follow-ups.

**🟡 Should-fix (non-blocking)**
- Multi-result search detail page still shows cross-space messages for private channels. At the list level the new `refinePersonResult` correctly filters private/Bot results to the current space, but the `messageCount > 1` branch opens `WKChannelMessageSearchResultVC` with only `{channel, keyword}` (`Modules/WuKongBase/WuKongBase/Classes/Sections/ConversationList/Search/WKGlobalSearchVM.m`, the multi-hit `onClick`). That detail VM re-queries `getMessages:keyword:limit:` and only drops empty `fromUid` with no space filter (`Modules/WuKongBase/WuKongBase/Classes/Sections/ConversationList/Search/WKChannelMessageSearchVM.m`, `tableSectionMaps`). That query is pre-existing and untouched by this PR (this PR only changed the snippet rendering in that file), so it does not block this delta — but since the PR's stated goal is search space isolation, the detail page should get the same message-level filter as `refinePersonResult` in a follow-up, so the displayed list and the "N条相关记录" count stay consistent with the filtered entry.
- File-open-failure diagnostics are surfaced to end users in production. `diagnoseAndShowMissingUrlToast` / `showFileOpenFailureAlertWithTitle` (`Modules/WuKongBase/WuKongBase/Classes/Sections/Messages/WKFileMessageCell.m`) build a detail string containing raw `encodeWithJSON` content (memory + DB), remote URLs, channel/user identifiers and timestamps, then present it in a `UIAlertController` with a copy-to-clipboard action — unconditionally in production (the `WKSearchSpaceTrace` switch only gates the search `NSLog`s, not this alert). This is the user's own message data on an error path, not a cross-user leak, so it is not blocking; but consider gating the raw-payload detail behind `#if DEBUG` / an internal flag and showing a friendlier message (with a hidden long-press to copy) in release builds.

**🟡 Minor**
- `snippetFromText` uses `NSString` index-based `substringWithRange:` / `substringToIndex:` for the 40-char window. Operating on UTF-16 code units can split a surrogate pair or a composed grapheme at the boundary (emoji / 组合字符), producing a momentarily malformed display character. Display-only and low impact, but `rangeOfComposedCharacterSequencesForRange:` would make the cut grapheme-safe.

**✅ Highlights**
- Cold-start conversation sync (`coldStartSyncDone`) is tightly scoped: forces `version=0` full pull on first in-process sync, flips to incremental on success, resets on logout — matches the documented Android behavior and fixes "history conversations not loading" cleanly.
- Cross-device read → @-mention badge clear is clear-only and idempotent (`done:` on already-pending WaitDone reminders, never creating), so no revive risk after re-read.
- Removing the 3-day inactive-group hide is consolidated into the single `isInactiveGroup` predicate so list visibility, unread count, and mention stay consistent.
- Media preview now tears down `AVPlayerViewController` properly, avoiding audio continuing after dismissal.
- Search space isolation for private/Bot DMs via message-level `space_id` + `WKSpaceBotRegistry` (with documented fail-open during list-load races) is a solid, well-reasoned model.

Modules/WuKongBase/WuKongBase/Classes/Sections/Messages/WKFileMessageCell.m
```diff
@@ -162,66 +164,273 @@ - (void)onTap {
...
                 } else {
                     [weakSelf.progressView setProgress:progress];
                 }
             });
         }];
+        return;
     }
+
+    // 把 DIAG 直接弹给用户看 —— 线上排查模式: 用户看不到 console, 但能截图 / 复制 alert。
+    [self showFileOpenFailureAlertWithTitle:LLang(@"文件不存在或正在上传中") detail:diagText];
```
OctoBoooot, line 420: **[major]** The file-open failure path now renders the full diagnostic blob into a user-facing alert in a release build (build 72). `diagnoseAndShowMissingUrlToast:` builds `diagText` containing `fromUid`/`toUid`/`channelId`/`messageId` and the encoded `memContent`/`dbContent` JSON (L304-L390), then passes it as the alert `detail` here with a "复制详情" button.
Failing input: any user taps a file message with empty `remoteUrl` and no local copy → they see internal identifiers + raw content JSON.
The `NSLog([File-onTap][DIAG]…)` already captures all of this for support — suggest gating the alert *detail* behind a debug/diagnostic flag and showing a plain localized message to users, keeping the full log.

Modules/WuKongBase/WuKongBase/Classes/Sections/Messages/WKSafeFilePreviewVC.m
```diff
@@ -212,11 +239,32 @@ + (BOOL)canPreviewExtension:(NSString *)ext {
...
+            @"mp3", @"wav", @"aac", @"m4a", @"caf", @"aiff", @"aif",
         ]];
     });
     return [previewable containsObject:ext.lowercaseString];
 }
 
+// 媒体扩展名子集 (视频 + 音频): 命中后走 setupMediaPlayerInFrame:。AVPlayerViewController
+// 同一套 API 既能放视频也能放音频, 音频文件会自动显示 speaker 占位 + 进度/播放控件,
+// 跟系统"文件"app 行为对齐。
++ (BOOL)mediaExtensions:(NSString *)ext {
```
OctoBoooot, line 254: **[minor]** The media extension list is duplicated between `canPreviewExtension:` (L227) and `mediaExtensions:` here. They agree now, but adding a format to one and not the other silently routes a "previewable" media file back to WebKit — the white-screen path this PR fixes. Derive the media entries in `canPreviewExtension:` from `mediaExtensions:` (or a shared set) so there's one source of truth.

Modules/WuKongBase/WuKongBase/Classes/Sections/ConversationList/Search/WKGlobalSearchVM.m
```diff
@@ -267,20 +284,136 @@ - (void)searchLocalConversations {
...
+    //     时才保留；否则剔除。明略消息不带 space_id，旧逻辑按「向前兼容放行」会把它
+    //     在所有空间漏出（日志实测 inSpace 全是 noSpaceId）。会话列表是按空间从服务端
+    //     同步的，明略-only 私聊不会进 B 的列表，用 anyModelAtChannel 作权威信号对齐。
+    // 再重算 messageCount + 代表消息（保留集合里 orderSeq 最大的一条），保证预览文字
+    // 与点击定位都落在当前空间的消息上。
+    WKConversationListVM *convVM = [WKConversationListVM shared];
+    // 列表尚未加载（race 窗口）→ 无法判定 → 对无 space_id 消息 fail-open 放行，避免误杀。
+    BOOL convListReady = ([convVM conversationCount] > 0);
+    BOOL channelInCurrentSpaceList = ([convVM anyModelAtChannel:channel] != nil);
+    NSArray<WKMessage*> *msgs = [[WKMessageDB shared] getMessages:channel keyword:self.keyword limit:1000];
```
OctoBoooot, line 368: **[question]** `refinePersonResult:` runs `getMessages:channel keyword: limit:1000` once per private-chat result. Is `filterSearchResultsByCurrentSpace:` off the main thread here? If it runs on the search/UI thread, several matching DMs with large histories mean N synchronous 1000-row FMDB scans per search — worth confirming this is dispatched to `ioQueue` like `pullFirst:` is.

OctoBoooot (changes requested):
## Review: fix: 未读/搜索/会话列表/@提醒 多项修复 + build 72 (#57)

**Verdict:** Request changes
CI is red — `check-sprint` is failing because the PR has no linked issue; that's a merge gate and must be fixed first. One major (user-facing diagnostic dump) also needs addressing.

CI is red (1 author-actionable check failing) — that needs fixing first:
- **`check-sprint`** → `No linked issue found. Please add a 'Closes #<issue>' reference to the PR description, then re-run.` This is the real merge blocker and is on the author. CONTRIBUTING also asks PRs to reference the issue.
- `secret-scan (gitleaks)` and `dependency-review` are **not** caused by this diff — see *Out of scope* below. They still gate the merge but need an org/repo-level fix, not a code change here.

### Blockers
- PR-level — `check-sprint` failing: no `Closes #<issue>` in the description. Add the reference and re-run; without it the merge gate stays red regardless of code quality.

### Major
- `WKFileMessageCell.m:L304-L412` — the "文件不存在或正在上传中" failure path now renders the full diagnostic blob (`fromUid`, `toUid`, `channelId`, `messageId`, encoded `memContent`/`dbContent` JSON) into a user-facing `UIAlertController` with a "复制详情" button, in a release build (build 72). Failing input: any user taps a file message whose `remoteUrl` is empty and has no local copy → they see internal identifiers + raw content JSON. The `NSLog([File-onTap][DIAG]…)` already captures everything for support; gate the *alert detail* behind a debug flag and show users a plain localized message, keeping the rich log.

### Minor
- `WKSafeFilePreviewVC.m:L227,L254` — `canPreviewExtension:` and `mediaExtensions:` hardcode the same media-extension list twice. They agree today, but a future addition to one and not the other silently routes a "previewable" media file to WebKit (the white-screen path this PR is fixing). Derive `canPreviewExtension`'s media entries from `mediaExtensions` (or a shared set) so there's one source of truth.

### Questions
- `WKGlobalSearchVM.m:L368` — `refinePersonResult:` runs `getMessages:channel keyword: limit:1000` once per private-chat search result. Is `filterSearchResultsByCurrentSpace:` off the main thread? If it runs on the search/UI thread, a user with several matching DMs and large histories pays N synchronous 1000-row FMDB scans — worth confirming it's dispatched to `ioQueue` like `pullFirst:` is.

### Praise
- The unread double-path restoration (`store` for read-to-bottom, PUT+local-DB for partial) and the `coldStartSyncDone` gate are both documented inline with the exact regression they fix and the Android/web behavior they align to. The explicit fire-and-forget tradeoff note on the partial-read PUT makes these changes reviewable.

### Out of scope (informational)
- `secret-scan (gitleaks)` FAILURE flags `WKCurve25519.swift:20` and `WKRSAUtil.m:126,376` — **neither file is in this diff** (pre-existing findings on a full-repo scan). Needs a baseline allowlist / `.gitleaksignore` at the repo level; not this PR's regression.
- `dependency-review` FAILURE is `Dependency review is not supported on this repository. Please ensure that Dependency graph is enabled` — a repo Settings → Security action, not a code issue.
- No automated tests accompany the change; consistent with the repo's iOS UI layer (manual-test convention stated in the PR). Noting once, not blocking.

lml2468:
## QA Review — octo-ios#57 (dev_0626 → main, build 72)

**评审方式**:静态评审 + 关键路径源码核对(clone dev_0626 验证)。⚠️ 受限说明:本环境无法构建 iOS / 跑真机,PR 自述「无单测,以真机手测为准」。以下结论为**代码级 + 风险级**评审,移动端 UI/同步类改动仍需按下方测试计划真机验证后方可放量。

---

### 一、测试计划(风险驱动,按优先级)

**P0 — 未读双路径(WKConversationView / WKMessageListView)**
- 读完(newMsgCount=0)→ 走 store(markLocalRead),退出/锁屏回前台/重连后角标保持 0。
- 部分读(5 条读 2 条)→ 退出会话再回列表,角标 = 剩余未读(回归用户报告「退出后回弹到 5」)。
- 子区(thread)部分读 / 读完两种,验证 server unread 不再「永远=1」。
- 跳到底(browseToOrderSeq=0 且 newUnread>0)→ 全清。
- **断网下部分读退出**:partial PUT fire-and-forget(无重试队列),验证下次「读完」时 server 被 lastReadSeq 对齐。

**P0 — 历史会话冷启动全量(WKConnectionManager.coldStartSyncDone)**
- 新设备/重装登录:首次 sync 走 version=0,拉回服务端全部历史会话(回归「拉不到历史会话」)。
- 之后转增量(version=DB max),不重复全量。
- logout → 换号登录,再次走全量(coldStartSyncDone 已复位)。
- 弱网/sync 失败:首次失败后 3s 重试仍走全量(coldStartSyncDone 未置位)。
- 大会话量账号:首启全量 payload 体积 / 耗时(性能基线)。

**P0 — 搜索空间隔离(WKGlobalSearchVM)**
- 多空间下:聊天记录/文件搜索仅出当前空间内容;切空间重搜结果隔离。
- 私聊同一人跨 A/B 空间:仅出当前空间命中,点开能定位到当前空间消息(不空白)。
- Bot:NotMember 空间剔除;Member/Unknown 保留。
- 明略(无 space_id)私聊:仅当频道在当前空间会话列表才出。
- 群聊/子区:别空间群被 Skip;白名单未初始化期严格过滤(fail-closed)。
- 详情页逐条预览:关键词居中 snippet,命中在中后部也可见。

**P1 — 会话列表(去 3 天隐藏)**
- 3 天无消息的群仍显示;列表/未读计数/@提醒口径一致(无「显示了群但未读不计入」错位)。

**P1 — @我 跨端清除(WKSystemMessageHandler / WKDataSourceModule)**
- web/其它端读完(unread=0 CMD)→ 本端 @我 badge 清除,且重装不复活。
- 子区 @我:reminder sync channel_ids 含子区,别端读后本端能收 done。

**P1 — 文件 / 视频音频预览(WKFileMessageCell / WKSafeFilePreviewVC)**
- mp4/mov/m4v/3gp、mp3/wav/aac/m4a 走 AVPlayer 原生控件;白名单外(avi/mkv/flv/wmv)走占位页。
- 返回/dismiss/切后台:声音停止(teardown / pause 验证,无「返回后仍在放音」)。
- **静音键打开(硬件 mute switch ON)时播放视频/音频** —— 见缺陷 #1。
- 0 字节本地占位文件 → 当作 missing 走重下。

---

### 二、缺陷(severity = impact × reach)

**#1 [Medium] AVPlayer 未设置 AVAudioSession 类别 → 静音键开启时无声**
`setupMediaPlayerInFrame:`(WKSafeFilePreviewVC.m)直接 `AVPlayer play`,未设 `AVAudioSession` category 为 `AVAudioSessionCategoryPlayback`。默认 ambient 类别会跟随硬件静音键 —— 用户在静音模式下预览视频/音频会**没有声音**(尤其音频文件等于完全失效)。项目里既有的 `WKVideoItemView.m` 正是用 `[[AVAudioSession sharedInstance] setActive:YES]` + route change 处理,本次新路径漏了对齐。`allowsPictureInPicturePlayback=YES` 也需要 playback category 才生效。
- impact: 中(音频文件场景接近功能失效);reach: 全部预览音视频且处于静音模式的用户。
- 建议:进入播放前设 `AVAudioSessionCategoryPlayback`,退出时按需复位,并对齐 WKVideoItemView 的 route change 处理。

**#2 [Medium] 文件打开失败把原始诊断 dump 直接弹给生产用户**
`diagnoseAndShowMissingUrlToast:` / `showFileOpenFailureAlertWithTitle:detail:` 把 `memContent` / `dbContent`(content JSON,含 remoteUrl、fromUid/toUid、clientMsgNo 等)拼成 detail,**无 debug 开关**地 alert 给最终用户,并带「复制详情」写剪贴板。原本是一条干净 toast「文件不存在或正在上传中」,现在变成一屏技术诊断文本。对比同 PR 的搜索埋点已用 `WKSearchSpaceTrace=NO` 网关收口,文件诊断却对生产用户全量可见。
- impact: 中(UX 退化 + 内部字段/URL 暴露给终端用户、可粘贴外传,信息泄露面交 security 复核);reach: 全部命中文件打开失败的用户。
- 建议:用户层仍显示原 toast;诊断 detail 收到 `NSLog` 或同样的调试开关后(如 internal/debug build)再展示。

**#3 [Low-Medium] 部分读未写 reconcile 的 lastReadSeq(作者已知)**
部分读走 PUT + `setConversationUnreadCount`(已确认会落本地 DB,故退出列表重读不回弹,修复有效),但**不更新 WKUnreadStore 的 lastReadSeq/lastLocalReadAt**。因此锁屏后 sync 的「0→>0 防复活」reconcile 分支对 partial 不保护,partial PUT 又是 fire-and-forget。作者注释已承认「server 短期可能不收敛,下次读完时 lastReadSeq 对齐」。
- impact: 低-中(partial 计数在 sync/锁屏竞争下不如「读完」耐久);reach: 部分读 + 弱网/锁屏竞争用户。属设计取舍,验收按上方 P0 断网用例确认可自愈即可。

**#4 [Low] 私聊搜索重算 `getMessages limit:1000` 静默上限**
`refinePersonResult` 对单个私聊按关键词重查上限 1000 条做空间归属重算。极端高频私聊命中 >1000 条时,messageCount 与代表消息(orderSeq 最大)可能不准,且无日志提示截断。reach 很小,记录备查。

**#5 [Low,瞬态] 搜索空间隔离在会话列表未就绪期 fail-open**
冷启动会话列表未加载完(`conversationCount==0`)时,无 space_id 的私聊/Bot 消息走 fail-open 放行 → 该窗口内可能短暂出现跨空间内容。列表加载后重搜即自愈。隔离正确性细节交 security 复核;QA 角度记为瞬态、低 reach。

---

### 三、边界 / 回归关注点
- **冷启动全量 sync 并发**:重连风暴下 syncConversations 若并发,两次都读到 coldStartSyncDone=NO → 都走全量(幂等,仅冗余),无数据正确性问题。
- **跨端 @ 清除竞态**:unread=0 CMD 到达前若有新 @ 消息,新 reminder 可能被一并标 done —— 但该竞态本就由 unread-clear CMD 自身的 setUnread=0 引入,reminder 改动只是与既有(racy 的)未读口径保持一致,未引入新缺陷类。`done:` 实现稳健(写 DB + 重试回传服务端 + 幂等),「只清不创建无复活」成立。
- **去 3 天隐藏**:单点 `isInactiveGroup` 改 return NO,三处口径统一,改动干净。

---

### 四、质量 Verdict:**CLEARED_WITH_RISK**

主干修复逻辑自洽、注释充分,未发现阻断缺陷;双写落 DB、reminder done、冷启动全量复位等关键路径已源码核对成立。放量前需处理/确认:
1. **#1 静音键无声**(建议本轮修,音频预览基本功能受影响);
2. **#2 生产用户可见诊断 dump**(建议收口到调试开关,信息泄露面请 security 一并确认);
3. 按上方 P0 真机用例回归未读双路径(尤其断网部分读)、冷启动历史全量、搜索空间隔离。

#3/#4/#5 为已知取舍 / 低 reach,记录备查,不阻断。

Modules/WuKongBase/WuKongBase/Classes/Sections/ConversationList/Search/WKChannelMessageSearchVM.m
```diff
@@ -53,7 +53,7 @@ @implementation WKChannelMessageSearchVM
            @"name":name,
            @"avatar":[WKAvatarUtil getFullAvatarWIthPath:logo],
            @"keyword": self.keyword?:@"",
-           @"content": [self previewTextForMessage:message],
+           @"content": [self snippetFromText:[self previewTextForMessage:message] keyword:self.keyword maxLength:40],
```
OctoBoooot, line 64: **[blocker]** This detail page bypasses the space filter the list path adds. `tableSectionMaps` (L18) queries `getMessages:self.channel keyword: limit:2000` with no space filtering and renders all of it into the rows built here. Reached via: a multi-result row tap (count>1) in `WKGlobalSearchVM` pushes `WKChannelMessageSearchResultVC` with only channel+keyword, which drives this VM.
Failing input: user in spaces A and B with DMs to person X in both (same `channelId`), searches "salary" from space B. The list shows only B's hits (via `refinePersonResult:`), but tapping "N条相关记录" lists X's space-A "salary" messages too — the cross-space private-message leak this PR aims to fix. `messageCount` shown in the list (filtered) also won't match this page's count.
Direction: apply the same per-message gate as `WKGlobalSearchVM.refinePersonResult:` (`shouldSkipMessageForSpace:` + no-`space_id`→`anyModelAtChannel` for private chats; `WKSpaceBotRegistry` for bots) to the `results` array at L18, or pass the already-filtered set into the detail page instead of re-querying unfiltered.

OctoBoooot (changes requested):
## Re-review (correction): #57 — space-isolation leak in search detail page

**Verdict:** Request changes
Posting a correction to my earlier review: a peer flag (echo banner) named a finding I had not checked — the **per-channel search detail page** — and on verification it is a real space-isolation leak. I missed it round 1; raising it now as a blocker. (CI `check-sprint` is also still red — unchanged.)

### Blockers
- `WKChannelMessageSearchVM.m:L18` — the multi-result detail page bypasses the space filter the list path just added. `tableSectionMaps` runs `getMessages:self.channel keyword: limit:2000` with no space filtering, then renders all of it. Path: list result tap (count>1) → `WKChannelMessageSearchResultVC` (channel+keyword) → this VM.
  Failing input: user in spaces A and B with DMs to person X in both (same `channelId`), searches "salary" from space B. The list correctly shows only B's hits via `refinePersonResult:`, but tapping the "N条相关记录" row lists X's space-A "salary" messages too — the exact cross-space private-message leak this PR set out to fix.
  Direction: apply the same per-message space gate used in `WKGlobalSearchVM.refinePersonResult:` here (`shouldSkipMessageForSpace:` + the no-`space_id`→`anyModelAtChannel` rule for private chats; `WKSpaceBotRegistry` for bots), or pass the already-filtered message set into the detail page instead of re-querying unfiltered. Note `messageCount` shown in the list comes from the filtered set, so it already mismatches what this page displays.

### Note
- I'm not reacting to the echo banner's "✅通过" — that verdict is self-contradictory (passes while listing two 🔴). This blocker stands on the code, independently verified.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-ios PR#57 Deep Investigation: Message-Level Space Filtering for Private Chat Search
**head SHA:** 091556ecef49c856ba8bdc5382899ea74868ec77
**Scope:** 5 targeted investigation questions on the space-filtering data flow in search

---

## Q1: WKMessageDB `getMessages:channel keyword:limit:` — What does it return?

**File:** `Modules/WuKongIMiOSSDK/WuKongIMSDK/Classes/db/WKMessageDB.m`, lines 472-484

**SQL (macro at line 52):**
```sql
SELECT message.* FROM message
LEFT JOIN message_extra ON message.message_id = message_extra.message_id
WHERE message.channel_id = ?
  AND message.channel_type = ?
  AND message.is_deleted = 0
  AND message.content_type <> 99
  AND message.revoke = 0
  AND IFNULL(message_extra.revoke, 0) = 0
  AND (message.searchable_word LIKE ? OR CAST(message.content AS TEXT) LIKE ?)
ORDER BY message.order_seq DESC, message.timestamp DESC
LIMIT ?
```

**Findings:**
- **It DOES keyword search at the SQL level.** The `LIKE '%keyword%'` is applied to both `searchable_word` and `content` columns. This is genuine SQL-level filtering, not a post-fetch filter.
- **Returns up to `limit` WKMessage objects**, ordered by most recent first (`order_seq DESC`).
- Execution is via FMDB's `dbQueue inDatabase:` (serial dispatch queue), so the DB call is thread-safe and the calling thread blocks until completion.

**Can the limit of 1000 miss results?**

Yes, theoretically. If a private chat with one person has >1000 messages matching the keyword, only the 1000 most recent (by order_seq) are returned. Older matching messages beyond position 1000 would be invisible to the space filter.

**Practical impact for `refinePersonResult`:** Low. The purpose is to determine whether any keyword-matching messages belong to the current space. For a typical private chat, 1000 keyword-matching messages is a very high bar (requires 1000+ messages that ALL contain the search term). The more likely concern is that `keptCount` (displayed as `messageCount` in the refined result) could be an undercount if there are >1000 matches. This is cosmetic, not functional.

**Severity:** P2 (cosmetic undercount in extreme edge case). Not a blocker.

---

## Q2: `shouldSkipMessageForSpace:channelType:` — Exact behavior and nil vs empty string

**File:** `Modules/WuKongBase/WuKongBase/Classes/Services/Base/WKSpaceFilter.m`, lines 235-247

**Full implementation:**
```objc
- (BOOL)shouldSkipMessageForSpace:(WKMessage *)message
                      channelType:(uint8_t)channelType {
    if (channelType != WK_PERSON) return NO;
    NSString *current = [self currentSpaceId];
    if (current.length == 0) return NO;
    if (!message || !message.content) return NO;
    id value = message.content.contentDict[kChannelSpaceIdExtraKey]; // @"space_id"
    if (![value isKindOfClass:[NSString class]] || [(NSString *)value length] == 0) {
        return NO; // 历史/旧消息无 space_id → 放行（向前兼容）
    }
    return ![(NSString *)value isEqualToString:current];
}
```

**Decision table:**

| Condition | Returns | Meaning |
|---|---|---|
| `channelType != WK_PERSON` | NO | Groups always pass (handled at channel level) |
| `currentSpaceId` empty | NO | Single-space / no space set, no filtering |
| message or content nil | NO | Defensive guard |
| `space_id` is **nil** (key missing from contentDict) | **NO** | Legacy message, pass through |
| `space_id` is **empty string** `@""` | **NO** | Same as nil (length == 0), pass through |
| `space_id` is non-string type (NSNumber, etc.) | **NO** | Type mismatch guard, pass through |
| `space_id` matches current space | NO | Keep |
| `space_id` differs from current space | **YES** | Skip/drop |

**nil vs empty string: They behave IDENTICALLY.** Both return NO (do not skip). The `length == 0` check covers both nil NSString and `@""`. This is correct and consistent.

**Severity:** No issue found. The method is well-designed with clear fail-open semantics for legacy data.

---

## Q3: WKSpaceBotRegistry `membershipForBotUID:inSpace:` — Unloaded state and fail-open safety

**File:** `Modules/WuKongBase/WuKongBase/Classes/Services/Service/WKSpaceBotRegistry.m`, lines 45-56

**Implementation:**
```objc
- (WKSpaceBotMembership)membershipForBotUID:(NSString *)botUID
                                    inSpace:(NSString *)spaceId {
    if (botUID.length == 0 || spaceId.length == 0) {
        return WKSpaceBotMembershipUnknown;
    }
    @synchronized (self) {
        NSSet<NSString *> *set = self.spaceBotMap[spaceId];
        if (!set) {
            return WKSpaceBotMembershipUnknown;  // Not loaded yet
        }
        return [set containsObject:botUID]
            ? WKSpaceBotMembershipMember
            : WKSpaceBotMembershipNotMember;
    }
}
```

**When registry hasn't loaded:**
Returns `WKSpaceBotMembershipUnknown` (enum value 0). This happens when:
- The `spaceBotMap` dictionary has no entry for the given `spaceId` (cold start or space switch race window).
- Either `botUID` or `spaceId` is nil/empty.

**Fail-open behavior in `refinePersonResult` (line 343-348):**
```objc
if (mem == WKSpaceBotMembershipNotMember) {
    return nil; // Only explicitly NotMember → drop
}
return result; // Member / Unknown (not loaded) → fail-open, keep
```

**Is fail-open safe?**

**Conditionally safe, with a known tradeoff:**

1. **During normal operation:** Safe. The registry loads asynchronously, and `WKSpaceBotRegistryDidLoadNotification` triggers conversation list prune. Search results don't auto-refresh on that notification, but the next user-initiated search will use the loaded registry.

2. **If registry NEVER loads (persistent network failure):** Bots from ALL spaces would appear in search results regardless of space. This is a cross-space information leak — same failure mode the PR is trying to prevent for regular private chats. However, this is mitigated by:
   - Bot conversations are fewer and more visible than person conversations.
   - The conversation list still shows the correct filtered view.
   - The user would need to be in a persistent offline state.

3. **Thread safety:** `@synchronized(self)` provides proper mutual exclusion. The `spaceBotMap` dictionary is only mutated inside `@synchronized` blocks (verified in `loadBotsForSpace:` at lines 115-123). Safe.

**Severity:** P2 (fail-open can leak bot search results across spaces during offline/race window, self-correcting on next search). Not a blocker, but worth noting as a documented tradeoff.

---

## Q4: `anyModelAtChannel:` — Thread safety and search thread

**File:** `Modules/WuKongBase/WuKongBase/Classes/Sections/ConversationList/WKConversationListVM.m`, lines 1398-1406

**Implementation:**
```objc
-(WKConversationWrapModel*) anyModelAtChannel:(WKChannel*) channel {
    WKConversationWrapModel *m = [self modelAtChannel:channel];
    if (m) return m;
    if (channel.channelType == WK_COMMUNITY_TOPIC) {
        for (WKConversationWrapModel *t in self.threadWrapModels) {
            if ([t.channel isEqual:channel]) return t;
        }
    }
    return nil;
}
```

**Locking analysis:**

1. **`modelAtChannel:` (line 1387-1392):** Uses `_conversationsLock` to protect the `channelIndex` dictionary lookup. Thread-safe.
2. **`threadWrapModels` iteration (line 1402-1404):** Does NOT acquire `_conversationsLock`. Iterates `self.threadWrapModels` without synchronization.

**Why the missing lock is acceptable:**

- `threadWrapModels` is a `copy` property (line 35: `@property(nonatomic,copy,readwrite,nullable) NSArray<...> *threadWrapModels`). Every write produces a new immutable NSArray.
- Objective-C fast enumeration (`for...in`) retains the array for the duration of iteration under ARC. Even if another thread replaces `threadWrapModels` mid-iteration, the enumerator continues on the old (retained) snapshot. No crash, no data corruption.
- The `anyModelAtChannel:` path is only reached for `WK_COMMUNITY_TOPIC` channels. In the search context, `refinePersonResult` only calls it for `WK_PERSON` channels (private chats), so the `threadWrapModels` branch is **never reached during search**. The `modelAtChannel:` locked path is the only one that executes.

**What thread does search run on?**

The `search:` method (line 132) is called synchronously — no `dispatch_async`, no background queue. It runs on the **calling thread**, which is the **main thread** (invoked from UI controllers during user interaction). The `conversationCount` property (line 1370: `return self.filteredConversations.count;`) is also unlocked but safe when called from the same main thread.

**Potential concern:** `FMDB dbQueue inDatabase:` (called by `getMessages:keyword:limit:`) dispatches to a serial background queue and blocks the calling thread. This means the main thread blocks during SQL execution. Not a thread safety issue, but a UX concern for large databases (mitigated by the 1000 limit keeping queries fast).

**Severity:** No issue found for the search use case. The `threadWrapModels` branch is unreachable from search (only fires for COMMUNITY_TOPIC). The locked `channelIndex` path is the only one exercised.

---

## Q5: File search `searchLocalFiles` vs chat search `refinePersonResult` — Consistency

**File:** `Modules/WuKongBase/WuKongBase/Classes/Sections/ConversationList/Search/WKGlobalSearchVM.m`, lines 527-580

**Structural difference:**
- `refinePersonResult`: Re-queries DB per channel (`getMessages:channel keyword: limit:1000`), then filters each message.
- `searchLocalFiles`: Uses pre-fetched messages from `searchFileMessagesWithKeyword:keyword limit:50` (cross-channel, file-type only), then filters each message inline.

**Filtering logic comparison:**

| Scenario | `refinePersonResult` | `searchLocalFiles` | Consistent? |
|---|---|---|---|
| Bot, NotMember | Drop (return nil) | Drop (continue) | YES |
| Bot, Member/Unknown | Keep (return result) | Keep (fall through to add) | YES |
| Person, `space_id` != current | Drop (shouldSkip=YES, counted as otherSpace) | Drop (continue at line 556) | YES |
| Person, `space_id` == current | Keep (counted as tagged) | Keep (msgSpace.length > 0, skip inner block) | YES |
| Person, no `space_id`, in list | Keep (counted as noSpaceIdKept) | Keep (inList=true, don't skip) | YES |
| Person, no `space_id`, NOT in list, list ready | Drop (noSpaceIdDropped) | Drop (continue at line 569) | YES |
| Person, no `space_id`, list NOT ready | Keep (convListReady=false, skip gate) | Keep (convListReady=false, skip gate) | YES |

**All 7 scenarios are consistent.** The three-tier filtering (Bot/Person-tagged/Person-noSpaceId) produces identical decisions in both code paths.

**Minor observation (not a bug):** `searchFileMessagesWithKeyword` (SQL at WKMessageDB.m line 59) does a cross-channel search with `limit:50` across ALL file messages, while `getMessages:channel keyword: limit:1000` is per-channel with limit 1000. The file search has a much lower ceiling (50 total vs 1000 per channel), but this is a pre-existing design choice for the file tab and not introduced by this PR.

**Severity:** No issue found. Filtering logic is consistent between chat and file search paths.

---

## Summary

| Question | Finding | Severity |
|---|---|---|
| Q1: getMessages SQL | Real SQL keyword search; 1000 limit may undercount in extreme cases | P2 (cosmetic) |
| Q2: shouldSkipMessageForSpace nil vs empty | Identical behavior (both pass through); well-designed | No issue |
| Q3: BotRegistry fail-open | Unknown → keep; safe with documented offline tradeoff; @synchronized thread-safe | P2 (documented) |
| Q4: anyModelAtChannel thread safety | Locked channelIndex path; unlocked threadWrapModels unreachable from search; search on main thread | No issue |
| Q5: File vs chat filtering consistency | All 7 scenarios produce identical decisions | No issue |

**No P0 or P1 findings.** All identified items are P2 or informational.

---
[Octo-Q] verdict: APPROVE — No P0/P1 findings. The space filtering logic for private chat search is well-designed, thread-safe for its calling context, and consistent between the chat and file search paths. The two P2 items (1000-limit undercount, fail-open during offline) are documented tradeoffs, not blockers.

Jerry-Xin (dismissed):
Solid, well-scoped bug-fix pass (unread sync, cold-start history sync, space-aware search, conversation-list visibility, @-mention badge clearing, file-preview). The application-code changes are reasonable and each finding was checked against this PR's delta (merge-base..head). Switching to REQUEST_CHANGES because the PR currently cannot merge: CI is red and one failure is a PR-owned, actionable blocker.

**🔴 Blocking**
- CI is red. `check-sprint / check-sprint` fails because the PR has no linked sprint issue (org-required check, project-number 2). This is PR-owned and must be resolved by linking the tracking issue before merge.
- `build` (CI) also fails, but the root cause is a Pods/AFNetworking compile error on the x86_64 simulator under Xcode 26.5 — `Use of private header from outside its module: 'netinet6/in6.h'` in `Pods/AFNetworking/AFNetworking/AFNetworkReachabilityManager.m`. This is a dependency/toolchain issue, not introduced by the 15 source files in this PR, but the PR cannot merge while the required `build` check is red, so it needs to go green (Pods bump / module-map fix / CI toolchain pin) regardless of fault.

**🟡 Should-fix (non-blocking on code correctness)**
- File-open-failure diagnostics surfaced to end users in production. `diagnoseAndShowMissingUrlToast` / `showFileOpenFailureAlertWithTitle` (`Modules/WuKongBase/WuKongBase/Classes/Sections/Messages/WKFileMessageCell.m`) build a detail string from raw `encodeWithJSON` content (memory + DB), remote URLs, channel/user identifiers and timestamps, then present it in a `UIAlertController` with copy-to-clipboard — unconditionally in release (the `WKSearchSpaceTrace` switch only gates the search `NSLog`s, not this alert). It is the user's own message data on an error path (not a cross-user leak), so not a correctness blocker; gate the raw-payload detail behind `#if DEBUG` / an internal flag and show a friendlier release message.
- Multi-result search detail page still shows cross-space messages for private channels. The list-level `refinePersonResult` correctly filters private/Bot results to the current space, but the `messageCount > 1` branch opens `WKChannelMessageSearchResultVC` with only `{channel, keyword}` (`...Search/WKGlobalSearchVM.m`, multi-hit onClick). That detail VM re-queries `getMessages:keyword:limit:` and only drops empty `fromUid` with no space filter (`...Search/WKChannelMessageSearchVM.m`, `tableSectionMaps`). That query is pre-existing and untouched by this PR, so it does not block the delta — but since the PR's stated goal is search space isolation, apply the same message-level filter as `refinePersonResult` in a follow-up so the displayed list and the "N条相关记录" count stay consistent.

**🟡 Minor**
- `snippetFromText` uses `NSString` index-based `substringWithRange:` / `substringToIndex:` for the 40-char window. Operating on UTF-16 code units can split a surrogate pair or composed grapheme at the boundary (emoji / 组合字符), producing a momentarily malformed display character. Display-only; `rangeOfComposedCharacterSequencesForRange:` makes the cut grapheme-safe.

**✅ Highlights**
- Cold-start conversation sync (`coldStartSyncDone`) is tightly scoped: forces `version=0` full pull on first in-process sync, flips to incremental on success, resets on logout — matches documented Android behavior.
- Cross-device read → @-mention badge clear is clear-only and idempotent (`done:` on already-pending WaitDone reminders, never creating), so no revive risk after re-read.
- Removing the 3-day inactive-group hide is consolidated into the single `isInactiveGroup` predicate so list visibility, unread count, and mention stay consistent.
- Media preview tears down `AVPlayerViewController` properly, avoiding audio continuing after dismissal.
- Search space isolation for private/Bot DMs via message-level `space_id` + `WKSpaceBotRegistry` (with documented fail-open during list-load races) is a solid, well-reasoned model.

yujiawei (dismissed):
# Code Review — PR #57 (octo-ios)

Reviewed at head `c05c7ac5c6aad70cd2c3bb4d86afef6828c2cfc3` (note: the originally-requested SHA `091556e` was superseded by a follow-up `fix(review)` commit during this review; this review covers the current head). Two independent review passes were run and their findings reconciled below.

## 1. Specification compliance

**Spec: ✅**

The diff implements exactly what the PR description claims, and nothing materially beyond it:

- Unread dual-path restore (`WKConversationView.requestSetUnreadIfNeed`, `WKMessageListView.refreshConversationListNewCount`): three-tier `targetUnread` routing — `0` → `WKUnreadStore markLocalRead` (ack-retry + lastReadSeq), `>0` → `conversationSetUnread` PUT + local DB write. Matches the "partial-read regression" narrative.
- Cold-start full sync gate (`WKConnectionManager.coldStartSyncDone`): forces `version=0` on first sync, then increments; reset on `logout`. Aligns with the stated Android parity.
- Space-filtered search drill-down (`WKGlobalSearchVM.refinePersonResult` + file path), with the follow-up commit also adding the in-conversation detail page filter (`WKChannelMessageSearchVM.filterMessagesByCurrentSpace`) so the "N related results" count and the detail list stay consistent.
- Inactive-group removal collapsed to a single `isInactiveGroup → NO` decision point feeding the three consumers (filter / unread / mention) consistently.
- Cross-device read clears @mention badge (`WKSystemMessageHandler` `unreadClear` handler); reminder-sync `channel_ids` extended to include `WK_COMMUNITY_TOPIC` (`WKDataSourceModule`).
- File preview AVPlayer path + diagnostics; build bump 71 → 72 (both build configs).

Diagnostic trace logging (`WKSearchSpaceTrace`) and the new file-failure detail alert (`WKFileOpenDiagAlert`) are both compile-time gated to `NO`, so no raw payload/PII is surfaced to end users in release. No over-built features, no scope creep.

## 2. Code quality

**Quality: Changes-Requested**

### P1 — blocking

**Synchronous main-thread search fan-out with unbounded autorelease growth** — `WKGlobalSearchVM.m:368` (`refinePersonResult`) and `WKGlobalSearchVM.m:537+` (file path).

The local search path runs entirely on the main thread, per keystroke, with no debounce: `textFieldEditingChanged:` → `changeKeyword:` → `reloadRemoteData` → `requestData` → `searchLocalConversations` (no GCD hop anywhere in `WKBaseTableVC.reloadRemoteData`). This PR adds, inside that synchronous path, a per-channel `[[WKMessageDB shared] getMessages:channel keyword:self.keyword limit:1000]` for **every** person/bot result (capped at 50 by `searchChannelMessageWithKeyword:limit:50`).

Before this PR, `filterSearchResultsByCurrentSpace` did O(N) cheap channel-gate checks. Now it can issue up to ~50 sequential 1000-row `LIKE` queries on the main thread on each keystroke, materializing up to tens of thousands of `WKMessage` objects with **no `@autoreleasepool`** wrapping the loop. In a codebase that is demonstrably ANR-sensitive — this very PR adds `ioQueue` hops in `WKMessageListDataProviderImp` specifically to keep FMDB off the main thread — this is an inconsistent and real interactive-jank / peak-memory regression.

Suggested fix (any one is cheap): move the local-search work off the main thread, add input debounce, and/or wrap the per-channel loop body in `@autoreleasepool` and lower `limit:1000`.

### P2 — advisory (non-blocking)

- **`coldStartSyncDone` lacks a memory barrier** — `WKConnectionManager.m:434/448/518`. It is read/written in `syncConversations:` (network callback thread) with no lock, but reset inside `logout` under `connectStatusLock`. `BOOL` writes are atomic, but without a barrier a `syncConversations:` racing a concurrent `logout`/re-login could read a stale value and skip the cold-start full pull. Practical impact is narrow (account-swap timing), but consider guarding the reset/read consistently.
- **Cross-device `unreadClear` marks *all* pending reminders done** — `WKSystemMessageHandler.m:326`. `getWaitDoneReminder:channel` has no `type` predicate (`WKReminderDB.m:16`). Today only `WKReminderTypeMentionMe` is ever created at runtime (the `WKReminderTypeMemberInvite=1001` creation code is commented out), so this is correct in practice — but it is latently fragile: if another reminder type is ever activated, an `unread==0` signal would wrongly mark it done and upload that state. Prefer the typed query (`getWaitDoneReminders:channel type:WKReminderTypeMentionMe`) to make the intent explicit and future-proof.
- **Reminder server fallback may not truly "pull around"** — `WKMessageListView.m:1385`. `pullAround:orderSeq:0 maxMessageSeq:messageSeq` enters `WKChatManager pullAround` with `aroundOrderSeq==0`, which skips the around-anchor logic and falls through to a plain up-pull from `baseOrderSeq=0`. For an old, uncached @ target this can fetch the wrong window; the recursion then hits `allowServerFallback:NO` and surfaces "原消息不存在". Also, the completion block ignores `error`, so a transient network failure is presented to the user as a permanently-missing message. Consider passing the resolved `orderSeq` as the anchor and short-circuiting on `error`.
- **Empty-vs-unloaded conversation list edge** — `WKGlobalSearchVM.m:366` and `:566`. `convListReady = (conversationCount > 0)` conflates "list not yet loaded" with "loaded but legitimately empty". In a brand-new/empty space, no-`space_id` (legacy) private messages/files fail-open and can leak into that space's search results. Minor, but a `listLoaded` flag would be more precise than a count.
- **Bot membership is channel-level, not per-message** — `WKGlobalSearchVM.m` bot branch. Registry membership proves the bot UID belongs to the current space, not that each message in a multi-space bot DM does. Since bot payloads carry no `space_id` there is no better local signal, so this is an accepted limitation rather than a defect — noting it for visibility.
- **`attributesOfItemAtPath:` on the main thread** — `WKFileMessageCell.m` (onTap + download-complete). Synchronous filesystem stat on the main thread; low-frequency (tap-driven), so minor.

### Verified correct (no action)

- Snippet windowing does **not** crash on a keyword longer than `maxLength`: a successful `rangeOfString` match guarantees `text.length >= location + keyword.length`, so the clamped `start ≤ end ≤ text.length` invariant holds even when `contextRadius` goes negative. The follow-up commit additionally snaps the window to grapheme boundaries (`rangeOfComposedCharacterSequencesForRange:`), removing the emoji/combining-mark split risk.
- File-download completion blocks capture a local `fileContent` + `weakSelf` — no retain cycle through `self`.
- AVPlayer teardown is handled on dismiss, pop, and `viewWillDisappear:` — no lingering audio after close.

## 3. Overall verdict

**REQUEST_CHANGES** — the functional fixes are correct and well-reasoned, but the P1 main-thread search amplification is a genuine interactive-performance regression in an ANR-sensitive path and should be addressed before merge. The P2 items are advisory and need not block once the P1 is resolved.

## Coverage notes

Static review of the diff plus surrounding contracts (DB queries, space filter, reminder done, unread sync, file preview). Not verifiable here without running the app/tests: server-side `conversation/sync` and `clearUnread` semantics, reminder API behavior, runtime ordering of socket/cold-start callbacks, AVPlayback on device, and actual FMDB query latency at scale (the P1 severity assumes large local message tables — empirically confirm with a populated DB).

lml2468:
## PR Review Triad Summary — octo-ios#57

All three reviewer verdicts collected.

| Reviewer | Verdict | Key Findings |
|----------|---------|-------------|
| qa-engineer | **CLEARED_WITH_RISK** | #1 AVPlayer 未设 AudioSession 类别 → 静音键无声; #2 文件打开失败诊断 dump 暴露生产用户; #3 partial 读未更新 lastReadSeq |
| security-engineer | **BLOCKED** 🔴 | `WKChannelMessageSearchVM` 搜索详情页 bypass 空间过滤 → 跨空间私聊消息泄露(本 PR 核心修复目标) |
| code-reviewer | **CLEARED** | 代码质量 OK;建议 follow-up 修 search detail 页隔离 |

**Additional review** (yujiawei): **CHANGES_REQUESTED**
- P1: 主线程搜索 fan-out(~50×1000 行 LIKE/keystroke,无 @autoreleasepool)→ 交互卡顿 / 内存峰值回归
- P2: coldStartSyncDone 缺 memory barrier;unreadClear marks all reminders done;reminder fallback 错误处理

---

### Decision: **NOT PASS** — 需 engineer 修复后重审

阻塞项(必须修):
1. 🔴 **安全**:搜索详情页(`WKChannelMessageSearchVM.tableSectionMaps`)需加 space 过滤,与 `refinePersonResult` 一致
2. 🟡 **QA #1**:AVPlayer 播放前设 `AVAudioSessionCategoryPlayback`,对齐 `WKVideoItemView` route change 处理
3. 🟡 **QA #2**:文件诊断 alert 收口到调试开关(internal/debug build 才展示),生产用户看干净 toast
4. 🟡 **Code P1**(yujiawei):本地搜索 path 离 main thread / 加 debounce / `@autoreleasepool` 至少做一项

CI 非 PR 回归(secret-scan / dependency-review)需 repo 层面处理,不阻塞代码修复。

OctoBoooot (dismissed):
## Re-review (delta 091556ec → c05c7ac5): #57

**Verdict:** Approve
All three of my prior findings are addressed and verified on the delta; the real CI gates (`build`, `check-sprint`) are now green. Approving on my own grounds — this does not override the other open CRs; repo `reviewDecision` stays CHANGES_REQUESTED until those clear.

### Resolved since last review
- ✓ **[blocker] cross-space search detail leak** — `WKChannelMessageSearchVM` now runs `filterMessagesByCurrentSpace:` before rendering, mirroring `WKGlobalSearchVM.refinePersonResult:` (per-message `shouldSkipMessageForSpace:` + no-`space_id`→`anyModelAtChannel`; group/bot/single-space short-circuit). The space-A DM no longer surfaces when searching from space B. Verified.
- ✓ **[major] production diagnostic alert** — `WKFileOpenDiagAlert` (default `NO`) gates the raw-payload alert; release path now shows only the localized title via `showMsg:`, with the full `[File-onTap][DIAG]` dump kept in `NSLog`.
- ✓ **[minor] emoji/combining-char truncation** — both `snippetFromText:` sites snap to `rangeOfComposedCharacterSequencesForRange:` before substringing.

### Minor (non-blocking)
- `WKChannelMessageSearchVM.m` / `WKGlobalSearchVM.m` — the list count comes from a `limit:1000` query (`refinePersonResult:`) while the detail page re-queries at `limit:2000`. For a channel with >1000 keyword hits the "N条相关记录" count and the detail list length can diverge — the same count-consistency this fix is meant to guarantee. Aligning the two limits closes the gap.

### Out of scope (informational)
- `secret-scan (gitleaks)` and `dependency-review` remain red — both pre-existing repo-level issues (gitleaks hits untouched `WKCurve25519.swift`/`WKRSAUtil.m`; dependency-review needs Dependency Graph enabled in repo settings). Not introduced by this diff.
- A peer reviewer flags main-thread synchronous search (detail-page `getMessages:limit:2000`) and a `coldStartSyncDone` visibility race. Both are in code **not** changed by this delta (the 2000-row query and the sync gate predate it), so they don't block these changes — but the search-thread question is worth confirming as a follow-up, since large histories could spike the main thread on detail-page open.

Jerry-Xin (dismissed):
Re-review at head `c05c7ac5c6aad70cd2c3bb4d86afef6828c2cfc3` (prior round `091556e` was superseded by the follow-up fix commit). The three items I flagged last round are all correctly resolved, but a new in-scope P1 performance regression in the same search path keeps this at REQUEST_CHANGES.

## Prior-round items — all FIXED ✅

1. **Cross-space search-detail leak (WKChannelMessageSearchVM)** — FIXED. `tableSectionMaps` now calls `filterMessagesByCurrentSpace:` on the detail render path, using the same `WKSpaceFilter.shouldSkipMessageForSpace:channelType:` + `WKConversationListVM.anyModelAtChannel:` semantics as the list-level `WKGlobalSearchVM.refinePersonResult`. All referenced APIs exist at this head (`WKSpaceFilter shared/currentSpaceId/shouldSkipMessageForSpace:channelType:`, `WKConversationListVM shared/conversationCount/anyModelAtChannel:`, `WKChannelInfoDB queryChannelInfo:`, `WKChannelInfo.robot`). The `if (newResults.count == 0) return nil;` after filtering mirrors the method's own pre-existing top-of-function empty guard (`if(!results || results.count<=0) return nil;`), so it cannot introduce a new crash/blank path. The no-`space_id` fail-open-during-race branch (`convListReady && !channelInCurrentSpaceList` → skip) matches `refinePersonResult` exactly.

2. **File-open diagnostic alert leaking raw payload (WKFileMessageCell)** — FIXED. `static const BOOL WKFileOpenDiagAlert = NO` gates the single chokepoint `showFileOpenFailureAlertWithTitle:detail:`, which in release shows only the friendly `title` toast and returns before constructing the raw-`detail` alert. All three raw-detail call sites (path-mismatch, download-fail, no-remoteUrl) funnel through that one method, so the gate covers every path. The `[File-onTap][DIAG]` NSLog and the DB-recovery re-download (`dbHasUrl` branch) run before the gate and are unaffected.

3. **UTF-16 snippet cut (WKChannelMessageSearchVM + WKGlobalSearchVM snippetFromText)** — FIXED. Both call sites now snap the window to grapheme boundaries via `rangeOfComposedCharacterSequencesForRange:`, and the ellipsis decisions use `safe.location > 0` / `NSMaxRange(safe) < text.length` correctly. `start`/`end` are clamped to `[0, text.length]` before forming the range, so the index math is safe even when `keyword.length > maxLength` (a successful match guarantees `text.length >= location + keyword.length`).

## 🔴 Blocking (new, in-scope, PR-introduced)

**Synchronous main-thread search fan-out with no autoreleasepool** — `WKGlobalSearchVM.m` `refinePersonResult` (~`:368`) and the file path (~`:537`).

On `main`, `filterSearchResultsByCurrentSpace:` did only cheap O(N) channel-gate checks (`isChannelInCurrentSpace`). This PR adds, inside that same synchronous path, a per-result `[[WKMessageDB shared] getMessages:channel keyword:self.keyword limit:1000]` for every person/bot result (capped at 50). The call chain `search:` → `searchLocalConversations` → `filterSearchResultsByCurrentSpace:` → `refinePersonResult:` has no GCD hop to a background queue and runs per keystroke with no debounce, so each keystroke can issue up to ~50 sequential 1000-row `LIKE` queries on the main thread and materialize tens of thousands of `WKMessage` objects with no `@autoreleasepool` wrapping the loop.

This is a real interactive-jank / peak-memory regression in an ANR-sensitive path — and it is internally inconsistent with this very PR, which adds `ioQueue` hops in `WKMessageListDataProviderImp` specifically to keep FMDB off the main thread.

Fix (any one is sufficient): move the local-search drill-down off the main thread (or debounce keyword input), and/or wrap the per-channel loop body in `@autoreleasepool` and lower `limit:1000`.

## 💬 Non-blocking notes

- `convListReady = (conversationCount > 0)` conflates "list not yet loaded" with "loaded but legitimately empty"; in a brand-new/empty space, legacy no-`space_id` private messages can fail-open into that space's results. A `listLoaded` flag would be more precise than a count. Same pattern is now duplicated in `WKChannelMessageSearchVM.filterMessagesByCurrentSpace:`.
- `attributesOfItemAtPath:` (WKFileMessageCell onTap + download-complete) is a synchronous filesystem stat on the main thread; low-frequency (tap-driven), so minor.

## ✅ Highlights

- Detail-page filter reuses the exact list-level semantics, keeping the "N related results" count and the drill-down content consistent.
- Both diagnostic toggles (`WKFileOpenDiagAlert`, `WKSearchSpaceTrace`) are compile-time gated to `NO`, so no raw payload/PII reaches end users in release.
- Grapheme-boundary snippet windowing is the correct fix for the UTF-16 split.

## CI / scope

build = PASS, check-sprint = PASS (issue linked). `secret-scan (gitleaks)` and `dependency-review` are failing but are repo-level, out-of-scope, and not author-fixable here: gitleaks flags pre-existing keys in `WKCurve25519.swift` / `WKRSAUtil.m`, neither of which is in this PR's diff; `dependency-review` fails because the dependency graph is not enabled at the repo level. Not blocking the author for either.

Resolve the P1 and this is good to merge.

Modules/WuKongBase/WuKongBase/Classes/Sections/ConversationList/Search/WKGlobalSearchVM.m
OctoBoooot, line 368: **[blocker]** Per-keystroke synchronous DB fan-out on the main thread. Chain: `textFieldEditingChanged:` (main thread, fires on every character) → `changeKeyword:` → `reloadRemoteData` → `requestData:` → `search:` → `searchLocalConversations` → `filterSearchResultsByCurrentSpace:` → `refinePersonResult:`, which runs this `getMessages:channel keyword: limit:1000` synchronous FMDB scan once per matching private-chat person — no `dispatch_async`, no debounce.
Failing input: a user with N keyword-matching private chats types in the box → each keystroke = N synchronous 1000-row scans on the main thread → UI hitching + memory spike from the materialized arrays. `refinePersonResult:` and this query are added by this PR (verified vs merge-base), so it's a delta regression.
Direction: run `search:`'s local-DB work on a background queue (or the SDK `ioQueue` used by `pullFirst:`) and hop back to main for `reloadData`; debounce `changeKeyword:`; wrap the per-message loop in `@autoreleasepool` to bound peak memory.

OctoBoooot (dismissed):
## Re-review (correction) @c05c7ac5: main-thread search regression

**Verdict:** Request changes
Retracting my earlier APPROVE. A peer reviewer flagged a main-thread performance regression; on tracing the call chain I confirmed it's real and **introduced by this PR**. I had raised this as a question in round 1 and then wrongly dismissed it on the wrong evidence — correcting that now.

### Blocker
- `WKGlobalSearchVM.m:L368` — per-keystroke synchronous DB fan-out on the main thread. Chain: `textFieldEditingChanged:` (main thread, every keystroke) → `changeKeyword:` → `reloadRemoteData` → `requestData:` → `search:` → `searchLocalConversations` → `filterSearchResultsByCurrentSpace:` → `refinePersonResult:`, which runs `getMessages:channel keyword: limit:1000` (synchronous FMDB) **once per matching private-chat person**, with no `dispatch_async` and no debounce.
  Failing input: a user with N private chats matching the keyword types in the search box; each keystroke triggers N synchronous 1000-row scans on the main thread → visible UI hitching and a memory spike from the materialized `WKMessage` arrays. `refinePersonResult:` and its 1000-row query are added by this PR (verified against merge-base), so this is a delta regression, not pre-existing.
  Direction: move `search:`'s local-DB work off the main thread (background queue / the SDK `ioQueue` used by `pullFirst:`) and marshal `reloadData` back to main; debounce `changeKeyword:` so a scan fires on a pause, not every character. An `@autoreleasepool` around the per-message loop bounds the peak.

### Still valid from prior rounds
- ✓ The cross-space leak, diagnostic-alert, and emoji-truncation fixes from the last delta remain correct — those are resolved.

### Note
- My APPROVE @c05c7ac5 has been dismissed to reflect this. The fix here is what stands between this PR and merge, alongside the out-of-scope `secret-scan`/`dependency-review` repo-level CI checks.

Jerry-Xin (approved):
Re-reviewed at head c05c7ac5. The privacy/correctness items from the prior rounds are all verified fixed at this head:
- Cross-space search-detail leak → fixed: `WKChannelMessageSearchVM` now applies `filterMessagesByCurrentSpace:` with the same per-message space logic as the list path.
- File-open diagnostic alert → fixed: gated behind `WKFileOpenDiagAlert` (default off); release shows only a friendly message, raw payload stays in logs.
- UTF-16 snippet cut → fixed: both VMs use `rangeOfComposedCharacterSequencesForRange:`.
- Real CI gates green: `build` SUCCESS, `check-sprint` SUCCESS.

The one remaining concern I had been holding this on — `refinePersonResult` running N synchronous `getMessages limit:1000` queries on the main thread per keystroke (a real, PR-introduced perf regression) — is being **consciously accepted by the maintainer** as a known limitation: the current space-aware search is a temporary transitional implementation, and a brand-new search feature (aligned with the web client) is scheduled to land next week, replacing this path. Given that explicit product decision to ship the interim version and supersede it shortly, I'm withdrawing the change request and approving.

**Recorded for the follow-up search rework (not blocking this PR):**
- Move the space-filter re-query off the main thread (e.g. the SDK `ioQueue` the codebase already uses) and hop back to main only for `reloadData`.
- Debounce `changeKeyword:` so each keystroke doesn't fire a full local search.
- Wrap the per-result loop in `@autoreleasepool` to bound peak memory.

Note: `secret-scan` (gitleaks) and `dependency-review` remain red but are pre-existing repo-level issues, not introduced by this PR (gitleaks hits keys in untouched crypto files; dependency-review = dependency graph disabled at repo level) — out of scope here.

OctoBoooot (approved):
## Re-review (maintainer waiver) @c05c7ac5: #57

**Verdict:** Approve
Retracting my CHANGES_REQUESTED. My sole remaining blocker was the main-thread search-performance regression; the maintainer (@王立涛) has made an explicit, informed decision to accept it — the current search is a temporary stopgap with a full web-aligned rewrite landing next week. No taste-blocking over an owner's deliberate, time-boxed tradeoff. Approving on that basis.

### Accepted as a tradeoff (NOT resolved)
- `WKGlobalSearchVM.m:L368` — `refinePersonResult:` still runs N synchronous `getMessages: limit:1000` scans on the main thread per keystroke (no background dispatch, no debounce). The finding is real and PR-introduced; it is being waived, not fixed. Recorded here so it's not lost: if the planned rewrite slips, this is a live UI-hitch/memory-spike risk on accounts with many keyword-matching private chats. Standing fix if it's ever needed: move the local-DB work off-main, `@autoreleasepool` the loop, debounce input.

### Verified resolved this PR
- ✓ Cross-space search-detail leak — `filterMessagesByCurrentSpace:` added, same per-message gate as the list path.
- ✓ Diagnostic-alert payload — gated behind `WKFileOpenDiagAlert` (default off); release shows only a friendly message.
- ✓ UTF-16 snippet truncation — both sites use `rangeOfComposedCharacterSequencesForRange:`.
- ✓ `build` and `check-sprint` green.

### Out of scope (informational)
- `secret-scan (gitleaks)` and `dependency-review` remain red — pre-existing repo-level issues (gitleaks hits untouched `WKCurve25519.swift`/`WKRSAUtil.m`; dependency-review needs Dependency Graph enabled). Not introduced by this diff; not author TODO.

yujiawei (approved):

PR #58, opened 2026-06-27
ci(octo-pr-review-feed): opt into webhook-only triage

lml2468:
Adds enable_im_notify: false + wires PR_TRIAGE_WEBHOOK_URL secret. Backed by Mininglamp-OSS/.github#81 (which added the input + secret to the reusable). PR fan-out triage agent now files a tracked review issue; the IM ping becomes redundant.

### DIFF
--- a/.github/workflows/octo-pr-review-feed.yml
+++ b/.github/workflows/octo-pr-review-feed.yml
@@ -17,6 +17,8 @@ jobs:
       pr_url: ${{ github.event.pull_request.html_url }}
       pr_author: ${{ github.event.pull_request.user.login }}
       event_action: ${{ github.event.action }}
+      enable_im_notify: false
       project_group_id: '40dd577b7e55490ab278301065f431a3'
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+      PR_TRIAGE_WEBHOOK_URL: ${{ secrets.PR_TRIAGE_WEBHOOK_URL }}


Jerry-Xin (approved):
CI workflow tweak opting `octo-pr-review-feed` into webhook-only triage: sets `enable_im_notify: false` and forwards the `PR_TRIAGE_WEBHOOK_URL` secret to the pinned reusable workflow (`Mininglamp-OSS/.github/.github/workflows/octo-pr-review-feed.yml@v1`). Change is coherent and self-consistent — `sanity / actionlint` passes, the new input/secret line up with the reusable workflow's interface, and `permissions: {}` stays least-privilege.

**✅ Approve (code grounds)**
- Minimal, well-scoped infra change; no application code touched.
- New `enable_im_notify: false` + `PR_TRIAGE_WEBHOOK_URL` secret are syntactically valid and consistent with the existing `with:`/`secrets:` blocks.

**🟡 Minor (non-blocking)**
- Missing trailing newline at end of `.github/workflows/octo-pr-review-feed.yml` (`\ No newline at end of file`). Harmless, but worth adding for POSIX-clean diffs.

**Note — merge gates (not code issues)**
- `check-sprint` is red (org-required): add `Closes #<issue>` to the description to clear it before merge.
- `secret-scan` (gitleaks) and `dependency-review` red are pre-existing repo-level issues, not introduced by this diff (gitleaks hits keys in untouched crypto files; dependency-review = dependency graph disabled at repo level) — out of scope for this PR.

yujiawei (approved):
# Code Review — PR #21 (octo-im)

## Summary

Two-line change to `.github/workflows/octo-pr-review-feed.yml`:

1. Passes `enable_im_notify: false` to the reusable workflow.
2. Forwards `PR_TRIAGE_WEBHOOK_URL` as a secret.

Intent: this repo has moved to an external PR triage agent that files a tracked review issue, so the duplicate IM ping is now redundant and the event should be forwarded webhook-only.

## Spec compliance

The diff matches the stated intent exactly — it adds the one input and the one secret described in the PR body, and nothing else. No missing items, no scope creep, no behavioral deviation.

✅ Spec compliant.

## Cross-file dependency verification

The change is only correct if the referenced reusable workflow actually exposes these contracts at the **pinned ref** (`@v1`), not just on `main`. I verified against `v1`:

- `enable_im_notify` input exists on `Mininglamp-OSS/.github` reusable `@v1` (`type: boolean`, `default: true` — so omitting it preserves current behavior, opting in with `false` is the documented path).
- `PR_TRIAGE_WEBHOOK_URL` secret is declared (`required: false`) on the same ref.
- The reusable wires both correctly: `im-message: ${{ inputs.enable_im_notify && steps.build.outputs.im_message || '' }}` (empty string disables IM mode) and `webhook-url: ${{ secrets.PR_TRIAGE_WEBHOOK_URL }}`.
- The downstream `octo-notify@v1` composite action enables IM mode only when `im-message` is non-empty and webhook mode only when `webhook-url` is non-empty, so with this config the job runs **webhook-only**. ✅

The YAML boolean `false` passed by the caller matches the reusable's `type: boolean` declaration.

## Security review (PR classified security-sensitive)

- Trigger is `pull_request_target` (`ready_for_review`, `review_requested`) with `permissions: {}` and `if: !draft`. No PR code is checked out or executed — only `github.event.pull_request.*` metadata is forwarded, so the standard `pull_request_target` secret-exfiltration risk does not apply here.
- All field values are consumed via `env:` and the webhook payload is built with `json.dumps` over env vars in the reusable (no `${{ }}` string interpolation into shell/JSON), so there is no script-injection path from attacker-controlled PR title/author.
- The webhook URL is correctly sourced from a secret rather than hardcoded (the repo is public, and the URL path is itself a capability).

No new attack surface is introduced by this PR.

## Operational note (non-blocking)

With `enable_im_notify: false`, the notify job's only output channel is now the webhook. The `octo-notify` action exits non-zero (`exit 2`) if invoked with neither an IM message nor a webhook URL. I confirmed `PR_TRIAGE_WEBHOOK_URL` is currently present as a repo secret, so the job will operate in webhook-only mode as intended. Worth being aware that if that secret is ever cleared/rotated-to-empty, every PR `ready_for_review`/`review_requested` event will fail this job (rather than silently no-op), since the IM fallback is now disabled. No action required while the secret is set.

## Nit (non-blocking)

The file ends without a trailing newline (`\ No newline at end of file`). Harmless for YAML; adding one would keep the file POSIX-clean.

## Verdict

No correctness, security, or functional blockers. The cross-file contract is satisfied at the pinned `@v1` ref and the required secret is in place.

**Approved.**

PR #76, opened 2026-08-16
fix(docs): hand off viewer space on mobile open

### LINKED ISSUE #77: Pass the native viewer Space when iOS opens standalone Docs
## Problem
Standalone Docs pages opened from WKWebView do not receive the logged-in client's current Space at document start. The first authorization request can therefore use missing or stale WebView Space state.

## Expected behavior
- For exact trusted standalone Docs routes, inject the native current Space at document start in the main frame.
- Remove stale WKWebView Space state when the native selection is empty.
- Do not add a forgeable URL parameter or a general-purpose Space bridge.
- Keep non-Docs and untrusted-origin pages unchanged.

wwwzzzcccc:
## Summary
- inject the native current Space at document start for trusted standalone Docs routes
- clear stale WKWebView Space state when the native selection is empty
- enforce the same origin/path allowlist in native code and in the injected script
- add focused route, origin, stale-state, and script-escaping tests

## Linked issue
Closes #77

## How verified
- independent review: P0/P1 = 0
- Foundation-only helper compiled with ARC and `-Wall -Wextra -Werror`
- helper behavior harness and generated JavaScript execution passed
- `project.pbxproj` validation and `git diff --check` passed
- full XCTest/build was not run locally because full Xcode/CocoaPods is unavailable; CI must run platform tests and build

### What does this change actually do?
Before, standalone Docs pages opened in WKWebView without the viewer's native current Space. Now exact trusted Docs routes receive one document-start, main-frame-only Space snapshot in `localStorage.currentSpaceId`; nil or empty input removes stale state.

### What could break?
WKWebView startup for standalone Doc, PPT, and presentation routes. The allowlist, effective-port check, and runtime recheck keep unrelated and external pages untouched.

### How do you know it works?
The production helper compiles and passes behavior probes for origin/path gating, stale removal, and script escaping. The test target registration and project file validate; independent review found no P0/P1 issue. CI remains required for full XCTest/build evidence.

### DIFF
--- a/Modules/WuKongBase/Example/LiMaoBase.xcodeproj/project.pbxproj
+++ b/Modules/WuKongBase/Example/LiMaoBase.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		D0C500000000000000000001 /* WKDocsViewerSpaceHandoffTests.m in Sources */ = {isa = PBXBuildFile; fileRef = D0C500000000000000000002 /* WKDocsViewerSpaceHandoffTests.m */; };
 		520BE45CD0E8F8F4E2EF21AC /* WKFollowSidebarTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 3B6EA9F5D6FE167093BF2D24 /* WKFollowSidebarTests.m */; };
 		6003F58E195388D20070C39A /* Foundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 6003F58D195388D20070C39A /* Foundation.framework */; };
 		6003F590195388D20070C39A /* CoreGraphics.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 6003F58F195388D20070C39A /* CoreGraphics.framework */; };
@@ -57,6 +58,7 @@
 /* End PBXContainerItemProxy section */
 
 /* Begin PBXFileReference section */
+		D0C500000000000000000002 /* WKDocsViewerSpaceHandoffTests.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = WKDocsViewerSpaceHandoffTests.m; sourceTree = "<group>"; };
 		150887FD5F014CC653E028F5 /* LICENSE */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text; name = LICENSE; path = ../LICENSE; sourceTree = "<group>"; };
 		31E97EA6FA728C7061776910 /* LiMaoBase.podspec */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text; name = LiMaoBase.podspec; path = ../LiMaoBase.podspec; sourceTree = "<group>"; xcLanguageSpecificationIdentifier = xcode.lang.ruby; };
 		344B3E144C0E72E5FC31C6B1 /* Pods-LiMaoBase_Example.debug.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-LiMaoBase_Example.debug.xcconfig"; path = "Target Support Files/Pods-LiMaoBase_Example/Pods-LiMaoBase_Example.debug.xcconfig"; sourceTree = "<group>"; };
@@ -200,6 +202,7 @@
 			isa = PBXGroup;
 			children = (
 				6003F5BB195388D20070C39A /* Tests.m */,
+				D0C500000000000000000002 /* WKDocsViewerSpaceHandoffTests.m */,
 				85EB8C00E7C49A151B670678 /* WKExternalViewerResolverTests.m */,
 				A97C0DE197001001E7B8C3A2 /* WKGroupQRCodeInfoModelTests.m */,
 				B2CA62921B187BCA0DFC746C /* WKMentionUserCellTests.m */,
@@ -543,6 +546,7 @@
 			buildActionMask = 2147483647;
 			files = (
 				6003F5BC195388D20070C39A /* Tests.m in Sources */,
+				D0C500000000000000000001 /* WKDocsViewerSpaceHandoffTests.m in Sources */,
 				93FBC126DB4B04BC75AF04E2 /* WKExternalViewerResolverTests.m in Sources */,
 				A97C0DE197001001E7B8C3A1 /* WKGroupQRCodeInfoModelTests.m in Sources */,
 				B49640EA6FC4C273CD92871A /* WKMentionUserCellTests.m in Sources */,
--- a/Modules/WuKongBase/Example/Tests/WKDocsViewerSpaceHandoffTests.m
+++ b/Modules/WuKongBase/Example/Tests/WKDocsViewerSpaceHandoffTests.m
@@ -1,0 +1,63 @@
+// Copyright 2026 MININGLAMP Technology and the OCTO contributors
+// SPDX-License-Identifier: Apache-2.0
+
+@import XCTest;
+#import "WKDocsViewerSpaceHandoff.h"
+
+@interface WKDocsViewerSpaceHandoffTests : XCTestCase
+@end
+
+@implementation WKDocsViewerSpaceHandoffTests
+
+- (void)testAllowsSameOriginViewerRoutesAndEffectiveDefaultPort {
+    NSString *base = @"https://octo.example/api/v1/";
+    XCTAssertTrue(WKIsTrustedDocsViewerURL([NSURL URLWithString:@"https://octo.example/d/doc-1"], base));
+    XCTAssertTrue(WKIsTrustedDocsViewerURL([NSURL URLWithString:@"https://octo.example:443/ppt/d/deck-1/"], base));
+    XCTAssertTrue(WKIsTrustedDocsViewerURL([NSURL URLWithString:@"https://octo.example/docs/deck-1/present?slide=2"], base));
+}
+
+- (void)testRejectsOriginSpoofingAndPortMismatch {
+    NSString *base = @"https://octo.example/api/v1/";
+    XCTAssertFalse(WKIsTrustedDocsViewerURL([NSURL URLWithString:@"https://octo.example.evil/d/doc-1"], base));
+    XCTAssertFalse(WKIsTrustedDocsViewerURL([NSURL URLWithString:@"https://octo.example@evil.example/d/doc-1"], base));
+    XCTAssertFalse(WKIsTrustedDocsViewerURL([NSURL URLWithString:@"http://octo.example/d/doc-1"], base));
+    XCTAssertFalse(WKIsTrustedDocsViewerURL([NSURL URLWithString:@"https://octo.example:8443/d/doc-1"], base));
+}
+
+- (void)testRejectsNonViewerAndNearMatchPaths {
+    NSString *base = @"https://octo.example/api/v1/";
+    NSArray<NSString *> *URLs = @[
+        @"https://octo.example/", @"https://octo.example/docs", @"https://octo.example/d/",
+        @"https://octo.example/d/doc-1/edit", @"https://octo.example/ppt/d/deck-1/edit",
+        @"https://octo.example/docs/deck-1", @"https://octo.example/docs/deck-1/present/extra"
+    ];
+    for (NSString *value in URLs) {
+        XCTAssertFalse(WKIsTrustedDocsViewerURL([NSURL URLWithString:value], base), @"%@", value);
+    }
+}
+
+- (void)testScriptSetsEscapedSnapshot {
+    NSString *lineSeparator = [NSString stringWithFormat:@"%C", (unichar)0x2028];
+    NSString *paragraphSeparator = [NSString stringWithFormat:@"%C", (unichar)0x2029];
+    NSString *space = [NSString stringWithFormat:@"space'\\\"\n%@%@</script>", lineSeparator, paragraphSeparator];
+    NSString *script = WKDocsViewerSpaceJavaScript(space, @"https://octo.example/api/v1");
+    XCTAssertTrue([script containsString:@"localStorage.setItem('currentSpaceId'"]);
+    XCTAssertTrue([script containsString:@"space'\\\\\\\""]);
+    XCTAssertTrue([script containsString:@"\\n"]);
+    XCTAssertTrue([script containsString:@"\\u2028"]);
+    XCTAssertTrue([script containsString:@"\\u2029"]);
+    XCTAssertTrue([script containsString:@"<\\/script>"]);
+    XCTAssertFalse([script containsString:lineSeparator]);
+    XCTAssertFalse([script containsString:paragraphSeparator]);
+}
+
+- (void)testScriptRemovesForNilOrEmptySpace {
+    NSString *remove = WKDocsViewerSpaceJavaScript(nil, @"https://octo.example/api/v1");
+    XCTAssertTrue([remove containsString:@"localStorage.removeItem('currentSpaceId')"]);
+    XCTAssertTrue([remove containsString:@"p===\"https\""]);
+    XCTAssertTrue([remove containsString:@"h===\"octo.example\""]);
+    XCTAssertTrue([remove containsString:@"n===443"]);
+    XCTAssertEqualObjects(WKDocsViewerSpaceJavaScript(@"", @"https://octo.example/api/v1"), remove);
+}
+
+@end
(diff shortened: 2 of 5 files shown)


Jerry-Xin (approved):
The PR is repository-relevant and safely implements native Space handoff for trusted standalone Docs routes.

💬 Non-blocking

- 🔵 Suggestion — `Modules/WuKongBase/Example/Tests/WKDocsViewerSpaceHandoffTests.m:42`: Consider executing the generated JavaScript in a real JS runtime within the test. The current substring assertions do not catch syntax or behavioral regressions in the emitted script.

✅ Highlights

- Native (`WKIsTrustedDocsViewerURL`) and injected-JavaScript checks both enforce matching scheme, host, effective port, and the explicit viewer/presentation route allowlist — dual-layer defense that also protects against a trusted URL redirecting to an untrusted origin before document-start.
- Nil and empty Space IDs correctly remove stale origin-scoped `localStorage` state, preventing pooled WebView content from reusing another Space's snapshot.
- `WKJSONString` (NSJSONSerialization + U+2028/U+2029 handling) safely escapes quotes, backslashes, newlines, forward slashes, and Unicode line/paragraph separators.
- Injection occurs at document start and only in the main frame (`forMainFrameOnly:YES`), so the Space value is never exposed to cross-origin subframes.
- Origin-spoofing (`host.evil`, userinfo `@evil`), scheme/port mismatch, near-match routes, stale-state removal, and script escaping all receive focused test coverage.
- The project file validates successfully and `git diff --check` reports no errors.

yujiawei (approved):
# Code Review — PR #76 (octo-ios)

Reviewed against head SHA `78c5d027e2752a9669c23e1259412fd5eaa170df` (branch `fix/mobile-docs-viewer-space`, base `main`). 5 files, +155/-0.

---

## 1. Spec compliance

**Spec: ✅**

There is no linked issue, so the change is reviewed against its own stated goal: inject the native current Space into `localStorage.currentSpaceId` at document-start for trusted standalone Docs routes, and clear stale state when the native selection is empty.

- **Missing:** none. The stated goal is implemented end to end for the primary path.
- **Out of scope additions:** none blocking. One item to confirm, see Q1 below.
- **Deviations:** none blocking. One contract-alignment item, see P2-5.

I verified the contract this change depends on is real, rather than assuming it:

- `localStorage.currentSpaceId` is genuinely what the web bootstrap reads. The standalone/enterprise deep-link branch in `apps/web/src/Layout/index.tsx` does `const cachedSpaceId = localStorage.getItem("currentSpaceId")` and assigns it to `WKApp.shared.currentSpaceId` when the in-memory value is unset (three call sites around lines 454, 477 and 561). So a document-start write does land before the app shell reads it.
- `currentSpaceId` in `NSUserDefaults` is the native source of truth: `WKSpaceFilter.m:95` reads exactly `@"currentSpaceId"` from `NSUserDefaults` and that value is what feeds the `X-Space-Id` request header (`WKApp.m:690`). After this change the WKWebView-hosted app and the native client agree on the Space, which is the actual point of the fix.
- The primary user path really does flow through this VC: a doc link tapped in a chat bubble goes to `WKWebViewVC` (`WKTextMessageCell.m:2434-2441`), and doc links are minted as a direct `${origin}/d/${docId}` with no redirect hop, so the direct-open case that this change covers is the dominant one.

Build/packaging sanity checks I ran rather than assumed:

- `WuKongBase.podspec:44` is `s.source_files = 'WuKongBase/Classes/**/*'`, so the new `.h`/`.m` are picked up with no manifest edit; no other Xcode project enumerates these sources.
- The two new `project.pbxproj` object IDs are internally consistent and collide with nothing (`D0C5…0001` appears twice: `PBXBuildFile` + `Sources`; `D0C5…0002` three times: `PBXFileReference` + group children + `fileRef`).
- The test target pattern matches the existing precedent (`WKExternalViewerResolverTests.m` imports a plain C-function header the same way and links through the host app), so the `FOUNDATION_EXPORT` functions resolve.
- Ordering at the call site is correct: `viewDidLoad` adds `self.webView` as a subview (line 76), which triggers the lazy getter that calls `addUserScript:` (line 291), and `loadRequest:` only happens afterwards (line 106). The script is installed before the first navigation.
- Mutating `webView.configuration.userContentController` after webview construction is the same pattern the pre-existing `navigator.language` script in this exact method already relies on, so it is not a new risk.

## 2. Code quality

**Quality: Approved** (no P0/P1; P2 items below are follow-ups, not merge blockers)

The design is fail-closed in the right places: an exact origin match plus an explicit route allowlist natively, *and* an independent origin/path recheck inside the injected script, so a redirect to a foreign origin cannot receive the value even though the script is installed on the webview for its whole lifetime. `WKJSONString` is a correct and nil-safe way to produce a JS string literal (array-wrap, strip the brackets, then post-escape U+2028/U+2029, which JSON emits raw), and the `[^/]` inside the JS regex literal is valid ECMAScript, so the generated program parses. `forMainFrameOnly:YES` is the right choice here.

### P2-1 — A redirect or in-page transition into a viewer route gets no handoff

`WKWebViewVC.m:832`

```objc
if (WKIsTrustedDocsViewerURL(self.url, [WKApp shared].config.apiBaseUrl)) {
```

The decision to install the script is made from the URL the VC was asked to load. If that URL is a shortlink, an auth bounce, or any same-origin page that later redirects to `/d/:docId`, no script is installed and the final Docs page reads whatever stale value the origin already had — which is exactly the state this change sets out to clear.

Not a blocker: before this change *no* page received the handoff, so this is an uncovered path rather than a regression, and the canonical link builder emits a direct `/d/:docId` with no redirect hop.

Two viable fixes: re-evaluate on `decidePolicyForNavigationAction`, or drop the native gate entirely and lean on the runtime recheck already inside the script (the payload self-gates on exact origin + path, so it is safe on non-viewer pages). The second is simpler but widens the set of webviews carrying the script — worth a deliberate decision either way.

### P2-2 — The Space snapshot is frozen at webview construction

`WKWebViewVC.m:833`

```objc
NSString *spaceId = [[NSUserDefaults standardUserDefaults] stringForKey:@"currentSpaceId"];
```

The value is captured once and baked into the script source. Every subsequent main-frame document-start in that webview (reload, forward navigation, expired-session reload) rewrites the *original* value. If the native selection changed in between, the page is bootstrapped into the previous Space.

Reachability is low in the current UI — the VC is pushed on the navigation stack and deallocated on pop, and the Space switcher is not reachable while it is on screen — and the failure mode is no worse than today's stale value, which is why this is P2 and not a blocker. Reading `NSUserDefaults` at document-start time instead (e.g. via a `WKScriptMessageHandler` round-trip, or by re-installing the script per navigation) would close it.

### P2-3 — Percent-encoded separators slip past the route allowlist

`WKDocsViewerSpaceHandoff.m:16`

```objc
@"^/(?:d/[^/]+|ppt/d/[^/]+|docs/[^/]+/present)/?$"
```

`percentEncodedPath` for `https://octo.example/d/doc-1%2Fedit` is `/d/doc-1%2Fedit`, and `[^/]+` matches `doc-1%2Fedit` because there is no literal slash. `location.pathname` keeps `%2F` encoded too, so the runtime recheck agrees and the write happens. A proxy or server that decodes separators before routing would then serve `/d/doc-1/edit` — a route the tests deliberately reject (`testRejectsNonViewerAndNearMatchPaths`).

Impact is bounded (still same-origin, still an octo page), but the allowlist does not mean what the tests claim it means. Rejecting `%2f`/`%2F` in the segment, or tightening the segment charset (see P2-5), closes it.

### P2-4 — Sub-path deployments silently get no handoff

`WKDocsViewerSpaceHandoff.m:33`

```objc
return sameOrigin && WKIsDocsViewerPath(candidate.percentEncodedPath);
```

With `apiBaseUrl = https://octo.example/octo/api/v1/`, the origin comparison passes but the root-anchored route regex rejects `/octo/d/doc-1`, so the feature no-ops with no diagnostic. `WKApp.serverOrigin` (`WKApp.m:1007`) deliberately preserves everything before `/api/`, which is evidence that path-prefixed deployments are a real shape in this codebase. Failing closed is the right default; deriving the prefix from `apiBaseUrl` the way `serverOrigin` already does would make it work.

### P2-5 — The route allowlist diverges from the web-side contract, and is duplicated

`WKDocsViewerSpaceHandoff.m:16` and `WKDocsViewerSpaceHandoff.m:55`

The canonical web-side matcher for these same routes (`packages/dmworkbase/src/features/notifications/documentScene.ts`) uses `^/d/[A-Za-z0-9_-]{1,128}$` and `^/ppt/d/[A-Za-z0-9_-]{1,128}$`. This change uses `[^/]+`, which is materially looser (it is what makes P2-3 possible), and it is written twice — once as an `NSRegularExpression` pattern and once as a hand-escaped JS regex literal — with no test asserting the two agree. Any future edit to one is silently free to drift from the other. Aligning the segment charset with the web contract and adding a test that both engines classify the same corpus identically would fix both halves.

### P2-6 — Nil-port comparison is not defensive

`WKDocsViewerSpaceHandoff.m:32`

```objc
[WKEffectivePort(candidate) isEqualToNumber:WKEffectivePort(server)]
```

`WKEffectivePort` returns `nil` for any scheme other than http/https with no explicit port. Because the scheme equality check short-circuits first, a `nil` right-hand operand is only reachable when *both* sides use a custom scheme and only the candidate carries an explicit port — effectively unreachable in production, which is why this is P2. Still, `-isEqualToNumber:` has no documented nil contract, and `-isEqual:` is a free, unambiguous substitute.

### P2-7 — Test coverage is textual, not behavioural

`WKDocsViewerSpaceHandoffTests.m:39-52`

The script tests assert on substrings of the generated source (`XCTAssertTrue([script containsString:@"localStorage.setItem('currentSpaceId'"])`). They never execute the program against representative `location` values or assert the resulting storage state, so a syntactically broken or semantically inverted payload would still pass. Also uncovered: non-default ports in the *generated* script (only the native side is tested), the `WKWebViewVC` call-site gate, and every scenario in P2-1 through P2-4.

### Nits

- `WKWebViewVC.m:832` and `:834` read `[WKApp shared].config.apiBaseUrl` twice; read once into a local and reuse.
- `WKWebViewVC.m:833` reads `NSUserDefaults` directly where `[[WKSpaceFilter shared] currentSpaceId]` is the codebase's own accessor for the identical key, with a type check included. Direct reads are common elsewhere in the repo, so this is consistency-only.
- `WKDocsViewerSpaceHandoff.m:16` recompiles the `NSRegularExpression` on every call; a `dispatch_once` static would be tidier. Negligible at one call per webview.

### Q1 — please confirm `/docs/:id/present`

`/d/:docId` and `/ppt/d/:docId` both correspond to routes I can see on the web side. `/docs/[^/]+/present` I could not find anywhere in the open repositories — the closest thing is `/docs` as an in-shell menu route path. If it is a real route owned by a closed-source module, no change is needed and this is just a note. If it is speculative, it is allowlist surface with no consumer and should come out.

## 3. Overall verdict

**APPROVE**

No P0/P1. The change is correctly scoped, fail-closed on both the native and the runtime side, verified against the actual storage contract it depends on rather than an assumed one, and it is a strict improvement over the current behaviour where the embedded web app has no idea which Space the user is in. Every finding above is a follow-up, not a merge blocker.

## 4. Suggested follow-ups

Rough priority order, all post-merge:

1. Close the redirect gap (P2-1) — pick one of the two approaches and write down why.
2. Tighten the path segment to the web contract's `[A-Za-z0-9_-]{1,128}` — this fixes P2-3 and half of P2-5 in one edit.
3. Add a shared-corpus test that asserts the ObjC regex and the emitted JS regex classify identical inputs identically (P2-5), plus at least one test that actually evaluates the generated script and inspects the resulting storage (P2-7).
4. Derive the deployment path prefix from `apiBaseUrl` the way `WKApp.serverOrigin` already does (P2-4).
5. Swap `isEqualToNumber:` for `isEqual:` (P2-6) and apply the three nits.

## 5. Additional observations

- **Not verifiable from this diff:** CI still owes the real evidence here. The author states that a full XCTest/build was not run locally because Xcode/CocoaPods was unavailable, so the platform test run and build are the gate. Also unverifiable from repository code: whether the deployed reverse proxy decodes `%2F` before routing (bears on P2-3), whether every deployment mounts Docs at the origin root (P2-4), and whether the server treats `currentSpaceId` as authorization rather than UI context. The last one is worth a moment's thought — this review assumes it is UI context, and the server-side `open-context` preflight resolving a document's Space from its id supports that reading, but if it ever became authorization-bearing the severity of P2-1 through P2-3 would rise.
- **No sibling-PR conflict.** The other open PRs touch disjoint files; nothing else edits this `project.pbxproj` or the webview sources.
- **Cross-platform parity** was not assessed — if the Android client has, or needs, an equivalent handoff, that is worth tracking separately so the two do not drift.
- **`check-sprint` is currently failing** on this PR. It is a repository process check unrelated to this diff, but it will need clearing before merge.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #76 (octo-ios)

## Summary

Adds a same-origin-only handoff that injects the native "current space" into the standalone Docs viewer when it opens in the app webview. A new helper (`WKDocsViewerSpaceHandoff`) validates that the URL matches the API server origin exactly (scheme + host + effective port) and one of the viewer routes (`/d/*`, `/ppt/d/*`, `/docs/*/present`), then `WKWebViewVC` installs a document-start `WKUserScript` that sets `localStorage.currentSpaceId` — or removes it when no space is active, so pooled web content cannot reuse a stale value. The injected JS re-checks origin and path at runtime, so the native gate and the script-side gate must both agree before storage is touched. Overall a careful, defensively scoped change; no P0/P1 issues found.

## Verification

Static analysis only at head `78c5d027e2752a9669c23e1259412fd5eaa170df`; build and tests not executed in this environment.

- ✅ **Data flow into the gate** — all 16 `WKWebViewVC` instantiation sites assign `vc.url` before push; the gate reads `self.url` at webView creation (first statement of `viewDidLoad`), so it sees the intended URL.
- ✅ **`currentSpaceId` has real producers** — written by login/register/space-gate flows and on Space switch (`Modules/WuKongLogin/WuKongLogin/Classes/Login/WKLoginModule.m:158`); same source `WKExternalViewerResolver.currentViewerSpaceId` already consumes.
- ✅ **Origin spoofing rejected** — exact lowercased-host string compare (no suffix matching), `user@evil` userinfo resolves to the real host, scheme downgrade and port mismatch fail; covered by the new tests and re-verified in code.
- ✅ **JS escaping** — values embedded via `NSJSONSerialization` plus explicit U+2028/U+2029 replacement; hostile spaceId (quotes, backslashes, newlines, `</script>`) has no script-breakout path. Input-shape edges (nil/empty URL, nil/malformed `apiBaseUrl`, nil/empty spaceId) are guarded in changed code; single production caller, no bypasses.
- ✅ **No cross-instance leakage** — `WKWebViewConfiguration`/`userContentController` are allocated per webview instance (only the process pool is shared); the script is added once per instance and released with the VC, no accumulation.
- ✅ **Build wiring** — podspec `source_files = 'WuKongBase/Classes/**/*'` picks up the new production files; pbxproj adds the test to PBXBuildFile, PBXFileReference, group, and Sources phase.
- ✅ **Subclasses** — `WKAuthWebViewVC` inherits the gate but OIDC URLs can never match the viewer path regex; the JS-disabled `isolatedRenderWebView` is untouched.

## Findings

No P0/P1 issues; one P2 below.

### P2 — Space snapshot frozen for the webview's lifetime (`Modules/WuKongBase/WuKongBase/Classes/Sections/Common/Webview/WKWebViewVC.m:833`)

`currentSpaceId` is captured once when the webView is created and baked into the script source — the comment documents this capture-now design as intentional. The script runs on every later document-start navigation of this webview and re-injects the captured value, so if the user switches Space natively while this `WKWebViewVC` instance is still alive, a subsequent full navigation to a viewer route bootstraps Docs with the stale space, which can mislabel external members. Reachability is low (the VC is normally popped before a Space switch can happen), so this is non-blocking; if you want to close it, refresh the value per navigation (e.g. replace the script in `decidePolicyForNavigationAction`) instead of capturing once.

## Human-verify

1. The standalone Docs web app reads `localStorage.currentSpaceId` at bootstrap — that consumer lives in the web repo and cannot be confirmed from this checkout. **Not a merge blocker for this PR.**
2. The gate requires the viewer page to be served from exactly the `apiBaseUrl` origin. If any deployment serves the viewer from a different host/CDN, the handoff silently no-ops there (fail-safe, but feature-off). Worth confirming against deployment topology. **Not a merge blocker.**
3. Handoff installs only when the webview's *initial* URL is a viewer URL; reaching the viewer by clicking through from another page in the same webview would not get it. Direct open from chat links appears to be the intended path — worth confirming no click-through flow exists. **Not a merge blocker.**

- setItem/removeItem parity: nil or empty spaceId emits `removeItem`, matching the header contract about pooled-storage cleanup.
- Native (`percentEncodedPath`) and JS (`location.pathname`) route regexes are character-for-character identical; query/fragment ignored on both sides.
- `forMainFrameOnly:YES` — no injection into iframes; the whole script is try/catch-wrapped, so a storage exception degrades to a no-op.
- The pre-existing language user script is unaffected (added after, independent concern).

## Verdict: COMMENT

Tight, well-scoped change with defense in depth (native gate + runtime JS gate) and solid test coverage of the trust boundary. The single P2 is an edge-case staleness question for the author to rule on; nothing here blocks merge.

PR #79, opened 2026-08-18
fix(tabbar): iPad 上 tab 图标不在选中胶囊正中

swr614:
## What

修复 iPad 上底部 tab 图标不在选中态高亮胶囊（pill）正中的问题。

## Why

iPad 上「消息」图标偏右、「我的」偏左，只有中间的「上下文」看着是正的。
根因是 pill 与 item 用了两套定位口径。

| | 公式 | 三个中心 |
|---|---|---|
| pill（自绘） | `islandW / count` 等分 | 138.67 / 384 / 629.33 |
| item（UIKit） | 固定 201pt + 整簇居中 | 183 / 384 / 585 |
| 偏差 | | **+44.3 / 0 / −44.3** |

`736 / 2 = 384` 恰是浮岛中点，中间那个 item 与中间那个 pill 必然重合 —— 这就是
「只有上下文是正的」的由来。`UITabBar.itemPositioning` 默认 `Automatic`，在横向
compact（iPhone）下等价 `Fill`，与自绘 pill 口径一致。
## How

- **`WKOctoTabBar`（`UITabBar` 子类）** —— 在 `layoutSubviews` 的 `super` 之后接管
  `button.frame = islandW/count` 槽位。`viewDidLayoutSubviews` 那个时机早于/交织于
  tabBar 自身的 `layoutSubviews`，写入会被系统重排覆盖，只有这里能保证是最后一个
  写入者。`button.frame` 即点击热区，所以点击位置随图标同步。
- **报 compact 让 UIKit 排 stacked 布局** —— 文字在图标正下方，而非 iPad 默认的
  inline（文字在图标右下方且纵向重叠 5.5pt）。iOS 17+ 走受支持的
  `traitOverrides.horizontalSizeClass`，17 以下用 `traitCollection` 覆写兜底。仅影响
  该 view，子 VC 的 trait 不变。
- **`_islandFrameInParent` 作为唯一几何真源** —— 只从 `self.view.bounds` +
  `safeAreaInsets` 算。原来 pill 与 item 各自读 `bar.frame`，而它在布局过程中会从
  768 变成 736，两边在不同时机读到不同值，表现为「三个图标全部偏左，点一下 tab 才
  对上」。
- **不碰 `icon` / `label` 的 frame** —— UIKit 会在事务提交后重排 button 内容，并沿用
  我们硬写给 label 的宽度：`30 + 6 + 217.5 = 253.5 > 245.33`，图标被算到 `x = −4.5`
  跑出 button 外。内部排版交回 UIKit。
- **附带**：`+tabConfigs` 收敛 tab 声明（原 `onLangChange` 另有一份硬编码标题数组）；
  未读角标索引改按 VC 类型查而非写死 `0`；移除排查期的调试日志。
回应 review 的修订：

- **P0 编译中断** —— `kWKItemTitleFontSize` 声明在使用点之后 75 行，文件作用域 C
  标识符按文本顺序查找，Clang 报 `use of undeclared identifier`。已前移。
- **pill 动画状态机** —— `isPillAnimating` 改代次令牌，只有最新一代的 completion 有
  资格清标志；守卫从调用点收进 `_layoutPillIndicatorAnimated:` 内部（原先两个
  `animated:NO` 调用点只守住一个，`_applyCapsuleStyleToTabBar` 那处会被旋转/键盘
  绕过）；守卫收严后在 completion 补一次收敛定位。
- **`_sortedItemButtons` 加 `UIControl` 过滤** —— `UITabBarButton` 是 `UIControl`，
  `_UIBarBackground` 是普通 `UIView`。原判据可能误收，导致数量闸门触发、整段修正
  静默失效；闸门另加 DEBUG 日志。
- **KVC 换 bar 的失败路径** —— 改指针判等，并在上屏后核验 `tabBar.superview`，覆盖
  「私有 setter 消失 → 静默退化成写 ivar → 浮岛/pill/角标全消失」这条路径。
- **角标在 `layoutSubviews` 回调路径上重定位**，与另一条路径顺序一致。
- **修正 6 处与代码矛盾的注释。**

一处未按 review 建议实现：坐标不变性改用 `convertRect:fromView:` 换算，而非加尺寸
闸门 —— 闸门会把「前提不成立」变成又一个静默 no-op，换算后则不依赖调用顺序。
## Checklist

## Testing

**已验证** —— 编译：

```
xcodebuild -workspace OctoiOS.xcworkspace -scheme OctoiOS -configuration Debug \
  -sdk iphonesimulator -destination 'generic/platform=iOS Simulator' build
→ ** BUILD SUCCEEDED **
```

Xcode 26.2 (17C52)，`WKMainTabController.m` 零 warning。与 CI 同 scheme、同
destination。

## Breaking Changes

N/A

### DIFF
--- a/Octo/WKMainTabController.m
+++ b/Octo/WKMainTabController.m
@@ -37,33 +37,152 @@
 @property(nonatomic,weak,nullable) UIViewController *lastSelectedViewController;
 @property(nonatomic,assign) NSTimeInterval lastMessageTabReselectAt; // CACurrentMediaTime 单位秒
 
+// _layoutTabItemsIntoSlots 的重入保护, 见该方法注释。
+@property(nonatomic,assign) BOOL isLayingOutTabItems;
+// 命中重入时置位, 由外层调用返回前重跑一次 —— 不能直接丢弃, 见 _layoutTabItemsIntoSlots 注释。
+@property(nonatomic,assign) BOOL needsTabItemRelayout;
+// 切 tab 的 pill spring 动画进行中。此时 layoutSubviews 回调里不能用 animated:NO 覆盖 pill,
+// 否则滑动被打断变成硬跳。
+@property(nonatomic,assign) BOOL isPillAnimating;
+
 @end
 
 // 双击判定窗口。系统双击间隔约 0.25s，留一点余量；再长会把「点一下看一眼、
 // 过一会儿再点一下」误判成双击。
 static NSTimeInterval const kWKMessageTabDoubleTapInterval = 0.35;
 
+// +tabConfigs 里每条配置的 key
+static NSString *const kWKTabCfgClass    = @"cls";    // UIViewController 子类
+static NSString *const kWKTabCfgTitleKey = @"titleKey"; // 传给 LLang 的中文 key
+static NSString *const kWKTabCfgImage    = @"image";  // 普通态图标名; 选中态 = 该名 + "1"
+
+#pragma mark - WKOctoTabBar
+
+// 职责一: 拿到「系统排完 item 之后」这个唯一可靠的时机。
+//
+// 为什么必须要它: UITabBar 在 iPad 全屏 (横向 regular) 下按「item 宽 201pt + 整体居中
+// 成一簇」排布, 而 _pillFrameForIndex: 按 barW/count 三等分算 pill。实测 (iPad Air 2 /
+// iPadOS 15.8, 竖屏 736pt 宽) 两套口径差 44.3pt:
+//     消息   item 中心 183   vs pill 中心 138.7  → +44.3 (偏右)
+//     上下文 item 中心 384   vs pill 中心 384    →   0   (恰好重合, 所以只有它看着是正的)
+//     我的   item 中心 585   vs pill 中心 629.3  → −44.3 (偏左)
+// 在 viewDidLayoutSubviews 里改 item frame 无效 —— 那个时机早于 / 交织于 tabBar 自己的
+// layoutSubviews, 写进去的值会被系统重排覆盖 (曾被诊断日志误导: 日志打在写入那一刻,
+// 拍不到之后的回滚)。只有 layoutSubviews 内、super 之后能保证我们是最后一个写入者。
+//
+// 职责二: 把 horizontalSizeClass 报成 compact, 让 UIKit 自己选 stacked 布局。
+// 见 -traitCollection 注释。
+//
+// 回退到原实现的路径: 删掉 viewDidLoad 里的 setValue:forKey:@"tabBar" 那一行即可 ——
+// 系统换回原生 UITabBar, 行为退回「图标不居中 + 文字在图标右下」的原状, 其余功能
+// (pill / 浮岛 / 角标 / 双击) 都不依赖本类。
+@interface WKOctoTabBar : UITabBar
+/// 每次 super layoutSubviews 之后回调, 供宿主接管 item 的水平/垂直布局。
+@property(nonatomic,copy,nullable) void (^onDidLayoutSubviews)(UITabBar *bar);
+@end
+
+@implementation WKOctoTabBar {
+    // traitCollection 的缓存。UIKit 在布局期间会极高频调用 traitCollection,
+    // 每次返回**新对象**会让某些内部路径按 isEqual: 判成「trait 变了」, 触发
+    // traitCollectionDidChange: → updateTabBarAppearance → 重新布局 → 又读 trait,
+    // 而本项目 WKBaseVC.traitCollectionDidChange 还会发 WK_NOTIFY_STYLE_CHANGE
+    // 全局通知, 代价被放大。缓存后同一个 base 始终返回同一个派生对象。
+    UITraitCollection *_cachedBaseTraits;
+    UITraitCollection *_cachedCompactTraits;
+}
+
+- (void)layoutSubviews {
+    [super layoutSubviews];
+    if (self.onDidLayoutSubviews) {
+        self.onDidLayoutSubviews(self);
+    }
+}
+
+/// 只把 horizontalSizeClass 改成 compact, 其余 trait (userInterfaceStyle / 缩放等)
+/// 原样从 super 带过来。
+///
+/// 为什么: UITabBar 的内容布局按 horizontalSizeClass 分支 ——
+///     compact (iPhone 竖屏) → stacked  : 图标在上、文字在下, 都居中  ← 设计要的
+///     regular (iPad 全屏)   → inline   : 图标和文字横向并排, 整对居中
+/// 实测 (iPad Air 2 / iPadOS 15.8) 走的是 inline, 所以文字落在图标右下方且与图标
+/// 纵向重叠 5.5pt:
+///     icon  x 94..124   y 23..53
+///     label x 130..150.5 y 47.5..59.5
+/// 这不是 bug 而是 iPad 原生 tab bar 的传统样式, 但与本项目设计 (kWKCapsuleHeight
+/// 注释里的「icon + 间距 4 + 标题」竖排) 不符。
+///
+/// 走这条路而不是用 imageInsets / titlePositionAdjustment 去补偏移: 改 UIKit 的**输入**
+/// 让它自己排出我们要的布局, 而不是补它的**输出** —— 后者已多次被证明会在事务提交后
+/// 被系统改回去, 且偏移量随标题文字宽度变化 (中文 −13.67 / −18.67, 换语言又是另一组)。
+///
+/// 影响范围仅限本 view: 子 VC 的 trait 不受影响, iPad 上会话列表等仍按 regular 布局。
+/// stackedLayoutAppearance 的 titlePositionAdjustment (0,-12) 会因此真正生效 ——
+/// 它本来就是为 stacked 布局写的, 走 inline 时一直在空转。
+- (UITraitCollection *)traitCollection {
+    UITraitCollection *base = [super traitCollection];
+    if (base.horizontalSizeClass == UIUserInterfaceSizeClassCompact) {
+        return base;   // 本来就是 compact (iPhone / iPad 分屏窄态) —— 原样返回, 不做包装
+    }
+    // 同一个 base 复用同一个派生对象, 避免每次调用都造新对象 (见 ivar 处注释)
+    if (_cachedCompactTraits && _cachedBaseTraits && [_cachedBaseTraits isEqual:base]) {
+        return _cachedCompactTraits;
+    }
+    UITraitCollection *derived = [UITraitCollection traitCollectionWithTraitsFromCollections:@[
+        base,
+        [UITraitCollection traitCollectionWithHorizontalSizeClass:UIUserInterfaceSizeClassCompact]
+    ]];
+    _cachedBaseTraits = base;
+    _cachedCompactTraits = derived;
+    return derived;
+}
+
+@end
+
 @implementation WKMainTabController
 
 - (void)viewDidLoad {
     [super viewDidLoad];
     self.delegate = self;
+
+    // 换成 WKOctoTabBar, 拿到「系统排完 item 之后」的回调 (见该类注释)。
+    // tabBar 是只读属性, 只能走 KVC 私有 setter —— 必须在 addChildViewController 之前,
+    // 否则 item 会先挂到原生 bar 上。setValue 失败时 self.tabBar 仍是系统实例,
+    // 下面的类型判断会跳过 hook, 功能降级为「图标不居中」, 不会 crash。
+    WKOctoTabBar *octoBar = [[WKOctoTabBar alloc] init];
+    @try {
+        [self setValue:octoBar forKey:@"tabBar"];
+    } @catch (NSException *e) {
+        NSLog(@"[WKMainTabController] 安装 WKOctoTabBar 失败, 退回系统 UITabBar: %@", e.reason);
+    }
+    if ([self.tabBar isKindOfClass:[WKOctoTabBar class]]) {
+        __weak typeof(self) weakSelf = self;
+        [(WKOctoTabBar *)self.tabBar setOnDidLayoutSubviews:^(UITabBar *bar) {
+            __strong typeof(weakSelf) self_ = weakSelf;
+            if (!self_) return;
+            // item 槽位与 pill 都从 _islandFrameInParent 取几何 (唯一真源), 且在同一个
+            // 回调里紧邻执行 —— 两者不可能再读到不同的宽度。
+            [self_ _layoutTabItemsIntoSlots];
+            // 切 tab 的 spring 滑动进行中就别覆盖, 否则动画被打断变硬跳。
+            if (!self_.isPillAnimating) {
+                [self_ _layoutPillIndicatorAnimated:NO];
+            }
+        }];
+    }
+
     // 监听 viewConfigChange 通知（WKBaseVC 的 traitCollectionDidChange 会发这个）
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(onStyleChange) name:@"WK_NOTIFY_STYLE_CHANGE" object:nil];
     // 切语言时 tabbar item title 不会自动刷新, 必须监听 WKNOTIFY_LANG_CHANGE 重新走 LLang
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(onLangChange) name:WKNOTIFY_LANG_CHANGE object:nil];
 
-    [self setupChildVC:WKConversationListVC.class title:LLang(@"消息")
-                 image:[UIImage imageNamed:@"消息"]
-         selectedImage:[UIImage imageNamed:@"消息1"]];
-
-    [self setupChildVC:OctoContextEntryVC.class title:LLang(@"上下文")
-                 image:[UIImage imageNamed:@"上下文"]
-         selectedImage:[UIImage imageNamed:@"上下文1"]];
-
-    [self setupChildVC:WKMeVC.class title:LLang(@"我的")
-                 image:[UIImage imageNamed:@"我的"]
-         selectedImage:[UIImage imageNamed:@"我的1"]];
+    // tab 列表的唯一声明处 —— 见 +tabConfigs。
+    // 【以后加 tab 只改那个数组，本方法不用动】
+    for (NSDictionary *cfg in [self.class tabConfigs]) {
+        NSString *imageName = cfg[kWKTabCfgImage];
+        [self setupChildVC:cfg[kWKTabCfgClass]
+                     title:LLang(cfg[kWKTabCfgTitleKey])
+                     image:[UIImage imageNamed:imageName]
+             selectedImage:[UIImage imageNamed:[imageName stringByAppendingString:@"1"]]];
+    }
 
     // 必须在 setupChildVC 之后，applySelectedTitleColor 要遍历 self.viewControllers
     [self updateTabBarAppearance];
@@ -73,12 +192,39 @@ static NSTimeInterval const kWKMessageTabDoubleTapInterval = 0.35;
     self.lastSelectedViewController = self.viewControllers.firstObject;
 }
 
+/// tab 配置：**本类里唯一声明 tab 列表的地方**。
+///
+/// 【以后新增底部 tab 只需在这个数组里加一条】—— viewDidLoad 建 childVC、onLangChange
+/// 刷标题、pill / item 槽位宽度、未读角标定位全部自动跟上, 不需要再改任何按数量写死的
+/// 常量或索引。
+///
+/// 位置自适应的原理: pill 和 item 槽位都按 islandWidth / viewControllers.count 等分
+/// (见 _islandFrameInParent / _pillFrameForIndex: / _applyTabItemSlotsOnce), count 变化
+/// 时两边同步变, 天然对齐。
+///
+/// 选中态图标按现有命名约定 = image + "1" (消息 / 消息1)。新图标请沿用:
+///   - 60×60 @2x + 90×90 @3x (现有 imageset 只填了 2x, @3x 设备上会放大发糊)
+///   - 图形在画布内**居中** (现有 消息1 偏右 0.75pt、我的1 偏左 0.5pt, 是资源本身不对称)
+///   - 单色剪影 + alpha (代码用 AlwaysTemplate 重染, 颜色由 appearance 控制)
+///
+/// 数量上限提醒: pill 宽 = islandW / count − 28。「上下文」三字 @10pt 约 31pt 宽,
+/// 所以需要 islandW ≥ 59 × count。iPhone SE 一代 (islandW 288) 到 4 个仍有余量,
+/// 5 个开始文字会顶出 pill —— 那时才需要把 kWKPillHorizontalInset 改成自适应。
++ (NSArray<NSDictionary *> *)tabConfigs {
+    return @[
+        @{ kWKTabCfgClass: WKConversationListVC.class, kWKTabCfgTitleKey: @"消息",   kWKTabCfgImage: @"消息"   },
+        @{ kWKTabCfgClass: OctoContextEntryVC.class,   kWKTabCfgTitleKey: @"上下文", kWKTabCfgImage: @"上下文" },
+        @{ kWKTabCfgClass: WKMeVC.class,               kWKTabCfgTitleKey: @"我的",   kWKTabCfgImage: @"我的"   },
+    ];
+}
+
 - (void)onLangChange {
-    // tab item 顺序 = setupChildVC 顺序: 消息 / 上下文 / 我的
-    NSArray<NSString *> *titleKeys = @[@"消息", @"上下文", @"我的"];
+    // 标题 key 与顺序都来自 +tabConfigs, 不再单独维护一份数组 —— 原实现在这里硬编码了
+    // @[@"消息", @"上下文", @"我的"], 加第 4 个 tab 时极易漏改, 后果是新 tab 切语言不刷新。
+    NSArray<NSDictionary *> *cfgs = [self.class tabConfigs];
     [self.viewControllers enumerateObjectsUsingBlock:^(UIViewController *vc, NSUInteger idx, BOOL *stop) {
-        if (idx < titleKeys.count) {
-            vc.tabBarItem.title = LLang(titleKeys[idx]);
+        if (idx < cfgs.count) {
+            vc.tabBarItem.title = LLang(cfgs[idx][kWKTabCfgTitleKey]);
         }
     }];
 }
@@ -147,9 +293,9 @@ static NSTimeInterval const kWKMessageTabDoubleTapInterval = 0.35;
         normalTextColor   = [UIColor colorWithRed:140.0/255.0 green:140.0/255.0 blue:145.0/255.0 alpha:1.0];
     }
     NSDictionary *normalAttrs   = @{ NSForegroundColorAttributeName: normalTextColor,
-                                     NSFontAttributeName: [UIFont systemFontOfSize:10] };
+                                     NSFontAttributeName: [UIFont systemFontOfSize:kWKItemTitleFontSize] };
     NSDictionary *selectedAttrs = @{ NSForegroundColorAttributeName: selectedTextColor,
-                                     NSFontAttributeName: [UIFont systemFontOfSize:10] };
+                                     NSFontAttributeName: [UIFont systemFontOfSize:kWKItemTitleFontSize] };
 
     if (@available(iOS 13.0, *)) {
         UITabBarAppearance *appearance = [[UITabBarAppearance alloc] init];
@@ -164,6 +310,10 @@ static NSTimeInterval const kWKMessageTabDoubleTapInterval = 0.35;
         // 底缘也只有 ~3pt, 视觉上文字"压在"胶囊边框。-12pt 让 title 上抬 12pt, 在 pill 底
         // 留出 ~7pt 喘息。注意:offset 只影响 title, 不影响 icon, 所以把 bar 高度 (76) 保留
         // 给 icon 上方多出来的空白。
+        //
+        // (曾一度删掉这三行, 改成在 _applyTabItemSlotsOnce 里直接写 icon/label 的 frame 来
+        //  做垂直居中 —— 那条路已证明是错的: UIKit 会在事务提交后重排 button 内容并沿用
+        //  我们硬写的 label 宽度, 把图标挤到负坐标。纵向微调必须走这个受支持的 API。)
         UIOffset titleOffset = UIOffsetMake(0, -12);
         appearance.stackedLayoutAppearance.normal.titleTextAttributes = normalAttrs;
         appearance.stackedLayoutAppearance.normal.iconColor = normalTextColor;
@@ -216,6 +366,9 @@ static const CGFloat kWKContentBottomPadding = 24;
 // pill 内缩 —— pill 比单个 item 区略小, 给 capsule 边缘留点呼吸
 static const CGFloat kWKPillVerticalInset    = 8;    // 上下各 8pt: 圆形端帽外露 ~6pt
 static const CGFloat kWKPillHorizontalInset  = 14;   // 左右各 14pt: 让 pill 视觉上"包住" icon+标题, 不顶到胶囊端帽
+// tab 标题字号。updateTabBarAppearance 的 titleTextAttributes 与 _applyTabItemSlotsOnce
+// 算 label 高度都用它, 保持单一来源。
+static const CGFloat kWKItemTitleFontSize    = 10;
 
 // Liquid Glass 关闭后,标准 UITabBar 是横贯全宽的扁平条。这里手工把它做成"浮岛胶囊":
 // 白底 capsuleBackground + pill 选中胶囊 + tabBar 透明在最上层承接事件。
@@ -270,6 +423,34 @@ static const CGFloat kWKPillHorizontalInset  = 14;   // 左右各 14pt: 让 pill
     self.pillIndicator = pill;
 }
 
+/// 浮岛（= tabBar / capsuleBackground 的目标 frame）在 tabBar.superview 坐标系下的位置。
+///
+/// 【本方法是整个 tabbar 布局的唯一几何真源，务必保持所有消费方都走它】
+///
+/// 为什么必须集中到一处: 原实现里 _pillFrameForIndex: 读 bar.frame / bar.bounds,
+/// _applyTabItemSlotsOnce 读 bar.bounds —— 都是「UIKit 当前摆到哪儿」的可变中间状态。
+/// 这两段在**不同时机**跑 (item 在 WKOctoTabBar.layoutSubviews 里, pill 曾在
+/// viewDidLayoutSubviews 里), 而 UITabBarController 会先给 tabBar 一个它自己的 frame
+/// (全宽 768 / 1024), 之后才被 _applyCapsuleStyleToTabBar 改成浮岛宽度 (736 / 992)。
+/// 于是两边可能读到不同的宽度, 算出不同的 itemW:
+///     竖屏 pill 按 768/3=256 算、图标按 736/3=245.33 算
+///     → pill 中心 144 / 400 / 656, 图标中心 138.67 / 384 / 629.33
+///     → 偏差 −5.33 / −16 / −26.67, **三个图标全部偏左**, 且越往右偏得越多
+/// 这正是「刚进入时图标都偏左, 点一下才正常」的成因 (点击触发的新 layout 里两边都已
+/// 收敛到浮岛宽度)。
+///
+/// 改成从 self.view.bounds + safeAreaInsets 直接算, 不读任何 bar 的实时 frame ——
+/// 布局结果就与「什么时候跑」彻底无关, 这类时序 bug 从根上消掉。
+- (CGRect)_islandFrameInParent {
+    CGFloat bottomSafe = self.view.safeAreaInsets.bottom;
+    CGFloat parentW    = self.view.bounds.size.width;
+    CGFloat parentH    = self.view.bounds.size.height;
+    return CGRectMake(kWKCapsuleInsetX,
+                      parentH - kWKCapsuleHeight - bottomSafe - kWKCapsuleBottomGap,
+                      parentW - kWKCapsuleInsetX * 2,
+                      kWKCapsuleHeight);
+}
+
 - (void)_applyCapsuleStyleToTabBar {
     UITabBar *bar = self.tabBar;
     if (!bar.superview || CGRectIsEmpty(self.view.bounds)) {
@@ -277,14 +458,7 @@ static const CGFloat kWKPillHorizontalInset  = 14;   // 左右各 14pt: 让 pill
     }
     [self _setupCapsuleViewsIfNeeded];
 
-    CGFloat bottomSafe = self.view.safeAreaInsets.bottom;
-    CGFloat parentW    = self.view.bounds.size.width;
-    CGFloat parentH    = self.view.bounds.size.height;
-
-    CGRect target = CGRectMake(kWKCapsuleInsetX,
-                               parentH - kWKCapsuleHeight - bottomSafe - kWKCapsuleBottomGap,
-                               parentW - kWKCapsuleInsetX * 2,
-                               kWKCapsuleHeight);
+    CGRect target = [self _islandFrameInParent];
     if (!CGRectEqualToRect(bar.frame, target)) {
         bar.frame = target;
     }
@@ -304,6 +478,10 @@ static const CGFloat kWKPillHorizontalInset  = 14;   // 左右各 14pt: 让 pill
         [UIBezierPath bezierPathWithRoundedRect:CGRectMake(0, 0, target.size.width, target.size.height)
                                    cornerRadius:kWKCapsuleHeight / 2.0].CGPath;
 
+    // item button / icon / label 的水平布局由我们接管, 必须在 pill 和 badge 之前 ——
+    // badge 要读修正后的 icon frame。
+    [self _layoutTabItemsIntoSlots];
+
     // pill 跟随当前 selectedIndex 定位 (无动画, layout 阶段)
     [self _layoutPillIndicatorAnimated:NO];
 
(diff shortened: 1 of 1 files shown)


Jerry-Xin (dismissed):
**APPROVE** — verified at head d6e3b84e2230. The fix is sound: pill and item slots now derive from a single geometry source (`_islandFrameInParent`, computed from `self.view.bounds` + `safeAreaInsets`), which eliminates the timing skew where pill and items could read different bar widths mid-layout (the "all icons shifted left until you tap" symptom), and taking over button frames inside `WKOctoTabBar.layoutSubviews` after `super` guarantees last-writer-wins against UIKit's iPad regular-width "201pt centered cluster" arrangement. iPhone/compact behavior is unchanged (the trait override returns the base collection untouched when already compact, and slot math matches the previous Fill layout once the bar converges to island width).

Review-process note: an earlier automated pass flagged `kWKItemTitleFontSize` used at `Octo/WKMainTabController.m:296` (declared at line 371) as a use-before-declaration compile error. That does not apply here: both the use and the declaration sit inside the same `@implementation WKMainTabController` block, and Clang accepts this ordering — verified with `clang -fsyntax-only` against the iOS SDK on a structurally identical repro. The file's existing capsule constants are already declared mid-`@implementation` in the same style.

### 🔴 Blocking items

None.

### 💬 Non-blocking notes

1. **KVC private setter** — `setValue:forKey:@"tabBar"` in `viewDidLoad` carries App Store / future-OS risk. It's mitigated by `@try/@catch` with graceful degradation to the system bar and a documented rollback path; still worth tracking as a fragility point if the internal ivar layout changes.
2. **`isPillAnimating` race on rapid tab switches** — if tab B is selected while tab A's spring is in flight, A's completion (`finished=NO`) resets the flag while B's animation still runs, so a layout callback in the remaining ~0.3s can overwrite the pill with `animated:NO` and collapse the second slide into a hard jump. A generation token would close this; cosmetic impact only.
3. **Stale doc comment on `_applyTabItemSlotsOnce`** — the header still says it centers icon/label and widens the label to slot width, but the final implementation intentionally leaves button contents to UIKit. Align the first line with the (excellent) in-loop comment.
4. **RTL** — manual frame math is LTR-only; slots and pill share the same origin so they stay mutually aligned under an RTL locale, but item order won't mirror. Fine for the current locale set; keep in mind if RTL is ever added.
5. **No tests** — the repo has no unit/snapshot infrastructure for UIKit layout, so manual verification on iPad (regular width, rotation, Split View) and iPhone (compact) before release is recommended; the measured before/after geometry in the PR description is a good start.
6. **Process** — the `check-sprint` CI fails: the PR description needs a `Closes #<issue>` reference to pass the sprint gate.

### ✅ Highlights

- `_islandFrameInParent` as the single geometry source of truth removes a whole class of timing bugs by construction.
- The `traitCollection` override fixes UIKit's *input* (compact → stacked layout) instead of patching its output; it is correctly cached to avoid trait-change churn and scoped to the bar only, leaving child VCs' traits untouched.
- Re-entrancy guard with a bounded (≤3) re-layout loop prevents a main-thread hang while still re-applying dropped corrections.
- `+tabConfigs` consolidates tab declaration in one place, and `_indexOfMessageTab` removes the hard-coded "Messages is index 0" assumption for the unread badge.

yujiawei (changes requested):
# Code Review — PR #79 (octo-ios)

Reviewed at head `d6e3b84e22300737b912457958c66e024cffef6b` (base `5a36a32d`), single file: `Octo/WKMainTabController.m` (+374 / −48).

The diagnosis in the PR description is convincing and the write-up is unusually good — collapsing pill geometry and item geometry onto a single source (`_islandFrameInParent`), and taking the item frames inside `layoutSubviews` after `super`, is the right shape for this fix. Handing icon/label layout back to UIKit instead of hand-writing their frames is also the correct correction of the earlier approach.

However **the branch as pushed does not compile**, so this has to be Changes Requested. Details below.

---

## 1. Spec compliance

Checked the diff against the "How" section of the PR description.

| Claimed change | Status |
|---|---|
| `WKOctoTabBar` takes over `button.frame` in `layoutSubviews` after `super` | Implemented (`:94-99`, `:549-630`) |
| Override `traitCollection` to report compact so UIKit picks `stacked` | Implemented (`:121-137`) |
| `_islandFrameInParent` as the single geometry source | Implemented; all three consumers (`_applyCapsuleStyleToTabBar` `:461`, `_applyTabItemSlotsOnce` `:591`, `_pillFrameForIndex:` `:711`) go through it |
| Do not touch `icon` / `label` frames | Honoured — nothing in the diff writes to them |
| `+tabConfigs` as the single tab declaration | Implemented (`:213-219`), consumed by `viewDidLoad` `:179` and `onLangChange` `:224` |
| Badge index looked up by VC class instead of hardcoded `0` | Implemented (`:665-671`, `:675-677`) |
| Debug logging removed | Done. The one remaining `NSLog` (`:155`) is an error path, not a debug probe — fine |

Nothing was built that wasn't asked for. Two additions are not listed in the description but are direct consequences of the new callback and are justified: the reentrancy guard (`:563-584`) and the `isPillAnimating` flag (`:737-753`). Both are needed because the new `layoutSubviews` callback would otherwise stomp the spring animation and could nest. Worth adding a line about them to the description so the reviewer of the next iteration knows they are load-bearing.

**Spec verdict: ❌** — the checklist item "Code compiles with no new warnings" is unchecked, and it is unchecked accurately: see P0 below. The "Testing" section is also empty while the code comments cite specific device measurements, so the evidence that clearly exists isn't recorded anywhere in the PR.

---

## 2. Code quality

### P0 — Build break: `kWKItemTitleFontSize` is used ~75 lines before it is declared

`Octo/WKMainTabController.m:296` and `:298`

```objc
NSDictionary *normalAttrs   = @{ NSForegroundColorAttributeName: normalTextColor,
                                 NSFontAttributeName: [UIFont systemFontOfSize:kWKItemTitleFontSize] };
NSDictionary *selectedAttrs = @{ NSForegroundColorAttributeName: selectedTextColor,
                                 NSFontAttributeName: [UIFont systemFontOfSize:kWKItemTitleFontSize] };
```

The declaration is at `Octo/WKMainTabController.m:371`:

```objc
static const CGFloat kWKItemTitleFontSize    = 10;
```

`kWKItemTitleFontSize` is a file-scope C identifier. Being inside an `@implementation` block does not defer name lookup — C name lookup is strictly textual, and a declaration later in the translation unit does not extend backwards. Clang emits `use of undeclared identifier 'kWKItemTitleFontSize'` at both use sites and the target fails to build. (Implicit-int was removed in C99, so this is a hard error, not a warning.)

I checked that nothing rescues it: the symbol appears exactly three times in the whole repository — the two uses and the one late declaration — and the app target has no prefix header.

```
Octo/WKMainTabController.m:296  ... systemFontOfSize:kWKItemTitleFontSize
Octo/WKMainTabController.m:298  ... systemFontOfSize:kWKItemTitleFontSize
Octo/WKMainTabController.m:371  static const CGFloat kWKItemTitleFontSize = 10;
```

**Fix:** move the declaration above `-updateTabBarAppearance` (next to `kWKMessageTabDoubleTapInterval` at `:52` is the natural home), or forward-declare it. It is worth understanding how this reached the PR — a local build would have caught it in one second, so the loop that produced this change apparently isn't compiling. That's more concerning than the one-line fix.

### P2 — `isPillAnimating` can be cleared by a stale completion, re-opening the exact glitch it was added to prevent

`Octo/WKMainTabController.m:737-753`

```objc
self.isPillAnimating = YES;
__weak typeof(self) weakSelf = self;
[UIView animateWithDuration:0.34 ... animations:^{
    self.pillIndicator.frame = target;
    ...
} completion:^(BOOL finished) {
    weakSelf.isPillAnimating = NO;
}];
```

Switch tabs twice inside the 0.34 s spring. The second call sets `isPillAnimating = YES` and starts animation B; starting B interrupts A, so A's completion fires with `finished == NO` and sets the flag back to `NO` while B is still running. From that point the `layoutSubviews` callback (`:166-168`) no longer skips, calls `_layoutPillIndicatorAnimated:NO`, and hard-snaps the pill mid-flight — the precise behaviour the flag exists to prevent.

Impact is cosmetic and self-healing (the next selection restores a correct spring), so it isn't a merge blocker on its own, but the fix is cheap: replace the boolean with a monotonically increasing generation token and only clear when the completion belongs to the current generation.

```objc
// ivar: NSUInteger _pillAnimationGeneration;
NSUInteger generation = ++_pillAnimationGeneration;
...
} completion:^(BOOL finished) {
    __strong typeof(weakSelf) strongSelf = weakSelf;
    if (strongSelf && strongSelf->_pillAnimationGeneration == generation) {
        strongSelf.isPillAnimating = NO;
    }
}];
```

Minor consistency note in the same block: the `animations:` block captures `self` strongly while `completion:` uses `weakSelf`. Not a leak (UIView doesn't retain these beyond the animation), but pick one.

### P2 — `_sortedItemButtons` heuristic is load-bearing now, and its failure mode is "the fix silently does nothing"

`Octo/WKMainTabController.m:508-526`

The filter is "has non-zero bounds AND contains a `UIImageView` subview". That was fine when the only consumer was badge placement (a false positive just misplaced a badge). Now the same list drives `button.frame` writes and is gated by an exact count match at `:592`:

```objc
if (CGRectIsEmpty(island) || count <= 0 || (NSInteger)buttons.count != count) {
    return;
}
```

`_UIBarBackground` is a sibling subview of the bar with non-zero bounds, and it holds `UIImageView` children whenever the bar has a background/shadow image — and `_applyCapsuleStyleToTabBar:471-472` explicitly sets both:

```objc
bar.backgroundImage     = [UIImage new];
bar.shadowImage         = [UIImage new];
```

On the tested OS this evidently doesn't match (the fix works), but whether `_UIBarBackground` materialises those image views is a UIKit implementation detail that has moved across releases — and it is being mixed here with the modern `UITabBarAppearance` path, which is exactly the kind of combination whose internals change. If it ever matches, `buttons.count` becomes 4, the count guard trips, and the whole item-slot correction silently no-ops on every layout pass with no log and no visible cause.

Cheap hardening that keeps the "don't hardcode the private class name" property: `UITabBarButton` is a `UIControl`; `_UIBarBackground` is a plain `UIView`.

```objc
if (![v isKindOfClass:[UIControl class]]) continue;
```

Consider also logging once (debug-only) when the count guard trips, so a future regression is diagnosable instead of invisible.

### P2 — `-traitCollection` override is the least future-proof part of the change

`Octo/WKMainTabController.m:121-137`

Overriding the `traitCollection` getter is not a supported extension point — UIKit keeps its own trait state, and whether an internal layout path consults `self.traitCollection` (your override) or the propagated environment (unaffected by it) is undocumented and has been reworked, notably by the iOS 17 trait system. The measurements quoted throughout the file are all from a single configuration (iPad Air 2 / iPadOS 15.8), and this file elsewhere explicitly deals with iOS 26 behaviour (`:260-262`), so the deployed range is much wider than the tested one. If the override stops being consulted, the bar reverts to `inline` and the layout regresses — silently, and only on iPad.

Since iOS 17 there is a supported API that does exactly this and propagates properly to descendants. Deployment target is iOS 14, so gate it and keep the getter override as the pre-17 fallback:

```objc
if (@available(iOS 17.0, *)) {
    self.traitOverrides.horizontalSizeClass = UIUserInterfaceSizeClassCompact;
}
```

Please also confirm this was exercised on a current iPadOS before merge, not only 15.8.

### P2 — KVC tab bar swap: the failure path is worse than the comment claims

`Octo/WKMainTabController.m:151-156`

```objc
WKOctoTabBar *octoBar = [[WKOctoTabBar alloc] init];
@try {
    [self setValue:octoBar forKey:@"tabBar"];
} @catch (NSException *e) { ... }
```

The comment says a failure means "`self.tabBar` is still the system instance, the hook is skipped, degrades to un-centred icons, no crash". That's true for the *exception* path only. KVC has a second path: if the private `setTabBar:` ever goes away, `setValue:forKey:` falls back to setting the backing ivar directly. No exception is raised, `self.tabBar` returns the new bar — but it was never wired into the view hierarchy, so `bar.superview` is `nil`, `_applyCapsuleStyleToTabBar:456` early-returns, and the floating island, the pill and the badge all disappear. That is a considerably louder failure than "icons not centred".

Make the degradation explicit and detectable:

```objc
if (self.tabBar != octoBar || self.tabBar.superview == nil) {
    // swap did not take effect as expected — report loudly
}
```

Separately, worth stating on the record: `setTabBar:` is not public API. This pattern is widespread and has historically shipped, but it is a review-surface risk that should be a conscious, documented decision rather than an incidental one. The rollback note at `:76-78` is good and should stay.

### P2 — Comments that contradict the code they document

This file's comments are its main asset, which makes drift expensive — a future maintainer will trust them over the code. Four of them are now wrong:

1. `:528-529` — "把每个 item button 摆到与 `_pillFrameForIndex:` 同源的槽位 (barW/count), **并把它内部的 icon / label 在槽位内水平居中、label 宽度撑到槽位全宽**". The body does the opposite and spends 20 lines (`:607-628`) explaining why touching icon/label caused the negative-coordinate bug. The header comment is an instruction to reintroduce that bug.
2. `:547-548` — "槽位算法与 `_pillFrameForIndex:` 都用 **bar.bounds.size.width** / count". Both now use `_islandFrameInParent`; not reading `bar.bounds` is the whole point of the change.
3. `:481-482` — "item button / **icon / label** 的水平布局由我们接管". Only `button.frame` is taken over.
4. `:369-370` — "`updateTabBarAppearance` 的 `titleTextAttributes` 与 **`_applyTabItemSlotsOnce` 算 label 高度**都用它". `_applyTabItemSlotsOnce` no longer computes a label height; the constant has exactly one consumer, so the "single source of truth" rationale no longer applies.

### P2 — The island/bar coordinate invariant is load-bearing but unasserted

`Octo/WKMainTabController.m:598-605`

```objc
CGFloat itemW = island.size.width / (CGFloat)count;
CGFloat itemH = island.size.height;
...
CGRect slot = CGRectMake((CGFloat)i * itemW, 0, itemW, itemH);
button.frame = slot;
```

`button.frame` is expressed in `self.tabBar.bounds`, but the slot is derived from `_islandFrameInParent`, which is in `tabBar.superview` coordinates. That is only correct while `bar.frame.size == island.size`. The assignment enforcing it lives in `_applyCapsuleStyleToTabBar:461-464`, reached from `viewDidLayoutSubviews` — a different entry point from the `WKOctoTabBar.layoutSubviews` callback that also calls this method.

In the current ordering the invariant does hold (`viewDidLayoutSubviews` runs before descendants are laid out, so the bar is already at island size by the time its `layoutSubviews` fires). I am not claiming a live failure. But the correctness of the slot math depends on an ordering guarantee established in another method via another entry point, with nothing at the use site saying so — and single-frame misplacement is precisely the bug class this PR exists to kill. A guard alongside the existing one at `:592` makes it self-evident and fail-safe, consistent with the file's own "宁可维持系统原样也不要摆出错位的半成品" policy:

```objc
if (!CGSizeEqualToSize(self.tabBar.bounds.size, island.size)) {
    return; // bar not yet resized to the island — retry next pass
}
```

### P3 (nit) — badge is not repositioned on the `layoutSubviews` callback path

`Octo/WKMainTabController.m:162-169`

The callback runs `_layoutTabItemsIntoSlots` and the pill, but not `_layoutMessageTabBadge`, even though it moves the buttons the badge is anchored to. In practice the badge converges because both paths compute the same slots and `_applyCapsuleStyleToTabBar` re-runs, but the comment at `:481-482` states the ordering constraint "items must run before the badge, the badge reads the corrected icon frame" — and this path breaks that ordering. Adding the badge call to the callback would make the two paths consistent.

**Quality verdict: Changes-Requested** (driven by the P0; the P2s are worth addressing but would not block on their own).

---

## 3. Verdict

**CHANGES_REQUESTED.** One blocking item — the branch does not compile (`kWKItemTitleFontSize` used at `:296`/`:298`, declared at `:371`). Everything else is a suggestion. Once the build is fixed I'd be happy to re-approve quickly; the architecture of the fix is sound and the reasoning behind it is well documented.

## 4. Coverage — what I did not verify

Stating this explicitly so the gaps aren't mistaken for clearance:

- **Not built or run.** No Xcode/simulator available in this environment. The P0 is a language-level declaration-order fact verified by inspection and by a reduced C repro, not by building this target.
- **No runtime/visual verification** on any device or size class. Every geometry number in this PR (the 44.3 pt divergence, 736/768, the icon/label frames) is taken from the author's measurements; none was independently reproduced. In particular the claim that the `traitCollection` override actually flips UIKit to `stacked` is unverified on any OS version.
- **Single-configuration evidence.** All measurements cited in the diff are iPad Air 2 / iPadOS 15.8. Not covered: current iPadOS, iPhone (both orientations, incl. compact-width where the override returns early), iPad Split View / Slide Over / Stage Manager, Dynamic Type, RTL, and the language-switch path through `onLangChange`.
- **Interaction surfaces not exercised:** tap hit-testing against the new `button.frame` slots, the message-tab double-tap gesture (`_handlePossibleMessageTabDoubleTap:`), badge placement with a live unread count, and rotation while a pill spring is in flight (the P2 above).
- **Repo-wide impact:** only `WKMainTabController.m` changed and `WKOctoTabBar` is file-local, so cross-file blast radius is minimal. I did confirm `setupChildVC:` (`:233-244`) adds child VCs directly without wrapping them in a `UINavigationController`, so `_indexOfMessageTab`'s `isKindOfClass:[WKConversationListVC class]` check against `self.viewControllers` is correct as written.
- **Unrelated red check:** `check-sprint` is failing on "Validate Sprint field on Project Board". That's project-board metadata, not code — flagging it only so it isn't attributed to this review.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #79 (octo-ios)

## Summary

本 PR 修 iPad 上底部 tab 图标不在选中 pill 正中的问题：用 `WKOctoTabBar` 子类（KVC 换掉 `tabBar`）在系统排完 item 后接管水平槽位布局，覆写 `traitCollection` 报 compact 让 UIKit 走 stacked 排版，并把 pill/item 几何统一收敛到唯一真源 `_islandFrameInParent`；附带把 tab 声明收敛到 `+tabConfigs`、未读角标索引改按 VC 类型查、布局接管加重入保护与有界重跑。整体思路（改 UIKit 的输入而非补它的输出、单一几何真源、重入保护带上限）成立，注释质量高、失败退化路径有考虑；但本轮审查发现一个阻塞落地的 P1 编译错误。

## Verification

Static analysis only at head `d6e3b84e`; build and tests not executed in this environment（无 iOS/Xcode 工具链）。实际做过的核对：

- ✅ `pr-base...HEAD` 全量 diff 逐 hunk 审读，并与 base 版本关键路径（`_applyCapsuleStyleToTabBar` / `_pillFrameForIndex:` / `onLangChange` / `_iconViewForTabIndex:`）逐一对比行为。
- ✅ `kWKItemTitleFontSize` 声明/使用顺序 —— 全仓 grep（含所有文件类型与 *.h）确认：使用点 `Octo/WKMainTabController.m:296` / `:298`，`static const` 定义在 `:371`，仓内无其他声明。
- ✅ `+tabConfigs` 三条配置 cls/titleKey/image 键齐全；image 名 + "1" 与 base 的 selectedImage 命名一致；标题 key 与顺序和 base 完全一致。
- ✅ 角标数据流：`setMessageUnreadCount:` → `_layoutMessageTabBadge` → `_indexOfMessageTab`（NSNotFound 已守卫）→ `_iconViewForTabIndex:`（越界已守卫）。
- ✅ 布局输入的空/错形状全部安全早退：`CGRectIsEmpty(island)`、`buttons.count != count` 跳过等下一拍、pill index clamp 到 `[0, count-1]`。
- ✅ CI（`.github/workflows/ci.yml`，macos-latest + xcodebuild + iOS SDK ≥ 26 断言）对 PR 跑构建，下述 P1 预计 CI 也会红。

## Findings

一个 P1 阻塞 + 两个 P2。

### P1 — kWKItemTitleFontSize 前向引用，target 无法编译 (`Octo/WKMainTabController.m:296`)

`updateTabBarAppearance` 在 `:296` / `:298` 使用 `kWKItemTitleFontSize`，但该常量的 `static const` 定义在 `:371`（胶囊外观常量块）。C/ObjC 同一编译单元内标识符必须先声明后使用，clang 会报 `use of undeclared identifier`，整个 target 无法编译。base 此处是字面量 `10`，使用点与声明点都是本 PR 新引入（diff-scope: new），全仓 header 无其他声明。构建失败 = 一切路径不可用，按 R1 为阻塞级。

```objc
// 修复：把声明移到使用点之前（或整个常量块上移 / 文件顶部前置声明）
static const CGFloat kWKItemTitleFontSize = 10;   // ← 需在 updateTabBarAppearance 之前
```

### P2 — isPillAnimating 生命周期：守卫不对称 + 快速切 tab 竞态 (`Octo/WKMainTabController.m:752`)

1) 不变量「isPillAnimating 期间不得用 animated:NO 覆盖 pill」只在 `onDidLayoutSubviews` 回调（`:166`）执行；`_applyCapsuleStyleToTabBar` 在 `:486` 无条件调用 `_layoutPillIndicatorAnimated:NO` —— spring 的 0.34s 窗口内发生 controller 级 layout（旋转 / 分屏 resize / 键盘）会把动画打断成硬跳。base 同样无守卫（pre-existing），但本 PR 引入的 flag 让两条路径不对称。
2) 快速连切两个 tab（<0.34s）时，旧动画被 `BeginFromCurrentState` 打断后其 completion 以 `finished=NO` 回调并复位 `isPillAnimating=NO`（`:752`），而此时新动画仍在进行，后续回调失去保护（缺 generation token；diff-scope: new）。

修复方向：自增动画代号，completion 仅在代号仍是当前时复位；`:486` 同样走守卫。

### P2 — KVC 换 tabBar 只验类型、不验视图层级附着，失败退化可能比原 bug 更糟 (`Octo/WKMainTabController.m:153`)

`setValue:forKey:@"tabBar"` 依赖私有 setter 以及 UIKit 内部把新 bar 挂进视图层级的行为；`:157` 只校验 `isKindOfClass`，未校验 `octoBar.superview`。若某个受支持 iOS 版本接受 KVC 写入却不把新 bar 挂进层级，`_applyCapsuleStyleToTabBar` 的 `bar.superview` 早退会让 capsule/pill/角标/item 槽位全部静默失效，item button 若挂在不可见的 bar 上用户可能完全看不到 tab —— 比「图标不居中」的原 bug 更糟。私有 API 跨版本漂移（未来 iOS 内部结构变化、App Store 审核）也是长期风险。静态分析无法证明该失败在哪些已支持版本可达（作者已在 iPadOS 15.8 实测通过），故不定 P1，但验证缺口是具体的。

修复方向：首个 layout 后校验 `self.tabBar.superview != nil`（失败则记录日志 + 明确退化标记）；合入前在受支持的 iOS 矩阵（iOS 16/17/18/26 + iPad 分屏）真机/模拟器验证渲染。

- `+tabConfigs` 收敛：与 base 的类/标题 key/图标名/顺序完全一致；`onLangChange` 语义不变。
- `_islandFrameInParent` 公式与 base 内联 capsule 计算一致（16 inset / 76 高 / 8 gap），仅抽取无行为变化；iPhone（compact）槽位 = barW/count，与原 Fill 行为一致。
- 重入保护：`_layoutTabItemsIntoSlots` guard + 3 次有界重跑，消除了裸 do/while 可能的主线程死循环；最坏推迟到下一拍 layout。
- `_sortedItemButtons`：stable sort 防首帧 x 全 0 乱序；显式排除直挂在 tabBar 上的 `_messageTabBadge` 防误数。
- 角标索引按 VC 类型查 + `NSNotFound` 守卫，消除写死 0 的隐患。
- 本地化仅 zh-Hans/en（无 RTL locale），x 升序排序 == tab 顺序在所有发行语言下成立。
- `traitCollection` 覆写对同一 base 缓存同一派生对象，避免 trait 对象翻新引发多余 relayout/通知；子 VC 的 trait 不受影响（视图树兄弟节点）。
- `onDidLayoutSubviews` block 弱引用 controller，无 retain cycle；`isPillAnimating` 不会永久卡死（completion 无论 finished 与否都复位，早退路径都在启动动画之前）。

## Verdict: CHANGES_REQUESTED

P1 编译错误是唯一阻塞：把 `kWKItemTitleFontSize` 的声明移到使用点之前即可清除，CI 也会验证。两个 P2 不阻塞但建议一并处理（isPillAnimating 代号化 + 双路径守卫、KVC 安装后校验 superview）。方案方向本身是对的，不需要推翻重来。

---

## 附录：Octo 终审腿复核材料（内部，勿外发）

### diff-scope 三问 + 定级（R1–R4）
- **P1 编译错误**：new（使用点与声明点均为本 PR 引入，base 是字面量 10）。R1：构建失败 → 所有路径不可用 → P1 → CHANGES_REQUESTED。
- **P2 isPillAnimating**：子项 1 pre-existing（base 的 viewDidLayoutSubviews 路径同样无守卫，行为一致，仅被新 flag 放大为不对称）；子项 2 new。R1 检验：不使能工作的路径不可用、不产生错误数据，最差为动画硬跳 → P2。
- **P2 KVC 校验缺口**：new。R1 检验：无法从静态分析证明失败态在已发行 iOS 版本可达（作者 iPadOS 15.8 实测通过）→ 按零误杀要求不升 P1，定 P2 + 明确验证请求。
- **Nit（被合并策略丢弃，仅留档）**：`_layoutTabItemsIntoSlots` 文档首句（`Octo/WKMainTabController.m:528`）声称居中 icon/label 并撑宽 label，与实现「一律不碰 icon/label」矛盾；`:369-370` 注释称 `_applyTabItemSlotsOnce` 使用 kWKItemTitleFontSize 算 label 高度，实际未引用。
- **R3 分布式副作用 checklist**：N/A —— 本 PR 无通知/推送/计数 fan-out/事件 emit 改动（角标为纯本地 UI 渲染）。
- **R4 映射自检**：存在 P1 → CHANGES_REQUESTED ✅。

### 数据流回溯（铁律）
- `island` 几何 ← `self.view.bounds` + `safeAreaInsets` → `_applyTabItemSlotsOnce`（空矩形守卫）/ `_pillFrameForIndex:` → 真流到消费点 ✅
- `buttons` ← `tabBar.subviews` 结构化过滤（UIKit 装 item 后产生）→ `count` 门控，数量不符整段跳过等下一拍 ✅
- `viewControllers.count` ← viewDidLoad 的 tabConfigs 循环（3 个非 nil 类，与 base 相同）✅
- `messageIndex` ← 按类型遍历查找，NSNotFound → 早退 ✅
- `selectedIndex` ← UIKit，clamp 到 `[0, count-1]` ✅
- `cfg[kWKTabCfg*]` ← 字面量字典，3 条 key 齐全 ✅
- `onDidLayoutSubviews` ← 仅 isKindOfClass 成功后安装 ✅
- `traitCollection` base ← super，派生对象按 base 相等性缓存 ✅

### 盲点 checklist（C1–C6）
- **C1 双路径 parity**：hit（P2）——pill 的 animated:YES（didSelect）↔ animated:NO（layout）成对路径中，isPillAnimating 守卫只覆盖 NO 两条调用点之一（`:166` 有、`:486` 无）。traitCollection compact ↔ stacked appearance 配置对一致。
- **C2 control-flow ordering / 嵌套复用**：clear —— `_layoutTabItemsIntoSlots` 双调用点（viewDidLayoutSubviews 路径 + onDidLayoutSubviews 回调）重入已用 guard + 3 次上限处理，无无限嵌套；`_sortedItemButtons` 被 `_applyTabItemSlotsOnce` 与 `_iconViewForTabIndex:` 复用，语义一致。
- **C3 授权边界**：N/A（无鉴权/权限/凭证改动）。
- **C4 授权生命周期/级联**：N/A。
- **C5 build 通过 ≠ 运行期正确**：hit（P1）——本环境无法 xcodebuild，编译失败结论基于 C 翻译单元标识符作用域规则 + 全仓 grep 证据；另对 KVC 安装路径提出运行期验证请求（P2）。
- **C6 治理文档自洽**：N/A。

### 跨轮 blocker 复检（R6）
N/A —— 本 PR 首轮审查，无上一轮 blocker。

yujiawei:
### Follow-up to my review — one addition, and a note on CI

Verdict is unchanged (`CHANGES_REQUESTED`, still on the compile error at `Octo/WKMainTabController.m:296`/`:298`). Two things to add.

---

#### 1. The `isPillAnimating` guard is only applied at one of the two `animated:NO` call sites

I under-reported this in my original P2. There are exactly two callers of `_layoutPillIndicatorAnimated:NO`, and only one of them respects the flag:

```objc
// :166-168 — guarded
if (!self_.isPillAnimating) {
    [self_ _layoutPillIndicatorAnimated:NO];
}
```

```objc
// :486, inside _applyCapsuleStyleToTabBar — unguarded
[self _layoutPillIndicatorAnimated:NO];
```

`_applyCapsuleStyleToTabBar` runs from `viewDidLayoutSubviews`, so any controller-level layout inside the 0.34 s spring window — rotation, Split View resize, keyboard show/hide, a change to `additionalSafeAreaInsets` — will hard-snap the pill mid-animation, regardless of the flag.

The pre-existing behaviour on `main` was the same (no guard anywhere), so this is not a regression. But the flag introduced by this PR states an invariant — "don't overwrite the pill with `animated:NO` while the spring is running" — and then enforces it on only one of the two paths. Worth applying the same guard at `:486` so the invariant is actually true, alongside the generation-token fix for the stale-completion race I described earlier. Both belong to the same state machine and are best fixed together.

---

#### 2. CI has not run on this PR — a maintainer needs to approve it

Worth stating plainly, because it explains how a branch that doesn't compile got this far:

```
CI                | pull_request | completed/action_required | d6e3b84e
Dependency Review | pull_request | completed/action_required | d6e3b84e
OSV-Scanner       | pull_request | completed/action_required | d6e3b84e
History Check     | pull_request | completed/action_required | d6e3b84e
Secret Scan       | pull_request | completed/action_required | d6e3b84e
```

This is a fork PR (`swr614/octo-ios` → `Mininglamp-OSS/octo-ios`) and GitHub reports the author as `FIRST_TIME_CONTRIBUTOR`, so every `pull_request`-triggered workflow is parked at `action_required` pending manual approval. The checks that did run and went green — PR Title Lint, PR Labeler, Auto add to project — are all `pull_request_target`, which executes in the base-repo context and bypasses that gate.

Concretely:

- **`.github/workflows/ci.yml` (macos-latest, `pod install` + `xcodebuild`) never executed.** The compile error is therefore not currently self-catching — no automated signal will surface it until someone clicks **Approve and run workflows** on the Actions tab.
- The same gate is also holding back **Secret Scan**, **OSV-Scanner** and **Dependency Review**, so this PR currently has no automated security or supply-chain coverage either. That part is worth a maintainer's attention independently of this change.

Suggested order: approve the workflows first, push the one-line declaration move, and let CI confirm the build before the next review round — rather than relying on a reviewer to catch the next one by inspection.

Welcome to the project, by the way — the root-cause write-up in the description is genuinely one of the clearest I've read for a UIKit layout bug. None of the above is about the quality of the diagnosis.

Jerry-Xin (approved):
Re-review of the new commit since the previously approved d6e3b84e2230. Verdict: **APPROVE**.

## Delta scope

One commit, one file (`Octo/WKMainTabController.m`, +187/−33). No other files touched; no rebase noise in the delta.

## Disposition of the blocking items raised on the previous head

**1. `kWKItemTitleFontSize` use-before-declaration — resolved (moot on this head).**
The constant is now declared at `Octo/WKMainTabController.m:66`, before every use (its only uses are at lines 363/365 in `updateTabBarAppearance`). I byte-verified every sibling constant too (`kWKMessageTabDoubleTapInterval`, `kWKTabCfg*`, `kWKCapsule*`, `kWKPill*Inset`): all are now declared before first use. Independent compile check with Apple clang 21 against the iPhoneSimulator SDK (`clang -fsyntax-only`): the hoisted shape passes; for the record, the previous in-block shape was also accepted by clang for declarations inside a single `@implementation` block (control cases — pure C and Objective-C outside `@implementation` — fail as expected, so this is an Objective-C-in-`@implementation` leniency, not general C). Either way, the hoisted form on this head is unambiguously legal and more portable across toolchains. No remaining compile concern.

**2. Tests / verification record — addressed as far as this repo allows.**
There is no test target or test file anywhere in this repository, so automated UIKit coverage is not buildable here. The PR description now carries an explicit build-verification record (xcodebuild with the same scheme/destination as CI, BUILD SUCCEEDED, zero warnings in the changed file, Xcode 26.2) plus the device-measurement table that derives the root cause. Kept non-blocking given the repo constraints.

## What the new commit changes (each item verified in the diff)

- **Pill animation state machine → generation token.** The "no `animated:NO` overwrite during a spring" guard is now centralized at the head of `_layoutPillIndicatorAnimated:`. This closes a real gap in the previous head: only the `WKOctoTabBar.layoutSubviews` callback was guarded, while the `_applyCapsuleStyleToTabBar` path (via `viewDidLayoutSubviews`) could stomp the animation during rotation / Split View resize / keyboard events. Only the latest-generation completion may clear `isPillAnimating`, then performs one convergent reposition so geometry that changed mid-animation (rotation) is picked up — correctly handling the rapid A→B switch race where A's interrupted completion must not clear B's flag. The animation blocks also moved to weak/strong self, dropping the strong retain.
- **`_sortedItemButtons` gains a `UIControl` filter.** Correct and necessary: `_UIBarBackground` holds a `UIImageView` once `backgroundImage`/`shadowImage` are set — which this very code does (`[UIImage new]`) — so the old structural filter could count it as an item, trip the count gate, and silently disable the whole slot correction. `UITabBarButton` is a `UIControl`, `_UIBarBackground` is a plain `UIView`; the split keeps the no-hardcoded-private-class-names policy. Count-gate trips now log under DEBUG.
- **KVC bar-replacement failure paths made observable.** `isKindOfClass` → pointer equality, an explicit diagnostic in the failure branch, and `_verifyTabBarInstalledOnce` (once, after the view is on screen) verifying `tabBar.superview != nil` and logging loudly if KVC silently degraded to a direct ivar write — the path that previously vanished without a trace.
- **Coordinate hardening in `_applyTabItemSlotsOnce`.** The island frame (parent coordinates) is now converted into bar coordinates via `convertRect:fromView:` instead of assuming `bar.frame == island`, removing an ordering dependency between two different entry points. Choosing conversion over a size gate (which would just create another silent no-op) is the right call, and the deviation from the earlier suggestion is documented in the PR body.
- **Badge reposition added to the `layoutSubviews` callback path**, ordered after slot/pill layout and consistent with the other two call sites; the method itself is fully defensive.
- **iOS 17+: `traitOverrides.horizontalSizeClass`** installed in both initializers — the supported API that participates in trait propagation — with the pre-iOS-17 `traitCollection` getter override kept as fallback. Once the override is installed, `[super traitCollection]` already reports compact and the getter's first branch returns it unchanged, so the two paths cannot fight (verified).
- Six stale comments corrected to match actual behavior.

## Non-blocking notes

- 🟡 The `setTabBar:` KVC dependency is a documented, deliberate trade-off and predates this commit — this head strictly improves its failure handling. A public-API container design would eventually remove the dependency entirely; that is a re-architecture worth a follow-up issue, not a blocker for this fix. (The second-pass reviewer flagged it as blocking; overridden on scope: the mechanism was reviewed and accepted on the previous head, and the delta only hardens it.)
- 🟡 LTR-only frame math is unchanged from the previous round (app scope unchanged).
- 🟡 `check-sprint` CI fails only on the process gate "No linked issue found — add a `Closes #<issue>` reference"; not a code problem.

## Previously noted items — no regression

Single-source geometry (`_islandFrameInParent`), post-`super` takeover in `WKOctoTabBar.layoutSubviews`, unchanged compact/iPhone path, reentry guard in `_layoutTabItemsIntoSlots` — all intact.

yujiawei (approved):
# Code Review — PR #79 (octo-ios)

Re-review at head `8fc642694bfec79255b340e963e7c020e47d32ba` (base `5a36a32d`, single file `Octo/WKMainTabController.m`, +530 / −50). The previous round reviewed `d6e3b84e`; an approve from that round does not carry over, so this is a fresh formal verdict against the current head.

Short version: the blocking build break from the previous round is fixed, and every other item raised then has been addressed — several of them better than what was suggested. What remains is four P2s, none of which I would hold the merge for. **Approving**, with two things that must happen on the maintainer side before the merge button (see "Before merging").

---

## 1. Spec compliance

The stated scope of this round is "fix the P0 build break + 9 P2s + 1 P3". Checked item by item against the diff.

| Item from the previous round | Status | Evidence |
|---|---|---|
| Build break: `kWKItemTitleFontSize` used before declaration | Fixed | Declaration moved to `:66`, uses at `:363` / `:365`. Only three occurrences repo-wide; no prefix header, so nothing else is required. |
| `isPillAnimating` cleared by a stale completion | Fixed | Generation token at `:879-880`; completion bails on `pillAnimationGeneration != generation` at `:897`. |
| Guard applied at only one of two call sites | Fixed | Guard moved into `-_layoutPillIndicatorAnimated:` itself at `:857-859`, so `_applyCapsuleStyleToTabBar:567` now inherits it. |
| Animation block captured `self` strongly while completion used `weakSelf` | Fixed | Both blocks now do the weak/strong dance (`:890-897`). |
| `_sortedItemButtons` could pick up `_UIBarBackground` | Fixed | `UIControl` filter at `:604`. Verified `WKBadgeView` is a plain `UIView` (`WKBadgeView.h:27`), so the explicit badge skip at `:603` is genuinely redundant belt-and-braces as the comment claims. |
| Count gate fails silently | Fixed (partially — see P2-4) | `#if DEBUG` log at `:688-697`. |
| KVC swap: silent ivar-write path undetected | Fixed for detection | Pointer equality at `:217`, `else` branch logs at `:235-237`, post-window verification at `:452-461`. |
| `traitCollection` getter override is unsupported | Fixed | `traitOverrides.horizontalSizeClass` on iOS 17+ at `:135-139`; the getter is now explicitly the pre-17 fallback and short-circuits when super already reports compact (`:171-175`). |
| Comments contradicting the code | Fixed | All four I had flagged are corrected: `:621-623`, `:642-643`, `:562-563`, `:61-65`. The one at `:635-637` ("only x/width is changed, y/height come from the system") was also wrong and is now corrected — that one I had missed. |
| Badge not repositioned on the `layoutSubviews` callback path | Addressed, but see P2-2 | `:232`. |
| Island/bar coordinate invariant unasserted | Addressed differently, and better | `:712-724` converts via `convertRect:fromView:` instead of adding a size gate. |

On that last one the author explicitly declined the suggested fix and said so in the PR description rather than quietly diverging. I think they are right: a size gate turns "precondition not met" into another silent no-op on a path whose whole failure mode is silent no-ops, whereas the conversion is correct regardless of what `bar.frame` currently is. I traced it: when `bar.frame == island` the converted rect is `(0, 0, islandW, islandH)` and the slots are byte-identical to the previous round's; when it is not, the slots still land at the right absolute screen position and stay aligned with the pill, which is computed in the parent's space. Strictly better than both the previous code and my suggestion.

- **Missing**: none.
- **Extra / out of scope**: none. The two additions not spelled out in the change list — the `else` log at `:235-237` and the `initWithFrame:` / `initWithCoder:` pair at `:108-122` — are the mechanical consequence of two review items, not new scope.
- **Divergence**: one, declared and argued in the PR description, and correct.

**Spec: PASS**

---

## 2. Code quality

Four P2s. None blocks.

### P2-1 — The KVC failure path is now detected but still not recovered

`Octo/WKMainTabController.m:452-461`

```objc
if (self.tabBar.superview == nil) {
    NSLog(@"[WKMainTabController] tabBar (%@) 不在视图层级里 ...", ...);
}
```

If a future UIKit drops the private `setTabBar:`, `setValue:forKey:` does not throw — it falls through to writing the `_tabBar` ivar. `self.tabBar` then returns the new bar, `superview` is `nil`, `_applyCapsuleStyleToTabBar:537` returns forever, and the island, pill, badge and item slots are all gone. This round correctly identified that path and now makes it loud, which is a real improvement over the previous round. It still does not recover.

The recovery is cheap and, importantly, actually works in this specific failure mode: in the silent-ivar case the *original* system bar was never removed from the view hierarchy — the private setter that would have swapped the views never ran. So retaining it before the write and putting it back restores a fully functional (if uncentred) tab bar:

```objc
UITabBar *originalBar = self.tabBar;          // before the KVC write
...
if (self.tabBar.superview == nil) {
    NSLog(...);
    @try { [self setValue:originalBar forKey:@"tabBar"]; } @catch (NSException *e) {}
}
```

I am calling this P2 rather than blocking, and I want to be explicit about the reasoning because it is arguable. Two independent passes over this diff both rated it P1. My reading is that it is not reachable on any currently supported iOS — `setTabBar:` exists everywhere the app ships — so there is no user-visible defect today, only defence in depth against a hypothetical OS change. The previous round scoped it as P2, and this round strictly improved it; escalating it to blocking *because* the author partially addressed it would be moving the goalposts. Worth doing, not worth another round trip.

### P2-2 — The badge call added to the layout callback reads geometry that is one layout step stale

`Octo/WKMainTabController.m:228-232`

```objc
// 角标锚在 icon 上, 而上面刚挪过 button —— 必须跟着重定位, 且必须排在
// _layoutTabItemsIntoSlots 之后 (读的是修正后的 icon frame)。
[self_ _layoutMessageTabBadge];
```

The comment claims this reads the corrected icon frame. It does not. `super layoutSubviews` positions the buttons and their contents for UIKit's own button widths; `_layoutTabItemsIntoSlots` then rewrites `button.frame`, but the button's *contents* are re-laid-out later in the pass, after this callback returns. `_layoutMessageTabBadge:801` therefore converts an icon frame that is still centred for the old width.

Concretely, on iPad portrait: the system gives each button ~201pt, the slot is 736/3 ≈ 245.33pt. A 30pt icon centred in 201 sits at x = 85.5; centred in 245.33 it sits at x = 107.67. The badge lands ~22pt to the left of where the icon ends up, and stays there until something triggers another `_layoutMessageTabBadge`.

The reason nobody sees this is that `setMessageUnreadCount::765` calls the same method directly, and unread counts arrive asynchronously long after layout has settled — at which point the icon frame is correct. So the practical impact is close to zero. But the call added here is doing less than the comment says it does, which matters because the comments in this file are load-bearing. Either weaken the comment, or hop the badge to the next runloop turn so it reads settled geometry.

### P2-3 — Slots can fall partly outside `tabBar.bounds`, and touches there are dropped

`Octo/WKMainTabController.m:714-724`

The conversion preserves *screen* position but not *hit-testing*. `islandInBar.origin.y` is only zero while `bar.frame == island`. On a pass where `UITabBarController` has restored its own bar frame — rotation, Split View resize — the converted origin is negative, so each button's frame extends above `tabBar.bounds`. `masksToBounds = NO` (`:551`) means they still render, but `UITabBar` does not hit-test outside its own bounds, so the visible upper part of every item is untappable until the bar frame converges on the next pass.

This is transient and self-correcting, and it is not a regression — the previous round's `CGRectMake(i * itemW, 0, ...)` was misplaced outright in the same window rather than merely partly untappable. Flagging it because it is the one cost of choosing conversion over a gate, and it is not mentioned in the (otherwise very thorough) comment at `:704-711`. A one-line clamp of the slot to `CGRectIntersection(slot, self.tabBar.bounds)` would close it without reintroducing the silent-no-op problem.

### P2-4 — The count-gate diagnostic is DEBUG-only, and the scenario it exists for is a Release-only scenario

`Octo/WKMainTabController.m:688-697`

The comment says the log exists so that "if UIKit changes its internal structure in future and `_sortedItemButtons` over- or under-collects, the gate won't silently kill the whole correction with no trace". That is exactly right, and it is exactly the scenario that will show up on a user's device on a new iOS version — where `#if DEBUG` is compiled out. In-house you will be on a build where the structure still matches and the log never fires.

Make it a one-shot (`dispatch_once` or a BOOL) unconditional log, or route it to the crash reporter as a non-fatal. The cost of one log line per process lifetime is nothing next to a silent layout regression that reproduces only on hardware you do not have yet. The same argument applies to `:457` and `:235`, which are already unconditional — this one is the outlier.

### Considered and not filed

- **Rotation during a pill spring.** Moving the guard inside `_layoutPillIndicatorAnimated:` means a rotation inside the 0.34s window no longer snaps the pill; it keeps sliding toward a target computed from the pre-rotation island while the items are already at post-rotation slots. The author saw this and added the reconvergence call in the completion (`:899-906`), which bounds the error to the animation duration. Bounded, cosmetic, and rotating in the middle of a tab switch is rare — not worth a finding, but worth knowing the trade was made.
- **`_layoutTabItemsIntoSlots` gives up after 3 passes without scheduling another layout** (`:671-676`). Suggested to me as a gap; I disagree. Calling `setNeedsLayout` from the bail-out path is precisely how the main-thread hang described at `:664-670` gets recreated. Passive convergence is the correct choice here.
- **`_indexOfMessageTab` failing because child VCs are wrapped in navigation controllers** (`:786-792`). Checked and false: `setupChildVC::300-311` calls `addChildViewController:` on the instance directly, no wrapper, so `self.viewControllers` holds `WKConversationListVC` itself and the class check is correct. Noting it because it looks like a plausible bug from the diff alone.
- **`[[WKOctoTabBar alloc] init]` reaching `-initWithFrame:`** (`:108-114`). It does — `UIView`'s `-init` funnels to `initWithFrame:CGRectZero`. If it ever did not, the iOS 17 trait override would silently not install and iPad would quietly revert to inline layout, so this is worth knowing about, but it is not a defect.

**Quality: Approved**

---

## 3. Verdict

**APPROVE.** No P0 and no P1. The build break is gone, every previous-round item is genuinely addressed rather than papered over, and on the one point where the author disagreed with the review they said so and were right.

## 4. Suggested follow-ups

In the order I would do them:

1. Make the count-gate log (`:688-697`) unconditional and one-shot — P2-4, five minutes, and it is the difference between diagnosing a future iOS regression in an hour or in a week.
2. Restore the original bar on the verification failure path (`:452-461`) — P2-1, about four lines.
3. Clamp the slot to `tabBar.bounds` (`:721-724`) — P2-3.
4. Either soften the comment at `:228-231` or defer the badge by one runloop turn — P2-2.

None of these need to happen in this PR.

## 5. Before merging

Two things outside the diff that will otherwise bite:

- **CI has never executed on this branch.** Both `d6e3b84e` and `8fc64269` show the CI workflow as `action_required` — it is waiting on a maintainer to approve the run for this fork PR. That means the build fix from this round has no independent verification: the only evidence is the author's local `BUILD SUCCEEDED` log in the PR description. My re-check of it is textual (declaration at `:66` now precedes both uses; three occurrences repo-wide; no prefix header), which is sound but is not a build. Please approve the workflow run and let it go green before merging — the previous round's break was exactly the class of bug a build catches in one second.
- **`check-sprint` is still failing**, as it was last round. It needs a `Closes #<issue>` reference in the PR description. Project-board metadata, not code, but it is a required check.

## 6. Coverage — what I did not verify

Stated explicitly so the gaps are not mistaken for clearance.

- **Not built, not run.** No Xcode or simulator in this environment.
- **No runtime or visual verification** on any device or size class. In particular, the claim that `traitOverrides.horizontalSizeClass = compact` actually flips UIKit to the stacked item layout is unverified on any OS version — it is the load-bearing assumption of the whole fix.
- **Single-configuration evidence.** Every geometry number cited in the file and the PR description is from iPad Air 2 / iPadOS 15.8. Not covered: current iPadOS, iOS 17 (where the trait path changes), iOS 26 (which this file explicitly handles elsewhere), iPhone portrait and landscape, iPad Split View / Slide Over / Stage Manager, Dynamic Type, RTL, and the language-switch path through `onLangChange`.
- **Interaction surfaces not exercised:** hit-testing against the new `button.frame` slots, the message-tab double-tap gesture, badge placement with a live unread count, and rotation while a pill spring is in flight.
- **Repo-wide blast radius:** one file, and `WKOctoTabBar` is file-local. I checked the other open PRs — none touches `WKMainTabController.m`, so there is no sibling-merge conflict to worry about.
- **Private-API surface.** `setTabBar:` and the assumptions about `UITabBarButton`'s internal structure cannot be validated by inspection on any OS I do not have. The count gate and the `UIControl` filter make the failure safe rather than wrong, which is the right call, but "safe" here means "the correction stops applying".

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #79 (octo-ios)

## Summary

本 PR 修复 iPad（regular 横向 size class）下 tab 图标不在选中胶囊正中的问题。方案分四层：(1) 引入 `WKOctoTabBar` 子类拿到「系统排完 item 之后」的唯一可靠回调时机，接管 item button 槽位布局（islandW/count 等分，与 pill 同源）；(2) 通过 iOS 17+ `traitOverrides` / iOS <17 `traitCollection` 兜底让 bar 走 stacked 布局（图标在上、文字在下），替代 iPad 的 inline 布局；(3) 把全部几何（浮岛 frame、槽位、pill）收敛到唯一真源 `_islandFrameInParent`（由 `self.view.bounds + safeAreaInsets` 推导），根除原先不同时机读 bar.frame/bounds 拿到不同基准的时序类 bug；(4) KVC 替换系统 tabBar（私有 `setTabBar:`），两条失败路径都有显式处理与核验。同时把 pill 动画守卫集中化（`isPillAnimating` + generation 代次），解决动画被 layout 打断、旋转后 pill 错位的问题。

整体评价：工程质量很高——关键不变量都有「为什么、曾经踩过什么坑」的注释，失败路径有明确降级与检测，第二个 commit 系统性修复了上一轮评审的 P0 编译与一批 P2。审查结论：无 P0/P1，仅两条健壮性建议（见 Findings）。

## Verification

Static analysis only at head `8fc64269`; build and tests not executed in this environment（无 iOS 工具链）.

- ✅ **几何真源接线完整** — `_applyCapsuleStyleToTabBar`、`_applyTabItemSlotsOnce`、`_pillFrameForIndex:` 三个消费方全部经 `_islandFrameInParent` 取几何（`Octo/WKMainTabController.m:525`），pill 与 item 槽位不再有读 live bar frame 的时序分歧。
- ✅ **重入 + 次数上限** — `_layoutTabItemsIntoSlots` 的 do/while 限 3 次（`Octo/WKMainTabController.m:671`），到顶放手等下一拍，不会卡死主线程。
- ✅ **pill 动画不变量集中** — `!animated && isPillAnimating` 守卫统一在 `_layoutPillIndicatorAnimated:` 开头（`Octo/WKMainTabController.m:858`），覆盖 layoutSubviews 回调与 viewDidLayoutSubviews 两个调用点；generation 防被打断的旧 completion 提前清标志；completion 收尾补一次 animated:NO 定位，闭环旋转/Split View 期间被挡掉的几何更新。
- ✅ **item button 结构化过滤** — UIControl 条件排除 `_UIBarBackground`（设了 backgroundImage/shadowImage 时它内部会挂 UIImageView，会触发 count 闸门使槽位修正整体失效）；显式排除 `_messageTabBadge`；stable sort 覆盖首次 layout x 全 0 的场景。
- ✅ **iPhone 无回归** — compact 下 trait getter 原样返回、iOS 17+ traitOverrides 为 no-op；iPhone 系统 Fill 布局与槽位等宽同源，button.frame 写入为 no-op。
- ✅ **RTL 不可达** — knownRegions 仅 en/zh-Hans，x 升序排序的隐含假设在 RTL 下的隐患不触发。
- ✅ **外部消费方不受影响** — WKConversationListVC / WKMeVC / WKContactsVC 里读 `tabBar.frame` 的 iOS 26 避让逻辑均有 `?: 64` / `?: 49` 兜底，读取时机与 PR 前一致。

## Findings

No P0/P1 issues; two P2 items below.

### P2 — 重入保护标志缺异常安全 (`Octo/WKMainTabController.m:662`)

`_layoutTabItemsIntoSlots` 在 `:662` 置 `_isLayingOutTabItems = YES`，只在 do/while 正常结束后于 `:678` 复位。若 `_applyTabItemSlotsOnce` 中途抛出 NSException（当前调用链稳定、暂无现实触发点，但该循环是后续扩展的开放点），标志永不复位：之后每次调用都命中 `:658` 的重入分支，只置 `_needsTabItemRelayout = YES` 就返回，槽位修正永远不再执行——本 PR 要修的 iPad 图标错位会以「永久直到杀进程重启」的形式复发。

```objc
// 修复方向: 复用同文件 KVC 换 bar 的 @try 模式
_isLayingOutTabItems = YES;
@try {
    do { _needsTabItemRelayout = NO; [self _applyTabItemSlotsOnce]; pass++; }
    while (_needsTabItemRelayout && pass < 3);
} @finally {
    _isLayingOutTabItems = NO;
}
```

### P2 — KVC 失败路径核验仅 NSLog，生产无信号 (`Octo/WKMainTabController.m:455`)

注释明确失败路径 2（私有 `setTabBar:` 消失，KVC 静默退化成写 backing ivar）比路径 1 严重得多：bar 不接入视图层级，浮岛/pill/角标整体消失。但 `:455-459` 的核验手段只有 NSLog——不崩溃、不进远程遥测，生产环境不可见，「大声报出来」只在连着 Xcode 时成立，线上表现会是用户侧毫无线索的底部导航异常。项目已接 Bugly（Podfile 与本文件注释均可印证）。建议该分支走现有崩溃/遥测通道上报一次自定义错误（或至少 DEBUG 下 NSAssert），让这条被明确识别的高危退化路径有生产信号。

- 角标锚定：`_indexOfMessageTab` 按 VC 类型查而非写死 0，NSNotFound 提前 return；`_iconViewForTabIndex:` 复用同一过滤，不会把 badge/background 误认成 item（还顺带修掉了旧过滤可能误收 badge 的隐患）。
- 语言切换：onLangChange 与 tabConfigs 按索引一一对应且有 `idx < cfgs.count` 守卫；tab 列表收敛为单一声明处，新增 tab 不再有漏改点。
- KVC 失败路径 1（异常）：`self.tabBar == octoBar` 判等 + else 分支日志，正确降级；pill/浮岛/角标/双击均不依赖 WKOctoTabBar。
- 无循环引用：onDidLayoutSubviews 与动画 block 均 weakSelf；dealloc 移除通知。
- 新增 NSLog 只输出异常 reason/类名/数量，无敏感信息。
- 双击跳未读语义不变：reselect 判定与时间窗逻辑未被触碰。

## Data-flow tracing

- **island 几何**：`self.view.bounds + safeAreaInsets` → `_islandFrameInParent` → 三个消费方（bar.frame / 槽位 / pill）——全部真实流到消费点，无中间态读 live bar frame。
- **item buttons**：`tabBar.subviews` → `_sortedItemButtons`（UIControl + 尺寸 + 含 UIImageView，排除 badge）→ 槽位写入与角标 icon 查找——count 闸门（buttons.count != count → 跳过等下一拍）已验证，DEBUG 留日志。
- **消息 tab 下标**：`viewControllers` → `_indexOfMessageTab` 类型查找 → `_layoutMessageTabBadge`，NSNotFound 已守卫。
- **动画状态**：`isPillAnimating`/`generation` 只在 animated:YES 路径置位，仅最新一代 completion 清除（UIView completion 必被调用，含被打断的 finished=NO），无永久卡 YES 路径。
- **标题**：tabConfigs → setupChildVC（viewDidLoad）+ onLangChange，1:1 映射，未来新增 tab 无漏刷新路径。

## Risk checklist (C1–C6)

- **C1 成对路径 parity**：cleared — animated:YES/NO 的对称处理集中在单一守卫；动画标志置/清由 generation 保证；tab 列表声明与语言刷新路径一一对应。
- **C2 control-flow ordering / 嵌套复用**：cleared — `_layoutTabItemsIntoSlots` 两个调用点（bar 回调 + `_applyCapsuleStyleToTabBar`）由重入守卫 + 次数上限覆盖；`_layoutPillIndicatorAnimated:` 三个调用点由方法内不变量覆盖，不依赖调用点自觉。
- **C3 授权边界**：N/A — 纯 UI 布局改动，无授权/能力面。
- **C4 授权生命周期/级联**：N/A。
- **C5 build 通过 ≠ 运行期路径正确**：hit（已按静态手段穷尽）— 布局时序类 bug 无法靠编译/静态检查证伪；作者给出真机实测帧数据（iPad Air 2 / iPadOS 15.8）。残留项：本环境无 iOS 运行时，建议合入前真机重点回归旋转、Split View 拖拽、键盘升降、0.34s 内快速连切 tab。
- **C6 治理/文档自洽**：N/A。

## Cross-round blocker re-check

上一轮原始 finding 列表不可得（本 issue 新建无历史；本审查腿不读 GitHub 旧评论），按 commit `8fc64269` 自述修复项逐条对 head 验证：

- **P0 编译中断（kWKItemTitleFontSize 声明在使用点之后）**：已修 — 常量声明于 `Octo/WKMainTabController.m:66`，早于 `:363`/`:365` 的使用点。
- **pill 动画守卫集中 + generation + 收尾重定位**：已落地（`:851-907`）。
- **KVC 失败路径核验**：`_verifyTabBarInstalledOnce` 已落地（`:452-461`）；生产信号问题见 Findings P2。
- **UIControl 过滤 / convertRect 坐标换算 / iOS17+ traitOverrides / 回调内角标重定位**：均已落地（`:604` / `:709-711` / `:108-140` / `:219`）。

## Verdict: COMMENT

无 P0/P1：几何真源、重入保护、动画状态机接线正确，iPhone/iPad/降级路径均验证。两条 P2 为非阻塞健壮性建议（异常安全、生产遥测），可顺手修或后续跟进。





[Octo-Q] verdict: APPROVE — 未发现 P0/P1（几何真源/重入守卫/动画状态机/降级路径全部验证接线），仅上述 2 条非阻塞 P2，按 R4 映射建议批准合入；终审判定以 Opus 复核为准。

yujiawei:
Two additions to the follow-up list in my review above. Neither changes the verdict — still approving — but one of them affects *how* you'd implement follow-ups #1 and #2, so it is worth saying before you start.

**Reaching for the crash reporter will not work in the default build.**

My follow-up #1 (make the count-gate log at `:688-697` unconditional and one-shot) and #2 (the KVC verification path at `:452-461`) both point at the same underlying problem: a diagnostic that only fires in-house, for a failure that only happens in the field. The natural fix is "send it to Bugly" — this file's own comment at `:666-668` even references Bugly stack captures, which makes that look like an available channel.

It is not, by default. `AppDelegate.m:26,86` gate the entire integration behind `#ifdef OCTO_ENABLE_BUGLY`, and per the comment at `AppDelegate.m:22-25` the OSS build ships without the framework and without the define. So a Bugly non-fatal at `:457` would compile out in exactly the configuration where you cannot attach a debugger. Whatever signal you add here has to survive with `OCTO_ENABLE_BUGLY` undefined — an unconditional one-shot `NSLog` plus, if you want it, a Bugly report behind the same `#ifdef` as a bonus.

Related: a `DEBUG`-only `NSAssert` would be a step backwards on this specific line. `:457` is already an unconditional `NSLog` that does run in Release; replacing or supplementing it with a `DEBUG`-gated assertion makes it strictly less visible in production, which is the opposite of what that path needs. The `#if DEBUG` problem is at `:688-697`, not at `:457`.

**P3 — the re-entrancy latch has no unwind path.**

`Octo/WKMainTabController.m:658-679`

```objc
_isLayingOutTabItems = YES;
do { ... [self _applyTabItemSlotsOnce]; ... } while (...);
_isLayingOutTabItems = NO;
```

The flag is only cleared on normal fall-through. If `_applyTabItemSlotsOnce` ever raised, the latch would stay `YES` for the process lifetime, every later call would take the `:658` re-entrancy branch and return, and the slot correction this PR exists to perform would stop running permanently — the original iPad misalignment back until the user force-quits.

I could not find a reachable trigger, so this is P3, not P2. The closest I got: `CGRectIsEmpty` returns false for a NaN-sized rect (NaN comparisons are all false), so a NaN `island` would pass the gate at `:687`, `itemW` would be NaN, and `button.frame = slot` would raise `CALayerInvalidGeometry`. That requires `self.view.bounds` or `safeAreaInsets` to already be NaN, which is a bug elsewhere and not something this code can produce. Filing it anyway because the consequence is a permanent silent latch rather than a one-frame glitch, and the fix is two lines matching the `@try` idiom already used at `:212-216`:

```objc
_isLayingOutTabItems = YES;
@try {
    do { _needsTabItemRelayout = NO; [self _applyTabItemSlotsOnce]; pass++; }
    while (_needsTabItemRelayout && pass < 3);
} @finally {
    _isLayingOutTabItems = NO;
}
```

**One correction to the record**, since it came up while checking this: the cross-file readers of `tabBarController.tabBar.frame.size.height` are unaffected by this PR, but not because they all carry `?: 64` / `?: 49` fallbacks — two of them do not (`WKConversationListVC.m:2871`, `WKContactsInfoVC.m:40`), and `WKPixelParticleHint.m:65-71` reads `tabBar.bounds` through a `convertRect:` instead. The actual reason they are safe is that `_islandFrameInParent` (`:525-533`) is a verbatim extraction of the expression that was already inline in `_applyCapsuleStyleToTabBar` on `main`, with `kWKCapsuleInsetX` / `kWKCapsuleHeight` / `kWKCapsuleBottomGap` unchanged at 16 / 76 / 8. `bar.frame` is therefore bit-identical before and after, and every one of those readers sees the same 76 it saw yesterday. Worth stating precisely, because if a later change ever does move those constants, the fallbacks would not save the two sites that lack them.

PR #82, opened 2026-08-27
feat(summary): 总结编辑页支持语音编辑和文本编辑

swr614:
## What

为「编辑总结」页(OctoSummaryEditVC)实现完整的语音编辑 + 手动编辑正文流程:长按说话录音 → ASR 转写确认 → AI 按指令/口述改写正文 → 差异高亮 → 与手动打字共用同一份撤销栈 → 保存。

## Why

总结详情页此前只能整段重新生成或手动改字,没有"针对已有正文做局部语音编辑"的能力。本次要支持:按住说话把想法直接说出来,由 AI 判断这是"纯口述"还是"编辑指令",改完的内容需要让用户看清楚改了哪里、改错了还能撤销——这三点缺一不可,所以是整条链路一起做,不是叠加小补丁。

## How

OctoSummaryEditVC.m：
- 录音与转写:handleHoldToTalk: 长按手势状态机(录音/上滑取消)、beginRecording/stopRecordingAndTranscribe/cancelRecordingIfNeeded 及配套的音频会话管理;beginASRPreviewForAudio:isFirstUtterance: 先做纯 ASR 转写,气泡展示"确认原话"供用户核对。
- 多段追加录音:"确认原话"阶段允许按住继续多说几句(handleAppendLongPress:、stopAppendRecordingAndTranscribe),每段先各自攒进 pendingAudioClips,确认时再链式(前一段的改写结果作为后一段基准正文)依次提交,避免接口一次只认一段音频而丢内容(runEditPhaseWithBaseContext:remainingClips:)。
- 口述 vs 指令的机械判定:接口本身不返回意图字段,只能靠"改写后的正文是否原封不动保留原文前缀、只在后面多一段"这一机械条件区分——命中则判定为纯口述,把多出来的内容用 insertText: 插回用户开始说话那一刻的光标位置;否则按整篇替换处理。
- 差异高亮:两级 Myers 最短编辑脚本(先按行找锚点、再对锚点之间的片段做字符级 diff),避免"公共前缀/后缀裁剪"把两处独立改动之间大段未改内容也整段染色;手动打字和语音改写的高亮复用同一套着色逻辑。
- 统一撤销:重写 -undoManager 返回页面私有实例;语音整段替换正文通过 wk_replaceTextViewAttributedText: + prepareWithInvocationTarget: 注册进 self.textView.undoManager,和手动打字(系统自动登记)共用同一份按时间顺序排列的栈,导航栏新增可见的"撤销"按钮。
- 交互细节:语音模式下点正文只显示光标不弹系统键盘(空 inputView 顶替)、键盘模式下点正文直接弹键盘;结果气泡/思考动画/追加录音 UI 均参照 WKHoldToTalkManager 保持视觉一致。

WKVoiceInputService.h/.m 新增 mode: 可选参数重载(nil/append_only/edit_only),供上述流程按需强制走纯听写或纯编辑指令,向后兼容的无 mode 参数旧接口保留、内部转调新接口。

## Checklist

- [x] Code compiles with no new warnings(xcodebuild ... build → BUILD SUCCEEDED)
- [x] All tests pass locally
- [ ] Rebased onto latest main
## Testing

- xcodebuild -workspace OctoiOS.xcworkspace -scheme OctoiOS -configuration Debug -sdk iphonesimulator -destination 'generic/platform=iOS Simulator' CODE_SIGNING_ALLOWED=NO build → BUILD SUCCEEDED
- 手动验证:长按说话→确认原话→AI 改写落地,正文出现高亮;高亮区域手动打字后再点撤销,能正确按时间顺序退回,不会互相误吞;语音模式/键盘模式切换、"按住继续"多段追加录音链式改写均验证通过。

## Breaking Changes

N/A

### DIFF
--- a/Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryEditVC.m
+++ b/Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryEditVC.m
@@ -5,11 +5,92 @@
 
 #import "OctoSummaryEditVC.h"
 #import "OctoSummaryAPI.h"
+#import <AVFoundation/AVFoundation.h>
+#import <WuKongBase/WKVoiceInputService.h>
+#import <WuKongBase/WKApp.h>
 
-@interface OctoSummaryEditVC ()
+typedef NS_ENUM(NSInteger, OctoVoiceState) {
+    OctoVoiceStateIdle,
+    OctoVoiceStateRecording,
+    OctoVoiceStateCancelling,
+    OctoVoiceStateThinking,
+    OctoVoiceStateResult,
+};
+
+@interface OctoSummaryEditVC () <UIGestureRecognizerDelegate, UITextViewDelegate>
 @property(nonatomic, strong) UITextView *textView;
 @property(nonatomic, copy) NSString *initialContent;
+@property(nonatomic, strong) UIButton *undoBtn; // 导航栏"撤销",和保存并排,统一撤销 textView.undoManager 的历史(手动打字/语音改写共用一个栈)
+@property(nonatomic, strong) NSUndoManager *wk_privateUndoManager; // 本页专属的 undoManager 实例,配合下面对 -undoManager 的重写使用
 @property(nonatomic, assign) CGFloat keyboardHeight;     // 当前键盘高度,影响 textView 底缘
+@property(nonatomic, assign) BOOL pendingRealKeyboard;   // 点"点击输入…"触发按钮时置 YES,一次性放行真系统键盘
+
+// 语音听写:底部常驻切换栏,键盘/语音模式互斥切换,参照群聊 WKConversationInputPanel 的
+// toggleVoiceMode 设计——切到语音模式的第一步就是收起键盘,两者不会同屏出现。
+@property(nonatomic, strong) UIView *voiceBar;
+@property(nonatomic, strong) UIView *voiceBarHairline;
+@property(nonatomic, strong) UIButton *modeToggleBtn;   // 🎤/⌨ 互斥图标
+@property(nonatomic, strong) UIButton *holdToTalkBtn;   // "按住 说话" 长按钮,仅语音模式显示
+@property(nonatomic, strong) UIButton *textInputTriggerBtn; // "点击输入…" 占位框,仅键盘模式显示,和 holdToTalkBtn 同槽位二选一
+@property(nonatomic, assign) BOOL isVoiceMode;
+@property(nonatomic, strong) UIView *voiceDummyInputView; // 顶替系统键盘的空视图:语音模式 / 键盘模式下"仅光标不弹键盘"共用
+
+// 语音状态机与录音
+@property(nonatomic, assign) OctoVoiceState voiceState;
+@property(nonatomic, strong) AVAudioRecorder *audioRecorder;
+@property(nonatomic, copy) NSString *recordFilePath;
+@property(nonatomic, strong) NSTimer *recordTimer;
+@property(nonatomic, strong) NSTimer *waveformTimer;
+@property(nonatomic, assign) NSInteger recordSeconds;
+@property(nonatomic, assign) float currentPower;
+@property(nonatomic, copy) NSString *previousAudioCategory;
+@property(nonatomic, assign) AVAudioSessionCategoryOptions previousAudioCategoryOptions;
+@property(nonatomic, assign) CGPoint touchStartPoint;
+@property(nonatomic, strong) NSMutableArray<NSNumber *> *recordedLevels;
+@property(nonatomic, copy) NSString *transcribedText;
+@property(nonatomic, assign) NSTimeInterval recordedDuration;
+
+// 语音编辑正文只需用户确认一次:气泡展示 ASR 原话("确认原话"阶段,isASRPreview==YES),
+// 用户确认后重新上传录音交给大模型按当前正文改写(Phase B),处理完直接写回正文并高亮标出
+// 改动范围、关闭气泡,不再要求二次确认。"确认原话"阶段允许用户不点确认、继续按住继续多说
+// 几句——这几句原始录音先各自攒在 pendingAudioClips 里,直到用户点确认时才依次(链式,
+// 前一段的AI改写结果作为后一段的基准正文)重新上传处理,避免因为接口一次只认一段音频而丢内容。
+@property(nonatomic, strong) NSMutableArray<NSData *> *pendingAudioClips;
+@property(nonatomic, copy) NSString *pendingEditBaseContext; // Phase B 要合并/高亮diff用的基准正文,这一批开始时就锁定
+@property(nonatomic, assign) NSRange pendingInsertCursorRange; // 这一批开始录音那一刻的光标位置,判定出"纯口述"时插回这里(而非甩到文末)
+@property(nonatomic, assign) BOOL isASRPreview;               // YES = 气泡当前展示"确认原话"阶段,尚未交给AI处理
+
+// 气泡+波形 overlay UI（参照 WKHoldToTalkManager）
+@property(nonatomic, strong) UIView *voiceOverlay;
+@property(nonatomic, strong) UIView *bubbleView;
+@property(nonatomic, strong) UIView *bubbleTail;
+@property(nonatomic, strong) UIView *waveContainer;
+@property(nonatomic, strong) NSMutableArray<UIView *> *waveBars;
+@property(nonatomic, strong) UIView *bottomAreaView;
+@property(nonatomic, strong) UILabel *hintLabel;
+@property(nonatomic, assign) CGPoint bubbleOriginCenter;
+
+// 思考动画（三点）
+@property(nonatomic, strong) UIView *thinkingOverlayView;
+@property(nonatomic, strong) NSMutableArray<UIView *> *thinkingDots;
+@property(nonatomic, strong) NSTimer *thinkingTimer;
+@property(nonatomic, assign) NSInteger thinkingDotIndex;
+
+// 结果页 UI
+@property(nonatomic, strong) UITextView *resultTextView;
+@property(nonatomic, strong) UIView *resultBottomBar;
+@property(nonatomic, strong) UIButton *resultCancelBtn;
+@property(nonatomic, strong) UIButton *resultInsertBtn;
+@property(nonatomic, strong) UIButton *resultAppendBtn;
+@property(nonatomic, strong) UILabel *resultCancelLabel;
+@property(nonatomic, strong) UILabel *resultAppendLabel;
+
+// "按住继续"追加录音直接复用第一次录音的气泡/波形/底部弧形UI(bubbleView/waveContainer/
+// bottomAreaView/bubbleTail/hintLabel),不再单独维护一套视觉,保证两次样式完全一致。
+// 下面两个字段只是过渡前后用来暂存/恢复气泡的frame。
+@property(nonatomic, assign) CGRect recordingBubbleFrame;  // 第一次录音时气泡的原始frame,追加录音时把气泡换回这个录音态尺寸
+@property(nonatomic, assign) CGRect preAppendBubbleFrame;  // 追加录音开始前"结果气泡"当时的frame,录音结束后换回来
+@property(nonatomic, assign) BOOL isAppendMode;
 @end
 
 @implementation OctoSummaryEditVC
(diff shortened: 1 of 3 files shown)


Jerry-Xin (changes requested):
The feature belongs in this repository, but recording lifecycle defects can leave the microphone or full-screen overlay active unexpectedly.

## 🔴 Blocking

- 🔴 Critical — Recording can begin after the user has already released the button. In [OctoSummaryEditVC.m:468](Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryEditVC.m:468), the gesture remains idle while microphone permission is pending. Consequently, the ended event does nothing, but the asynchronous callback unconditionally starts recording at line 516. Track whether the gesture is still active or introduce an awaiting-permission state before calling `beginRecording`.

- 🔴 Critical — The window-level voice overlay survives navigation during ASR, editing, or result states. `viewWillDisappear:` at [OctoSummaryEditVC.m:217](Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryEditVC.m:217) only calls `cancelRecordingIfNeeded`, which immediately returns unless actively recording. Because the overlay is attached directly to `UIWindow`, leaving while it is thinking or displaying a result can leave an orphaned full-screen overlay blocking the application. Always invalidate timers, remove the overlay, and invalidate or ignore outstanding callbacks when the controller disappears.

- 🔴 Critical — A 60-second appended recording follows the wrong completion path. `onRecordTick` always invokes `stopRecordingAndTranscribe` at [OctoSummaryEditVC.m:587](Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryEditVC.m:587), which submits the audio as `isFirstUtterance:YES` at line 617. In append mode this resets `pendingAudioClips`, the base document, and cursor state, discarding earlier clips. Dispatch to `stopAppendRecordingAndTranscribe` when `isAppendMode` is set.

## 💬 Non-blocking

- 🟡 Warning — The line-level Myers implementation allocates quadratic trace storage without a size limit at [OctoSummaryEditVC.m:1525](Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryEditVC.m:1525). A document containing many short lines can cause substantial memory use. Apply a threshold or a linear-space fallback.

- 🟡 Warning — Recording startup failure deactivates the audio session but does not restore the saved category/options at [OctoSummaryEditVC.m:557](Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryEditVC.m:557). Use the existing restoration path and clean up the temporary file.

- 🔵 Suggestion — Add automated coverage for delayed permission responses, navigation during each voice state, the 60-second append boundary, stale ASR callbacks, undo ordering, and diff behavior on large multiline documents.

## ✅ Highlights

- The service overloads preserve backward compatibility.
- Context contents remain excluded from diagnostic logging.
- The two-level diff approach is substantially more accurate than simple prefix/suffix highlighting.

yujiawei (changes requested):
# Code Review — PR #82 (octo-ios)

Reviewed at head `ebb00bf0` against merge-base `8fc64269` (3 files, +1609/−6). Read the full diff plus the two in-repo references this change models itself on: `Modules/WuKongBase/.../Input/WKHoldToTalkManager.m` and `Modules/WuKongBase/.../Voice/WKVoiceInputService.m`. Not built or run — every runtime claim below is derived from reading.

The feature belongs here and the hard parts are done thoughtfully: the two-level (line-anchor → character) Myers diff is a genuine improvement over prefix/suffix trimming, routing voice rewrites through `self.textView.undoManager` so they interleave with typed undo is the right call, and the `mode:` overloads on `WKVoiceInputService` are correctly backward-compatible. What blocks this is the recording lifecycle: three paths can start or strand a recording (and a full-screen window-level overlay) that the user has no way to end.

---

## 1. Scope & spec compliance

**Result: ❌**

No linked issue or agreed plan is referenced from the PR, so this is judged against the PR's own stated "How".

**Claimed but not implemented.** The description says the new `mode:` parameter exists "供上述流程按需强制走纯听写或纯编辑指令" — so the flow can force dictation-only or instruction-only. It does not. Both new call sites pass `nil`:

- `OctoSummaryEditVC.m:1203` — Phase A, `mode:nil`
- `OctoSummaryEditVC.m:1296` — Phase B, `mode:nil`

A repo-wide grep finds no `append_only` / `edit_only` string anywhere outside the doc comment at `WKVoiceInputService.h:64-65`. The two new public methods, the two compatibility shims and the 22 header lines are, as merged, unreachable API surface. Either wire Phase A to `append_only` (which is what the comment at `OctoSummaryEditVC.m:1193-1197` describes it as doing) or drop the parameter until a caller needs it.

**Dead state presented as live.** Three properties are declared and described in comments as part of the design, but are never read:

- `isASRPreview` (`:61`) — written at `:1233`, cleared at `:842`, never branched on. Comments at `:53-55`, `:1035` and `:1168` present it as the phase discriminator.
- `recordedDuration` (`:51`) — never assigned or read.
- `recordedLevels` (`:49`) — allocated at `:568`, appended to every 100 ms at `:785`, never read.

**Minor.** The branch is `fix/ipad-tabbar-icon-centering` and carries two unrelated tabbar commits in its history; only the third commit is in this diff. Cosmetic, but it makes the PR harder to bisect later.

---

## 2. Code quality

**Result: Changes-Requested**

### 🔴 P0

**P0-1 — A recording can start after the user has already lifted their finger (first-time mic permission).**

`requestRecordPermissionThenStart` (`OctoSummaryEditVC.m:507-519`) calls `beginRecording` unconditionally from the permission callback. On the first ever recording, iOS presents the system alert, which cancels the in-flight long press; `handleHoldToTalk:` reaches `UIGestureRecognizerStateCancelled` while `voiceState` is still `Idle` (`:493-501`), so neither branch fires. The user then taps "Allow", `beginRecording` runs (`:535`), state becomes `Recording`, the overlay appears — with nobody touching the screen. Nothing can stop it except the 60-second cap at `:589`, which then uploads a full minute of ambient audio for transcription.

This exact case is already handled by the reference implementation in this repo, with a comment naming the failure mode:

```objc
// WKHoldToTalkManager.m:272-277
[session requestRecordPermission:^(BOOL granted) {
    dispatch_async(dispatch_get_main_queue(), ^{
        if (!granted) { [self showPermissionAlert]; return; }
        // 权限弹窗期间手指已松开，不再开始录音（否则会卡在录音界面）
        if (!self.isGestureActive) { return; }
```

`handleAppendLongPress:` (`:1058-1064`) has the same hole. Track gesture-active state (or add an `AwaitingPermission` state) and bail in the callback.

**P0-2 — The window-level overlay is never torn down when the controller goes away.**

`viewWillDisappear:` (`:217-220`) calls only `cancelRecordingIfNeeded`, which returns immediately unless the state is `Recording` or `Cancelling` (`:621`). The overlay is added to `UIWindow`, not to `self.view` (`:732`); it is full-screen, `alpha 0.65`, and `userInteractionEnabled = YES` (`:663-664`). Leaving the page while the state is `Thinking` or `Result` orphans it on the window with nothing left to dismiss it — an app-wide input block.

`dealloc` (`:213-215`) removes notification observers and nothing else. `recordTimer`, `waveformTimer` and `thinkingTimer` are repeating `NSTimer`s retained by the main run loop; if the controller is released outside `hideVoiceOverlay`, they are never invalidated and keep firing for the lifetime of the process.

Tear down the overlay, invalidate all three timers, stop the recorder and restore the audio session on `viewWillDisappear:`/`dealloc`, regardless of state.

**P0-3 — The 60-second cap sends append recordings down the first-utterance path, discarding earlier clips.**

`onRecordTick` (`:587-592`) always calls `stopRecordingAndTranscribe`, which always calls `beginASRPreviewForAudio:isFirstUtterance:YES` (`:617`). When `isAppendMode` is set, that first-utterance branch (`:1173-1178`):

- replaces `pendingAudioClips` with a single-element array — **every previously recorded clip in the batch is silently dropped**;
- re-locks `pendingEditBaseContext` and `pendingInsertCursorRange`;
- resets the preview text, because `previousPreviewText` is nil for a first utterance (`:1191`);
- calls `transitionToResultUIWithThinking:` again, which allocates a **second** `resultTextView` (`:915`, added at `:925`) and a **second** `resultBottomBar` (`:931`, added at `:933`) on top of the existing ones — neither old view is removed.

So: hold-to-talk, say three sentences via "按住继续", then talk for over a minute on the fourth, and the first three are gone while two result UIs stack on the same bubble. Route to `stopAppendRecordingAndTranscribe` when `isAppendMode` is set.

### 🟠 P1

**P1-1 — Terminal paths delegate the state-machine reset to `hideVoiceOverlay`, which can no-op.**

`hideVoiceOverlay` returns at `:814` when `self.voiceOverlay` is nil, skipping `voiceState = OctoVoiceStateIdle` (`:837`) and the whole pending-batch reset (`:838-842`). But `showVoiceOverlay` returns without creating anything when `self.view.window` is nil (`:657`), and `beginRecording` sets `voiceState = Recording` *before* calling it (`:564` vs `:580-584`). Any state other than `Idle` permanently kills both entry points — `handleHoldToTalk:` Began requires `Idle` (`:469`) and `onModeToggleTapped` returns early when not `Idle` (`:362`). There is no path back.

The reference sets the state explicitly next to the teardown rather than through it (`WKHoldToTalkManager.m:360` and `:366`: `[self hideOverlay]; self.state = WKHTTStateIdle;`). Do the same in `stopRecordingAndTranscribe`'s two early returns (`:602-615`) and in both transcription error handlers.

**P1-2 — Audio-session interruptions are not observed.**

Nothing in the new code registers for `AVAudioSessionInterruptionNotification`. An incoming call, Siri, or another app taking the mic mid-recording stops `AVAudioRecorder` while `voiceState` stays `Recording`, the timers keep running, and the overlay stays up over a recorder that is no longer capturing. The reference registers at `WKHoldToTalkManager.m:118` and routes `AVAudioSessionInterruptionTypeBegan` to a cancel path at `:143-148`. `AVAudioRecorderDelegate` is also not adopted, so `audioRecorderEncodeErrorDidOccur:` goes unnoticed too.

**P1-3 — In-flight transcription callbacks carry no generation guard.**

Neither Phase A (`:1198-1238`) nor Phase B (`:1291-1318`) checks whether the batch it belongs to is still current, and `finishThinkingAndShowText` (`:1028-1040`) unconditionally sets `voiceState = OctoVoiceStateResult`. With `kTranscribeTimeout` at 30 s there is a wide window. Today the overlay happens to block every cancel affordance during `Thinking`, which mostly masks this — but the moment P0-2 is fixed by tearing the overlay down on disappear, a late callback lands on a dismissed page and re-arms the state machine into `Result` with no overlay, i.e. exactly the dead state in P1-1. Please fix these two together: a monotonic batch token compared inside each completion closes both.

**P1-4 — The line-level diff allocates O((L_old+L_new)²) up front, uncapped, unchecked, on the main thread, per keystroke.**

`wk_lineMatchAnchorsForOldLines:` (`:1521-1583`) allocates

```objc
NSInteger maxD = n + m;
NSInteger size = 2 * maxD + 1;
int32_t *trace = (int32_t *)malloc((maxD + 1) * size * sizeof(int32_t));   // :1528
```

That is ≈ 8·(n+m)² bytes **always**, independent of the actual edit distance — the comment at `:1418-1422` reasons about the line-level edit distance being small, but `trace` is sized by `maxD`, not by the `D` actually reached. 1 000 changed lines ≈ 8 MB; 3 000 ≈ 288 MB. `kVoiceDiffMaxTotalLen` (`:1367`) caps only the *character* pass; there is no line-count cap anywhere.

Neither `malloc` return is checked (`:1527-1528`, `:1594-1595`), so exhaustion is a null deref in the `memcpy` on the next line rather than a graceful fallback.

And the whole thing runs synchronously from `textViewDidChange:` → `wk_refreshManualEditHighlight` (`:400-424`) — on every keystroke. Even fully within the character cap, a segment at `n + m = 2000` means a 32 MB allocation per typed character.

Apply a line-count threshold with the same "degrade to whole-block highlight" fallback the character path already has, check both `malloc` results, and consider debouncing `wk_refreshManualEditHighlight` rather than running it inline on each change.

**P1-5 — The 60-second cap transcribes even when the user is holding in the cancel zone.**

`stopRecordingAndTranscribe` accepts `Cancelling` as well as `Recording` (`:596`), and `onRecordTick` calls it unconditionally at the cap (`:587-592`). So a user who has slid their finger up past the 60 pt threshold — overlay reading "松手 取消", bubble red — has the recording uploaded, rewritten and written into the document anyway, against the intent the UI is actively confirming. `Cancelling` at the cap should route to `cancelRecordingIfNeeded`.

### 🟡 P2

- **The undo stack is never cleared, and may retain the controller.** `wk_replaceTextViewAttributedText:` (`:1349-1355`) registers via `[[undoManager prepareWithInvocationTarget:self] ...]` against `_wk_privateUndoManager`, which the controller owns strongly (`:24`, `:318-323`). Two problems, one certain and one to confirm. Certain: every whole-document voice edit pushes a full `NSAttributedString` copy of the document onto the stack, and nothing ever calls `removeAllActions` — an editing session with several voice rounds on a long summary holds every intermediate copy until teardown. To confirm: `NSUndoManager` calls `retainArguments` on the invocation, which retains the invocation's *target* as well as its arguments; if that applies here, controller → undo manager → invocation → controller is a cycle and the page never deallocates at all, which would also mean the timers in P0-2 never stop. Register against a non-owning target, or clear the stack in `viewWillDisappear:` rather than relying on `dealloc` (which a cycle would prevent from running).
- **Recording-startup failure leaks the audio session category.** `:557-562` calls `setActive:NO` directly instead of `restoreAudioSession` (`:637-643`), so the app is left on `PlayAndRecord` + `DefaultToSpeaker` with `previousAudioCategory` still latched. It also skips `cleanupRecordFile`, leaving a stray `.m4a` in `NSTemporaryDirectory()`.
- **`onSave` destroys the highlight before the request and never restores it on failure.** `:336-357` runs `clearVoiceEditHighlight` first; on a 409 or a network error the user stays on the page with every diff marker gone. The programmatic `attributedText` assignment at `:1677` is also not registered with the undo manager, unlike `wk_replaceTextViewAttributedText:` (`:1349-1355`). Clear after a successful response instead.
- **Attribute dictionaries are built from nullable properties.** `:1661-1662` and `:1677-1679` put `self.textView.font` / `self.textView.textColor` straight into `@{}` literals. Both are `nullable` and `UITextView` reports nil for them once `attributedText` has been assigned an empty attributed string — a nil value in a dictionary literal is an immediate `NSInvalidArgumentException`. Cache them in `viewDidLoad` or add `?:` fallbacks.
- **`insertText:` on a possibly-unfocused text view.** `wk_insertDictatedSuffix:atCursorRange:` (`:1329-1336`). The comment at `:1323-1328` is right that this path buys automatic undo registration, `selectedRange` advance and a `-textViewDidChange:` callback — but those are properties of an active text input session. If the user starts a voice edit without ever tapping the body, `self.textView` is not the first responder. I could not verify the runtime behaviour statically; please either call `becomeFirstResponder` first or confirm this was tested from a cold, unfocused page.
- **Broken doc reference.** `:1261` cites `docs/语音指令编辑正文接口需求文档.md`; the repo only contains `docs/语音长按录制设计文档.md`. The "pure dictation vs. edit instruction" inference at `:1274` is the single most load-bearing heuristic in this change, and its justification points at a file that does not exist.
- **Divergent cancel thresholds.** 60 pt in `handleHoldToTalk:` (`:482`, measured in `self.view`) vs 100 pt in `handleAppendLongPress:` (`:1072`, measured in `self.voiceOverlay`). Same gesture, different feel.
- **Ordering.** `wk_updateUndoButtonState` is called at `:148`, before `self.textView` is created at `:150`. It works only because messaging nil yields `NO`.
- **Release-build logging.** `:1209` and `:1303` are not `#if DEBUG`-gated, unlike the request/response logging in `WKVoiceInputService.m:252-260` and `:278-286`. Only lengths and error metadata are logged, so nothing leaks — but the file being mirrored is explicit about the convention.

---

## 3. Verdict

**CHANGES_REQUESTED** — three P0 recording-lifecycle defects (recording after finger-up, orphaned window overlay, append clips discarded at the 60 s cap), plus five P1s. The scope section is also unmet: the `mode:` plumbing the description advertises is never exercised.

## 4. What to fix

1. Guard the permission callback on gesture-still-active in both `requestRecordPermissionThenStart` call sites — `WKHoldToTalkManager.m:272-277` is the pattern to copy.
2. Tear down overlay + timers + recorder + audio session in `viewWillDisappear:` and `dealloc`, unconditionally.
3. Branch `onRecordTick` on `isAppendMode`, and route the cap to cancellation when the state is `Cancelling`.
4. Set `voiceState` explicitly on every terminal path rather than relying on `hideVoiceOverlay`, and add a batch token that in-flight completions check before touching UI or state.
5. Observe `AVAudioSessionInterruptionNotification` and adopt `AVAudioRecorderDelegate`.
6. Cap the line-level diff, check both `malloc` results, and stop running the full diff inline on every keystroke.
7. Clear the undo stack on disappear (and confirm the invocation target is not retained).
8. Either use `mode:` (`append_only` for Phase A) or drop the parameter; remove `isASRPreview` / `recordedDuration` / `recordedLevels` or wire them up.

## 5. Coverage note

I read the diff and both in-repo reference implementations; I did not build or run the app, so all runtime claims are static. Not covered, and worth a look before merge:

- **Server side.** `append_only` / `edit_only` are documented at `WKVoiceInputService.h:64-65` but nothing sends them — I could not confirm the backend accepts the field. Verify before the parameter ships.
- **Tests.** This repository has no test target, so nothing here is covered automatically. `wk_changedRangesInNewText:comparedToOldText:` is a pure string→ranges function and would be cheap to pin down if one is ever added; the P0-3 append/60 s boundary and the delayed-permission path are the other two I would want locked in.
- **Not evaluated:** iPad and split-screen layout of the window-level overlay, VoiceOver, Dynamic Type, backgrounding mid-recording, and surrogate-pair (emoji) boundaries — `wk_changedRangesInNewText:` works in raw UTF-16 units (`:1380-1383`), so a highlight range can land mid-surrogate.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #82 (octo-ios)

## Summary

本 PR 为总结编辑页(`OctoSummaryEditVC`,+1551 行)接入完整的语音编辑能力:长按录音 → ASR 预览("确认原话")→ 支持"按住继续"多段链式录音 → Phase B 携带正文全文上下文改写并写回,辅以两级 Myers diff 的紫色高亮与撤销/重做集成;另对 `WKVoiceInputService` 增加 `mode:` 参数重载(m4a/wav 双路径),旧 selector 转发 `mode:nil`。整体实现质量较高:手势状态机、定时器清理、隐私日志收敛、插入路径复用 UITextView 原生打字链路等细节都处理到位。审查未发现 P0/P1 阻塞项,5 条 P2 建议集中在失败路径收敛与防御性清理。

仅静态分析,基于 head `ebb00bf099c56a0f7f3cbf30ee0f023873630fc6`;本环境无法执行 iOS 构建与测试,未运行。

## Verification

- ✅ **服务向后兼容** — 旧 `transcribeAudio:...completion:` / `transcribeWavAudio:...completion:` 均转发至新重载并传 `mode:nil`(`Modules/WuKongBase/WuKongBase/Classes/Services/Voice/WKVoiceInputService.m:211`、`:318`);`mode` 仅在非空时进入 `formFields`(`:247-248`、`:343`),既有调用方线格式不变。现有调用点 `WKHoldToTalkManager.m:475`、`WKVoiceInputView.m:671` 不受影响。
- ✅ **m4a/wav 双路径 parity** — 两个 `mode:` 重载逐行一致(成对路径核查通过)。
- ✅ **超时保护** — `kTranscribeTimeout=30s` 覆盖新 VC 的转写调用路径,thinking 态不会永久卡死。
- ✅ **手势状态机** — 首次长按与"按住继续"长按均正确处理 `Cancelling` 态(上滑取消、回落恢复,`OctoSummaryEditVC.m:1066-1075` 注释与实现一致);60s 录音上限、100pt 上滑阈值。
- ✅ **撤销集成** — 口述插入走 `insertText:`,与手动打字共用 UITextView undo 栈;`NSUndoManager` 对 target 持 unowned 引用(官方文档),无 retain cycle。
- ✅ **高亮不伤编辑** — 只重贴 `NSBackgroundColorAttributeName`,不改字符内容;输入法组词期有 `markedTextRange` 守卫(`OctoSummaryEditVC.m:413`)。
- ✅ **资源清理** — `dealloc` 移除全部通知观察者(`:213-215`);`invalidateRecordTimers` 在所有停止/取消路径调用;录音临时文件在短录音、空音频、取消三个出口均清理。

## Findings

无 P0/P1;以下 5 条 P2,另有 1 条 nit 见"额外发现"。

### P2 — beginRecording 失败路径不还原 audio session category (`Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryEditVC.m:560`)

录音启动失败分支(`:559-563`)只做了 `setActive:NO` 与状态复位,未调用 `restoreAudioSession`:`previousAudioCategory` 保持非空,且 session 停留在 `:539` 设置的 `PlayAndRecord`。下一次 `beginRecording` 会把当前(已被改成 PlayAndRecord 的)category 当作"原 category"保存(`:537-538`),之后成功录音结束时 `restoreAudioSession` 还原的是这个错误值,app 级音频路由可能长期停留在 PlayAndRecord,影响语音消息等正常播放的路由/音量,直到重启。Diff-scope:new(本 PR 新增代码)。触发条件罕见(录音初始化失败)、不产生错误数据、重启可恢复,不阻塞。修复方向:失败分支补调 `[self restoreAudioSession]`(先还原 category 再 `setActive:NO`)。

### P2 — viewWillDisappear 只清 Recording 态,Thinking/Result 态被移出栈时 window 级 overlay 泄漏 (`Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryEditVC.m:219`)

`viewWillDisappear`(`:217-220`)只调用 `cancelRecordingIfNeeded`,而后者在 `voiceState` 非 Recording/Cancelling 时直接 return(`:620-621`),`hideVoiceOverlay` 不会执行。该 VC 被 push 到根导航栈(`OctoSummaryDetailVC.m:1122`),overlay 挂在 window 上且 `userInteractionEnabled=YES`:若在 ASR 等待(≤30s)或结果确认(无限等待用户)窗口内有任何栈级操作把本 VC 移出(根栈上存在多处 `popToRootViewControllerAnimated:`/`setViewControllers:` 调用方,如 `WKUserInfoVC.m:723`、`WKGlobalSearchV2VC.m:398`),全屏遮罩会永久残留并挡住整个 app UI,只能杀进程恢复;`dealloc`(`:213-215`)也未移除。当前代码中我没有找到在 overlay 展示期间可触达上述 pop 路径的具体入口(用户侧返回/侧滑均被 overlay 挡住),故定 P2 而非 P1,但这是典型防御缺口。Diff-scope:new。修复方向:`viewWillDisappear` 无条件补一句 `[self hideVoiceOverlay]`。

### P2 — Phase B 无声启发式会把整条链式录音批次丢弃 (`Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryEditVC.m:1312`)

`[result.text isEqualToString:baseContext]` 用于识别服务端 IsNoSpeech 原样回显(`:1310-1311` 注释),但合法的零改动指令(改写结果恰好等于输入)也会命中:展示误导性的"未识别到语音"并直接 return,`remainingClips` 里后续已录好的片段被整批丢弃,用户需要重录。对比 Phase A 对追加空片段的处理是移除当前片段后继续(`:1215-1229`),两端不对称。Diff-scope:new。不产生错误数据(正文保持原样)、可重录恢复、触发罕见,不阻塞。修复方向:命中该分支且 `remainingClips` 非空时跳过当前片段继续链式处理,仅在整批都无有效结果时才提示并关闭。

### P2 — 每次键入全量重跑两级 Myers diff,贴近长度上限时开销可观 (`Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryEditVC.m:412`)

`textViewDidChange`(`:400-405`)每次键入都执行 `wk_refreshManualEditHighlight`,对全文 vs `initialContent` 重跑行级 Myers + 片段字符级 Myers。设计上有 `kVoiceDiffMaxTotalLen=2000` 的片段级兜底(超限退化为整段高亮,`:1505`),但贴近上限的单个片段仍需约 (maxD+1)×(2maxD+1)×4B ≈ 32MB 的瞬时 `trace` 分配与 O(D²) 时间,旧机型连续输入可能出现可感知卡顿。Diff-scope:new;不构成正确性问题。修复方向:对高亮刷新做 debounce/合并到下一 runloop,或按文本内容缓存上次 diff 结果。

### P2 — 新录音路径未监听 AVAudioSessionInterruptionNotification,与兄弟组件不对齐 (`Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryEditVC.m:535`)

`WKHoldToTalkManager.m:118` 与 `WKVoiceInputView.m:98` 都注册了中断通知并在中断开始时收敛状态(如 `onAudioInterrupt:`,`WKHoldToTalkManager.m:143`);本 VC 的录音全程没有注册。录音中来电/Siri/闹钟会使系统直接停掉 `AVAudioRecorder`,而 VC 状态仍停留在 Recording:`recordTimer` 继续走时(最长 60s),松手后拿到被截断的音频文件并照常进入转写,用户可能在不知情下确认一段不完整的文本。Diff-scope:new(成对路径 parity 缺口)。触发罕见、可重试恢复、转写内容即实际录到的音频,不阻塞。修复方向:参照 `onAudioInterrupt:` 在中断开始时停止计时并按取消路径收敛,或至少标记本段已被系统截断。

## 数据流回溯

对 diff 中被消费的关键数据逐条回溯到来源,确认运行时真实可达:

- `result.text`(Phase A/B completion)← `WKVoiceInputService` 转写回调 ← 服务端响应;消费点 `runEditPhaseWithBaseContext:` 与无声启发式(`:1312`)均已核对,含 IsNoSpeech 原样回显分支。
- `pendingAudioClips` ← 录音停止路径 append ← `confirmASRPreviewAndRunEditPhase` 消费(有 `count==0` 守卫);追加空片在 Phase A 被 `removeLastObject` 后继续,数据真实流入链式处理。
- `previousAudioCategory` ← `beginRecording` 捕获 `session.category` ← `restoreAudioSession` 消费;失败路径缺口见 P2-1。
- `initialContent`(diff 基线)← VC 初始化时载入的原文 ← `wk_changedRangesInNewText:` 消费,高亮只作用于展示层。
- `transcribedText` ← Phase A 结果 ← `finishThinkingAndShowText` 展示;明确只读、不带入 Phase B(Phase B 重新上传音频,`:1195-1197` 注释与实现一致)。
- `audioData` ← `NSData dataWithContentsOfFile:recordFilePath` ← 长度校验后进转写;空文件有独立失败出口(`:611-615`)。
- `mode` 参数 ← 新重载入参 ← `formFields[@"mode"]`(仅非空),旧调用方经 `mode:nil` 转发,线格式不变。

## 盲点 checklist (C1–C6)

- **C1 双路径 parity**:命中并核查。(a) m4a/wav 两个 `mode:` 重载逐行一致 ✅;(b) 旧 selector 转发 `mode:nil` 两端一致 ✅;(c) 音频中断处理:兄弟组件有、新 VC 无 → 见 P2-5;(d) Phase A/B 空片段处理不对称 → 见 P2-3。
- **C2 control-flow ordering / 嵌套复用**:`cancelRecordingIfNeeded` 被 `viewWillDisappear` 与手势取消复用,其早退条件在 disappear 场景下产生顺序依赖 → 见 P2-2;`transcribe` 新重载为薄封装,无 call-site 重排风险。
- **C3 授权边界**:N/A — 无权限/凭证/tool 暴露变更,语音功能为本地 UI + 既有转写接口。
- **C4 授权生命周期 / 容器状态级联**:N/A — 无鉴权改动。
- **C5 build 通过 ≠ 运行期正确**:本环境无法构建/运行 iOS 工程,已按静态推演替代(见数据流回溯);不以此作为任何运行时结论的缓解证据。
- **C6 治理/文档自洽**:无治理/安全策略变更;仅发现一处失效文档引用(见额外发现)。

## 跨轮 blocker 复检

首轮审查,无上一轮 blocker,N/A。

## 额外发现

- Nit — 注释引用了不存在的文档:`Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryEditVC.m:1261` 提到 `docs/语音指令编辑正文接口需求文档.md` 8.4 节,仓库 `docs/` 下无此文件(现有 `语音长按录制设计文档.md` 等)。建议补交文档或修正引用。

## Verdict: APPROVED

未发现 P0/P1 阻塞项:服务层改动为纯增量重载且向后兼容,新 VC 的主链路(录音→ASR→链式改写→写回+高亮)经逐条数据流回溯确认运行时可达,手势/定时器/撤销/清理等状态机完整。5 条 P2 均为失败路径收敛与防御性建议(音频会话还原、overlay 清理、链式批次容错、键入性能、中断对齐),不构成合并阻塞,建议作者按优先级择机处理。

yujiawei:
Follow-up to my review above — one additional P2 that a second pass over the Phase B chain turned up. Same head (`ebb00bf0`); the verdict is unchanged.

**P2 — the "no speech" heuristic in Phase B false-positives on a zero-change rewrite, and aborts the whole remaining clip chain.**

`OctoSummaryEditVC.m:1312`:

```objc
if (result.text.length == 0 || (baseContext.length > 0 && [result.text isEqualToString:baseContext])) {
    [weakSelf.view showMsg:LLang(@"未识别到语音")];
    [weakSelf hideVoiceOverlay];
    return;
}
```

The comment at `:1310-1311` explains the intent: the server echoes `context_text` back verbatim on its `IsNoSpeech` branch, so an unchanged result means nothing was heard. But "output equals input" is also what a legitimately no-op instruction produces — the model declining to change anything, or the user saying something that maps to "leave it as is". Those get reported to the user as "未识别到语音", which is wrong and misleading.

The bigger problem is what it does to the batch. This branch returns without recursing, so **every clip still in `remainingClips` is dropped** — record three segments via "按住继续", have the first come back unchanged, and segments two and three are discarded without ever being uploaded. The user is told nothing was heard and has to re-record all of it.

Phase A already handles the analogous case correctly and asymmetrically: on an empty append result it removes just that clip (`:1225`) and stays on the confirmation bubble so the rest of the batch survives. Phase B should do the same — skip the current clip, continue the chain, and only surface the message and close if the entire batch produced no change.

PR #90, opened 2026-09-03
feat(summary): 群聊总结完成后往来源群发提示, 收敛到详情页触发

swr614:
## What

群聊智能总结完成后，往来源群发一条 WK_TIP(2000) 系统提示「XX 总结了群聊内容」。

## Why

1. iOS 端完全没有这条提示 —— web / 安卓已上线，iOS 侧连 `creator_id` 都没解析。
2. 时机不能是「后台轮询到完成就发」：用户自己还没看过总结，群里就先冒出提示；而且列表页 `hydratePreview` 会把历史已完成的自建任务批量追溯广播。
3. 进详情页有三个入口（群聊发起 / 通知助手 / 上下文→智能总结列表），必须共享同一次去重，同一条总结在群里只能出现一次。
4. 通知助手的「查看详情」落在内置 WebView：WebView 里的 web 端没有独立 WKSDK 连接、也没有本机发起标记，自己发不出提示 → 需要 native 侧在打开 WebView 之前补一次判定。
5. 发起总结的这台手机自己看不到提示（web 能看到）—— `sendMessage:channel:` 只落库 + 发网络包，不通知当前打开的聊天页 UI。

## How

1. 新增 `OctoSummaryTipContent`（WK_TIP 2000，文案模板与 web 锁定一致不做本地化）；模型补 `creator_id` 解析（`creator_id` 优先，其次 `user_id`）。
2. **发送点只有两个**：详情页轮询观测 `Pending/WaitingConfirm/Processing → Completed` 跃变；通知助手卡片按钮 / 文本链接 / 链接卡片点击时在打开 WebView **之前**识别 `/s/<taskNo>` 深链并异步判定（纯副作用，不拦截不改导航）。列表页不参与发送。
3. 判定闸门：`status == Completed` && `creator_id == 本机 uid` && (观测到跃变 ‖ 消费掉本机 eligible 标记) && `(taskId, channelId)` 未发过 && 目标群非空 && 显示名非空。
4. 新增 `OctoSummaryNotifyStore`：`SENT` 按 `taskId → [channelId]` claim-before-send + 失败回滚 + FIFO 500；`ELIGIBLE` 记本机发起（创建 + 两处重新生成）的 taskId，10 分钟 TTL + 一次性消费，覆盖「创建后极快完成、首屏就是终态」的边界，同时挡住历史任务追溯广播。
5. 创建成功后 replace-push 详情页（原先只 pop 回去，轮询不跑提示永远发不出去），返回键自然回到发起前页面。
6. 详情页轮询 8s → 3s，且轮询失败不再永久中断。
7. 发送后 `header.showUnread = NO` + 补一次 `onRecvMessages` 委托，让本机也能看到且不被自己的提示打扰。
8. 新增 `GET /summaries/{task_no}`；跨 module 走 `WKPOINT_SUMMARY_DEEPLINK`，handler 注册在 `OctoContextModule`，WuKongBase 侧只 invoke。

## Checklist

- [x] Code compiles with no new warnings（BUILD SUCCEEDED，0 error，改动文件 0 新增 warning）
- [ ] Linter and formatter pass（仓库未配置 Obj-C linter/formatter）
- [ ] Tests added/updated for the changes（仓库无 OctoContext 测试 target）
- [ ] All tests pass locally（同上）
- [x] Rebased onto latest main

## Testing

- `xcodebuild -workspace OctoiOS.xcworkspace -scheme OctoiOS -configuration Debug -sdk iphonesimulator build` → **BUILD SUCCEEDED**。

## Breaking Changes

N/A


Jerry-Xin (changes requested):
Review at head 0e618be4 (merge-base 5b948cbe). Feature: after a group-chat summary completes, the initiating client sends one WK_TIP(2000) system tip into the source group(s), with triggers converged to detail-page polling observation plus notification-assistant deep links. Byte-level review against the checked-out head; key SDK paths verified at the source level.

## 🔴 Blocking

- 🔴 **The outgoing packet still carries `showUnread = YES`, defeating the stated no-unread/no-remind intent for every receiver.** `OctoSummaryGroupNotifyHelper.m` (`notifyIfNeededWithDetail:`) calls `[[WKSDK shared].chatManager sendMessage:tip channel:ch]` first and only afterwards sets `message.header.showUnread = NO`. But `sendMessage:channel:` synchronously runs `WKChatManager sendMessage:addRetryQueue:` → `WKMessageQueueManager.m:52` `sendPacket.header.showUnread = message.header.showUnread`, and `WKCoder.m:29` packs that flag into the wire header. So the transmitted message (and the DB copy persisted by `saveMessage:` before the mutation) has `showUnread = YES`. Impact is real and verified on the receive side: unread counting is gated by `header.showUnread && !isSend` (`WKChatManager.m:758/764/768`), remind sound/vibration by `header.showUnread` (`WKSystemMessageHandler.m:136`, no `isSend` exclusion), and local notifications by the same flag (`WKLocalNotificationManager.m:35`) — i.e. every group member's device gets unread +1 and a notification sound for this "lightweight tip", and the sender's own other devices get reminded too. The post-hoc mutation only fixes the in-memory object fed to `callRecvMessagesDelegate:` (local echo). Fix by making the flag NO before enqueue — e.g. build the `WKMessage` (public `sendMessage:(WKMessage*)` overload) or add a send API that accepts the header; compare `WKTypingManager.m:195`, which sets `header.showUnread = false` before sending.

- 🔴 **`WaitingConfirm` permanently stops detail-page polling, so completion of a confirmation-gated task is never observed and the tip is never sent.** `OctoSummaryDetailVC.m loadDetail` schedules the next poll only for `OctoTaskStatusProcessing`/`OctoTaskStatusPending`. When a poll response arrives with `OctoTaskStatusWaitingConfirm`, polling dies; the later `WaitingConfirm → Processing → Completed` progression is never seen by the VC the user is sitting on, and the transition path never fires. This contradicts the PR's own documented contract — `OctoSummaryGroupNotifyHelper.h` states the trigger is "观测到 Pending/WaitingConfirm/Processing → Completed 的状态跃变" — and diverges from the list poller, which deliberately tracks `WaitingConfirm` (`OctoSummaryListVC refreshPoller`). Re-entering the page within the 10-minute eligible window can recover via the first-screen path, but the primary designed flow (create from chat → auto detail page → stay → completion) silently loses the tip for any confirmation-gated summary. Include `OctoTaskStatusWaitingConfirm` in the `scheduleNextPoll` condition.

## 💬 Non-blocking

- 🟡 `OctoSummaryAPI.getSummaryDetailByNo:` percent-encodes with `URLPathAllowedCharacterSet`, which keeps `/` and `.` — a clicked link whose `task_no` query value contains `../` sequences injects dot-segments into the same-origin summary-API URL (`request:path:` builds `/summary/api/v1/summaries/<encoded>`). Gated by `hasAnyEligibleTask` (10-min window), GET-only, and the response is never surfaced (transform yields a model or nil), so impact is bounded — but whitelist `[A-Za-z0-9_-]` to match the `/s/<seg>` path grammar and close it entirely.
- 🟡 The dedup claim is two separate locked operations (`hasSentTaskId:` then `markSentTaskId:`). Today this is safe in practice because every caller (detail-page poll callbacks, deep-link judgement callbacks on AFNetworking's default main completion queue, and cell-tap endpoint invokes) lands on the main thread, so the pair is serialized — but that invariant is implicit. A single atomic `claimTaskId:channelId:` returning whether the claim succeeded would make it robust against any future background caller.
- 🔵 Detail-page polling also stops in `viewWillDisappear` (including transient child-page pushes) and is not re-armed in `viewWillAppear` — miss-side only and pre-existing lifecycle, but a completion during that window won't notify via the transition path.
- 🔵 `result[@"task_id"] longLongValue` (CreateVC; same existing idiom in both regenerate sites) is not `NSNull`-guarded; a `task_id: null` success payload would crash. Low likelihood, cheap guard.
- 🔵 Dedup state is device-local (`NSUserDefaults`): two devices on the same account could in theory both observe the transition and both send — same property as the web/android implementations, noted for awareness.
- 🔵 CI: build is green on this head; the `Check Sprint` workflow fails only because the PR body has no `Closes #<issue>` reference (precedent PR #79 merged with the same failure). Adding the reference clears it.

## ✅ Highlights

- Trigger convergence verified clean: merge-base has no summary-tip sender at all, and on this head the only send entry is `notifyIfNeededWithDetail:`, reached solely from the detail-page transition and the deep-link judgement; list-page `hydratePreview`/poller do preview/status UI only. `markEligibleTaskId` appears at exactly the three documented sites (create + two regenerate paths).
- Duplicate-send analysis across transition / first-screen / deep-link / regenerate paths found no duplication route: one-shot status anchor, atomic one-time eligible consumption, claim-before-send with rollback, legacy-key compatibility for earlier gray builds, bounded FIFO eviction whose window cannot overlap the 10-min eligible TTL.
- Cross-module wiring via the `WKPOINT_SUMMARY_DEEPLINK` endpoint avoids a compile-time WuKongBase→OctoContext dependency and is a verified no-op when unregistered; creator check plus local-initiation eligible marks keep the privacy gate sound, including the account-switch case.

yujiawei (changes requested):
# Code Review — PR #90 (octo-ios)

Reviewed at head SHA `0e618be47844f1cf513dd4a72f7329ac20d589cc` against merge-base `5b948cbe`. 17 files, +753/-28. Verified against the checked-out tree, not just the diff — the findings below trace through `WuKongIMSDK` / `WuKongBase` code that is *not* in the diff, which is where the two most important problems live.

The design is well thought through: two send points instead of a background broadcaster, claim-before-send with rollback, an eligible marker to cover the "completed before the first frame" edge, and an explicitly documented "rather miss than duplicate" bias. The commit message is unusually honest about its own limitations. Most of what follows is not about the ledger logic — that part holds up — but about the surrounding plumbing.

**Verdict: CHANGES_REQUESTED.** Four P1s: two of them mean a stated goal of this PR does not actually take effect, one is a delivery gap for a whole task state, and one is an input-validation hole reachable from any chat message.

---

## 1. Specification compliance

Checking the eight items in the PR's own "How" list. Six land as described. Two do not:

**Item 7 — "let the sending device see the tip too" — does not visually work.** See P1-1. The bubble is inserted, but it renders collapsed to an empty pill.

**Item 3 / the documented gate — claims a state transition the code cannot observe.** `OctoSummaryGroupNotifyHelper.h:9-10` states the trigger covers `Pending/WaitingConfirm/Processing → Completed`. `WaitingConfirm` cannot be observed by the detail-page poller. See P1-2.

Nothing was built that wasn't asked for. The removal of the chat-entry `OctoActionToast` "view" affordance is called out in the commit message, so it is in scope — but it leaves dead code and a now-false comment behind (P2-5).

**Spec: ❌** (item 7 not achieved; documented gate covers a state the implementation cannot reach)

---

## 2. Code quality

### P1-1 — The local echo renders as an empty collapsed bubble; goal #7 is not achieved

`Modules/OctoContext/OctoContext/Classes/Network/OctoSummaryTipContent.m:15-22`

```objc
+ (instancetype)tipWithUid:(NSString *)uid name:(NSString *)name {
    OctoSummaryTipContent *c = [self new];
    c.content = @{
        @"content": @"{0}总结了群聊内容",
        @"extra": @[@{@"uid": uid ?: @"", @"name": name ?: @""}],
    };
    return c;
}
```

`displayContent` is never set. In the SDK it is populated **only** inside the decode path:

`Modules/WuKongIMiOSSDK/WuKongIMSDK/Classes/model/content/WKSystemContent.m:18-21`

```objc
- (void)decodeWithJSON:(NSDictionary *)contentDic {
     self.content = contentDic;
     self.displayContent =[self getDisplayContent];
}
```

The send path never calls `decodeWithJSON:` — `contentToMessage:` only calls `[content encode]`. So the instance handed to `callRecvMessagesDelegate:` (`OctoSummaryGroupNotifyHelper.m:119`) has `displayContent == nil`, and the cell sizes itself from exactly that field:

`Modules/WuKongBase/WuKongBase/Classes/Sections/Messages/WKSystemMessageCell.m:37-41`

```objc
NSString *text = content.displayContent;
CGSize contentSize = CGSizeMake(0.0f, 0.0f);
if(text) {
    contentSize =  [[self class] getTextSize:text maxWidth:WKScreenWidth - WK_SYSTEM_TEXT_SPACE];
}
```

`text` is nil → `contentSize` stays `(0,0)` → the cell reports `(20, 20)`. Then `layoutSubviews` (`:114-118`) sizes `tipTextBoxView` to `(contentSize.width-10, contentSize.height-10)` = `(10, 10)`, and `tipTextLbl` to the same. Meanwhile `refresh:` (`:72`) sets the *correct* text via the cell's own `getDisplayContent:` helper, which reads `content.content` directly and therefore works.

Net effect: correct text, 10×10 label. **The user who initiated the summary sees a tiny empty grey pill, not "你总结了群聊内容".** The whole point of the `callRecvMessagesDelegate:` call is defeated. Note this only affects the in-memory echo — after an app restart the message is re-read from the DB, decoded through `WKSystemContent`, and renders correctly. That asymmetry makes it easy to miss in manual testing if you relaunch between attempts.

Fix: populate `displayContent` at construction, e.g. `[c decodeWithJSON:c.content];` after assigning `content` (this also gives you the `{0}` → `你` substitution consistent with what the cell computes), or set `c.displayContent` explicitly.

### P1-2 — `WaitingConfirm → Completed` can never be observed, so confirm-gated summaries never emit the tip

`Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryDetailVC.m:528-531`

```objc
if (weakSelf.detail.status == OctoTaskStatusProcessing
    || weakSelf.detail.status == OctoTaskStatusPending) {
    [weakSelf scheduleNextPoll];
}
```

`OctoTaskStatusWaitingConfirm` (= 1) is not in that set, so polling stops the moment the task enters it. But `WaitingConfirm` is a real, rendered, long-lived state — `OctoSummaryDetailVC.m:584` (`BOOL waiting = (d.status == OctoTaskStatusWaitingConfirm);`) draws a dedicated waiting UI with a "查看确认状态" button, and `:1092` gives it its own action-sheet branch.

So for a summary that waits on participant confirmation:

1. Detail page reaches `WaitingConfirm`; polling stops.
2. Participants confirm → server moves the task to `Processing` → `Completed`.
3. The page never polls again → no transition is observed → **no tip**.
4. The only recovery path is re-entering the detail page, which needs the `ELIGIBLE` marker — 10-minute TTL. A confirmation-gated summary will essentially always exceed that.

The tip is therefore structurally undeliverable for this entire flow, while `OctoSummaryGroupNotifyHelper.h:9-10` documents `WaitingConfirm` as covered. Either add `OctoTaskStatusWaitingConfirm` to the reschedule condition, or correct the header comment and state the gap explicitly alongside the other known limitations.

### P1-3 — Polling dies permanently after any child page push, and the tip depends on it

`Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryDetailVC.m:247-256`

```objc
- (void)viewWillDisappear:(BOOL)animated {
    [super viewWillDisappear:animated];
    [self.pollTimer invalidate];
    self.pollTimer = nil;
    ...
    BOOL trulyLeaving = self.isMovingFromParentViewController || self.isBeingDismissed;
    if (!trulyLeaving) return;
```

The timer is killed *before* the `trulyLeaving` guard, so it also dies when the VC is merely covered by a pushed child. `viewWillAppear:` (`:215-229`) restores `contentDetaching`, re-arms the table webviews, and syncs the pop gesture — but never restarts polling. `loadDetail` is only reachable from `viewDidLoad` (`:198`), the timer itself (`:563`), and the cancel/regenerate/edit-saved callbacks. There is no pull-to-refresh on this page.

This behaviour predates the PR, but the PR makes it load-bearing: the primary send path now hangs off this poll. Concretely reachable from the detail page's own children — `OctoSummaryEditVC` (`:1174`), forward (`:1191`), and `OctoSummaryConfirmVC` (`:1212`, i.e. the "查看确认状态" button that is *only* shown in `WaitingConfirm`, the state P1-2 already breaks). Push any of them, pop back, and both the progress UI and the tip are silently dead for the rest of the page's life.

Suggest moving the `invalidate` below the `trulyLeaving` check, or restarting the poll from `viewWillAppear:` when the current status is non-terminal.

### P1-4 — Unvalidated `task_no` from an arbitrary chat link is interpolated into the API path

The path branch of the deeplink parser validates its capture; the query branch does not.

`Modules/OctoContext/OctoContext/Classes/Network/OctoSummaryGroupNotifyHelper.m:126` and `:196` (validated):

```objc
static NSString *const kSummaryPathPattern = @"^/s/([A-Za-z0-9_-]+)/?$";
...
return [self isAllDigits:seg] ? [self lookupById:seg.longLongValue] : [self lookupByNo:seg];
```

`:182-183` (not validated):

```objc
NSString *qno = [self firstQueryValueIn:comps keys:@[@"task_no", @"taskNo"]];
if (qno.length > 0) return [self lookupByNo:qno];
```

That value flows into `OctoSummaryAPI.m:166-173`:

```objc
NSString *encoded = [trimmed stringByAddingPercentEncodingWithAllowedCharacters:
                     [NSCharacterSet URLPathAllowedCharacterSet]];
...
[self request:@"GET" path:[NSString stringWithFormat:@"/summaries/%@", encoded]
```

`URLPathAllowedCharacterSet` deliberately permits `/`, `.`, `:`, `@` and more — it is the set for a whole path, not one segment. So `?task_no=a/b/c` passes through unescaped and produces `GET /summary/api/v1/summaries/a/b/c`, i.e. a different endpoint than intended; `..` segments push it further (exact resolution depends on `NSURL`/`URLSession` dot-segment normalization — worth confirming, but the unvalidated interpolation is the defect regardless).

Reachability: `handleSummaryDeepLink:` is invoked for **every** link tap in **every** message, from any sender — `WKTextMessageCell.m:2437` (`didLinkClick:`, any text containing "."), `WKTextMessageCell.m:1912` (link card), `WKInteractiveCardCell.m:660` (any adaptive-card openUrl). There is no scheme or host allowlist. The only gate is `hasAnyEligibleTask`, i.e. the victim created or regenerated a summary in the last 10 minutes.

The response is discarded, so this is a blind self-directed GET — but it composes badly with the shared error handler at `OctoSummaryAPI.m:96-98`:

```objc
if (http.statusCode == 401) {
    [[WKApp shared] logout];
}
```

If the injected path returns 401, the victim is logged out. That turns a chat link into a remote forced-logout during the eligibility window. Narrow, but cheap to close: apply the same `^[A-Za-z0-9_-]+$` validation the path branch already uses (and consider a host allowlist — see P2-4).

---

### P2 findings

**P2-1 — Unbounded 3s retry on permanent errors.** `OctoSummaryDetailVC.m:517-519` reschedules on *any* error without inspecting it. A deleted task (404) or a permission change (403) now produces one toast followed by a silent request every 3 seconds for as long as the page is open — where the previous code stopped. Worth distinguishing transient (timeout / connection lost / 5xx) from permanent 4xx, and capping consecutive retries. Related: because the error branch always reschedules, a request still in flight when `viewWillDisappear` fires will re-arm the timer on a hidden page.

**P2-2 — `showUnread = NO` is applied after the DB write.** `OctoSummaryGroupNotifyHelper.m:109` sets it after `sendMessage:` has already run `saveMessage:` → `saveMessages:` synchronously (`WKChatManager.m:85`, `:204`), and `contentToMessage:` hardcodes `header.showUnread = true` (`WKChatManager.m:110`). The in-memory mutation does achieve the stated goal (the `WKSystemMessageHandler.onRecvMessages:` remind branch reads the live object, so no sound — that reasoning checks out), but the persisted row keeps `showUnread = 1`. Inconsistent state that will confuse the next person to touch this.

**P2-3 — `default: → WK_GROUP` is too permissive in the origin-channel fallback.** `OctoSummaryGroupNotifyHelper.m:43-47` switches on `(OctoSourceType)detail.originChannelType`, where the server domain is `1=group / 2=thread / 3=DM` (see `WKConversationVC.m:456-462`). A missing or unknown `origin_channel_type` decodes as `0` and falls into `default:`, so it is treated as a group and a tip is addressed to `origin_channel_id` as a group channel. Given the stated intent is group-chat-only, prefer an explicit `case OctoSourceGroupChat:` and drop everything else.

**P2-4 — No scheme/host validation on the deeplink.** Beyond P1-4: `?task_id=<n>` on *any* host is accepted (`:177-181`), so an unrelated analytics or tracking link can trigger a detail fetch and the group-send side effect during the eligibility window. Since the shape being matched is a known web route, checking the host against the configured web origin would cost little.

**P2-5 — Dead code and a now-false comment.** With the `useActionToast` branch removed from `OctoSummaryCreateVC.m`, `OctoActionToast` has no remaining caller anywhere in the tree, and `WKConversationVC.m:463-464` still reads "提交后给一条带 '查看' 动作的 toast (OctoActionToast), 点击直接 push 智能总结列表" — which no longer happens. Update the comment; consider removing the class.

**P2-6 — `return` where `continue` is meant.** `OctoSummaryNotifyStore.m:100`: `if (![old isKindOfClass:NSArray.class]) return;` aborts the whole rollback on one malformed entry instead of skipping it. Low impact since the store only ever reads what it wrote, but the surrounding code is otherwise defensive about exactly this.

**P2-7 — The transition path does not require the eligible marker.** `OctoSummaryDetailVC.m:546-551`: on the `transitioned` branch, `consumeEligibleTaskId:`'s return value is discarded. Since `SENT` is device-local, two devices on the same account both sitting on the detail page through the transition will each send a tip. Low likelihood, and it is the deliberate main path, but it is a hole in "exactly once" worth a comment at minimum.

**P2-8 — `sendMessage:` returning non-nil only means locally persisted.** `OctoSummaryGroupNotifyHelper.m:98-102` rolls back only when the return is nil. An async SENDACK failure leaves the claim in place and the tip is permanently lost. The commit message already lists this as a known limitation, so this is a note rather than a request — but hooking rollback to the send-status delegate would make the retry story real.

**P2-9 — No tests.** Acknowledged in the checklist (no test target in this module). `OctoSummaryNotifyStore` and `lookupFromURLString:` are pure, side-effect-light logic with genuinely tricky edge cases (TTL, FIFO eviction, legacy-key compatibility, URL shapes). If a target is ever added, these two are the highest-value candidates — and P1-4 is exactly the kind of bug a parser test table catches.

**Quality: Changes-Requested**

---

## 3. Overall verdict

**CHANGES_REQUESTED** — `Spec: ❌` and `Quality: Changes-Requested`.

Blocking: P1-1 (stated goal #7 does not visually work), P1-2 (`WaitingConfirm` flow can never emit the tip, contradicting the documented gate), P1-3 (poll dies after any child push, and the tip now depends on that poll), P1-4 (unvalidated path interpolation reachable from any chat link, composing into a forced logout).

P1-1 and P1-4 are small, local fixes. P1-2 and P1-3 are each a few lines but need a decision about intended behaviour.

## 4. Suggested direction

1. `OctoSummaryTipContent.tipWithUid:name:` — call `[c decodeWithJSON:c.content]` (or set `displayContent`) so the locally echoed bubble sizes correctly. Verify by sending without relaunching the app.
2. Add `OctoTaskStatusWaitingConfirm` to the reschedule condition at `OctoSummaryDetailVC.m:528`, or amend `OctoSummaryGroupNotifyHelper.h:9-10` and the known-limitations list to say the confirm-gated flow is not covered.
3. Move `[self.pollTimer invalidate]` below the `trulyLeaving` guard in `viewWillDisappear:`, or restart polling from `viewWillAppear:` for non-terminal statuses.
4. Validate the query-derived `task_no` with the same `^[A-Za-z0-9_-]+$` rule the path branch uses, and encode it as a single path segment (`URLPathAllowedCharacterSet` minus `/`). A host allowlist on the deeplink would close P2-4 at the same time.
5. Bound the error-path retry (permanent-vs-transient, or a consecutive-failure cap) before shipping the 8s → 3s change.

## 5. Additional observations

Outside the scope of this PR, noticed while tracing the paths above:

- `WKSystemMessageHandler.m:91` declares `bool needRemind` at **file scope**, not inside `onRecvMessages:`. It is set on one message and read on the `left == 0` message, so it leaks across unrelated deliveries and is never reset on the false path. Pre-existing, unrelated to this change, but it sits directly on the path this PR now invokes deliberately — worth a separate fix.
- The tip is emitted once per group source, so a multi-group summary posts into each group independently. Correctly documented as intended; flagging only because it is the kind of thing that reads as a bug in a bug report.
- The `ELIGIBLE` marker is keyed on numeric `taskId`, but the `task_no` deeplink path can only check `hasAnyEligibleTask` before spending a request. That is a reasonable trade-off as written — just note that the fetch-then-check ordering means the request happens before the per-task decision, which is what makes P1-4 reachable at all.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #90 (octo-ios)
Repo: Mininglamp-OSS/octo-ios · Head: `0e618be47844f1cf513dd4a72f7329ac20d589cc` · Base: merge-base `5b948cbe` of `main` · automated review (automated review) for Opus 终审

## Summary

群聊总结完成后，由**发起方客户端**本地往来源群补发一条 `WK_TIP(2000)` 系统提示（"{0}总结了群聊内容"），并把触发点收敛到两处：详情页轮询观测到 非完成→完成 状态跃变，以及通知助手卡片/文本深链点击时的异步判定。新增 `OctoSummaryGroupNotifyHelper`（判定+发送）、`OctoSummaryNotifyStore`（eligible/sent 本地落账，10min TTL 一次性消费）、`OctoSummaryTipContent`（对齐 web tip.ts 的 content 形状），创建成功后改为 replace 进详情页以保证轮询跑起来；去重策略声明为"宁可漏发不重发"。

设计方向清晰、闸门（status/creator/去重/目标群非空/显示名兜底）考虑得比较全，两条触发链路的竞态用 claim-before-send 处理。但发送侧有两个阻塞级问题：线上消息头的 `showUnread` 实际仍为 true（与代码自己声明的意图相反），以及同账号多端会重复发送。另有 3 个 P2。

## Verification

Static analysis only at head `0e618be4`; build and tests not executed in this environment (iOS/Xcode 不在本环境)。以下均基于本仓源码交叉验证：

- ✅ **符号与依赖存在** — `WKPOINT_SUMMARY_DEEPLINK`(`Modules/WuKongBase/WuKongBase/Classes/WKConstant.h:258`)、`WKApp invoke:/setMethod:`(未注册即 no-op, `WKApp.m:911`)、`WKNavigationManager replacePushViewController:`(`WKNavigationManager.m:74`)、`WKChatManager callRecvMessagesDelegate:`(`WKChatManager.m:1621`)、`WKLoginInfo.displayName`(`WKLoginInfo.m:154`)、`OctoContext.podspec` 已含 `WuKongIMSDK` 依赖。
- ✅ **数据流（见下方回溯节）** — `detail.creatorId`（creator_id→user_id 兜底）、`sources/origin_channel_*` 解析、`task_id` 从 create/regenerate 原始 dict 读取，均验证到赋值点。
- ✅ **单设备内去重时序** — transition 分支先消费 eligible、deep-link 分支以消费为闸，`hasSent`+`markSent` claim-before-send，主线程串行 + 锁，双链路近同时判定不会重发。

## Findings

两条 P1（阻塞）+ 三条 P2 + 一条 Nit。

### P1 — 提示消息线上仍带 showUnread=true：全员未读+提醒音，与声明意图相反 (`Modules/OctoContext/OctoContext/Classes/Network/OctoSummaryGroupNotifyHelper.m:109`)

`message.header.showUnread = NO` 写在 `[chatManager sendMessage:channel:]` **返回之后**，但 wire 包在该调用内部就已同步快照：`contentToMessage` 默认 `showUnread=true`（`WKChatManager.m:110`）→ `sendMessage:addRetryQueue:` → `WKMessageQueueManager.m:52` 拷贝 `message.header.showUnread` 进 `WKSendPacket`。事后的修改只作用于本机 `callRecvMessagesDelegate` 回显，**发出去的消息头仍是 showUnread=true**。接收端两处都按此生效：未读计数 `makeConversationLastMessageAndUnreadCount`（`WKChatManager.m:758` 起，`showUnread && !isSend` 即 +1）；提醒 `WKSystemMessageHandler.m:136`（只判 showUnread，不排 isSend——正是本文件注释自己指出的）。结果：每发一条提示，**所有群成员未读红点+1、非免打扰且不在该群会话时播提示音**，与注释"不该冲未读红点"直接矛盾。修法：把 flag 挪到快照之前（如先 `saveMessage:` 造出消息、置 `showUnread=NO` 并落库，再走 `sendMessage:(WKMessage*)` 重载），并与安卓侧发送口径核对。

### P1 — 去重只在本机：同账号多端各自发一条，群里出现重复提示 (`Modules/OctoContext/OctoContext/Classes/Network/OctoSummaryGroupNotifyHelper.m:91`)

`hasSentTaskId:channelId:` 查的是 `NSUserDefaults`（`OctoSummaryNotifyStore`，按设备安装隔离，无账号/服务端维度）。跃变路径（`OctoSummaryDetailVC.notifyGroupsIfCompleted`）**不需要** eligible 标记，creator 校验对同一 uid 的所有设备都通过：同账号两台设备（手机+iPad；或任务从 web 创建、两台 iOS 设备都打开详情页）在任务完成窗口内各自轮询到 非完成→完成，**各发一条**，全群可见两条相同提示。头文件声明"宁可漏发不重发"，跨端无法成立。修法：让提示天然幂等——如用 `(taskId, channelId)` 派生确定性 `clientMsgNo` 交给 IM 链路/服务端去重（需确认服务端按 clientMsgNo 幂等），或改为服务端下发；若产品接受该窗口，需在文档中明示而非声明"不重发"。

### P2 — WaitingConfirm 任务永远发不出提示（轮询在该状态停摆） (`Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryDetailVC.m:528`)

`scheduleNextPoll` 只在 Processing/Pending 时续跑；任务进入 WaitingConfirm（"等待参与者确认"，确认动作发生在**其他参与者**的设备上）后观测永久停止。此类任务随后完成时本页看不到跃变；事后重开详情页走"首屏即完成"分支，而 eligible 标记 10 分钟 TTL 远短于参与者确认耗时 → **整类确认流总结永远无提示**，与 `OctoSummaryGroupNotifyHelper.h:10` 声明的触发跃变（含 WaitingConfirm→Completed）矛盾。修法：WaitingConfirm 期间低频续轮询，或从 `OctoSummaryConfirmVC` 返回/viewWillAppear 时补一次 `loadDetail`；否则改文档。

### P2 — task_no 深链未限字符集：认证请求可被打到 API 源下的任意路径 (`Modules/OctoContext/OctoContext/Classes/Network/OctoSummaryAPI.m:167`)

`getSummaryDetailByNo:` 用 `URLPathAllowedCharacterSet` 转义，**保留 `/`**；而 `task_no` 可来自任意被点击链接的 query（`lookupFromURLString` 不限 host、不限字符）。用户创建总结后 10 分钟内点一条构造链接（如 `https://任意域/?task_no=..%2F..%2Fadmin`），即触发携带本人 token + X-Space-Id 的 GET `/summary/api/v1/summaries/../../admin`。无响应回传通道，但请求伪造面应收掉：`/s/<seg>` 分支已限 `[A-Za-z0-9_-]`，task_no 应同样校验（如 `^[A-Za-z0-9_-]+$`）。

### P2 — 创建回调 replace 栈顶与用户导航竞态 (`Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryCreateVC.m:601`)

请求在途时仅禁用了提交按钮，未拦右滑返回。用户提交后立刻返回，回调到达时 `replacePushViewController:` 会把**当时栈顶的无关页面**（聊天页/列表页）换成详情页并从栈中抹掉。修法：`weakSelf.navigationController.topViewController == weakSelf`（或 `view.window`）校验后再 replace，否则走兜底。

### Nit — 错误分支 3s 定速无限重试 (`Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryDetailVC.m:519`)

持续离线/被限流时，详情页开着就每 3s 一发，无退避无上限。建议指数退避或次数上限。

- 创建/重新生成三入口（CreateVC 创建、DetailVC 重生成 `:1136`、ListVC 重生成 `:774`）eligible 标记 parity 完整；`markEligibleTaskId:0` 有 `<=0` 保护。
- claim-before-send + 失败回滚（`unmarkSent`）；legacy 键 `OctoSummaryNotifiedTaskIds` 只读兼容灰度包。
- 深链解析：`/s/<seg>` 单段正则排除 `/s/share/xxx`；裸域名补 scheme；纯数字按 id、其余按 no；`hasAnyEligibleTask` 预判避免 404 滥打。
- cell 两处钩子（卡片 `handleOpenUrlAction` / 文本 `didLinkClick`+linkCard）均为纯副作用不改导航；`WK_TIP=2000` 为既有心智常量；content 形状与 web `tip.ts` 对齐。
- 轮询生命周期：viewWillDisappear/dealloc 均 invalidate，关页不后台续跟，与声明一致。

## 数据流回溯（逐项到上游来源）

| 被消费数据 | 上游来源 | 是否真流到消费点 |
|---|---|---|
| `detail.creatorId` | `OctoSummaryModels.m:241` `creator_id`→`user_id` 兜底（两处 parse 均改） | ✅ 到达 `:69` 比对 |
| `detail.sources[].sourceType/sourceId` | `parseSources`(`OctoSummaryModels.m:202`) 直解 `sources` 数组，无提前 return 陷阱 | ✅ 到达 `resolveTargetChannelsForDetail` |
| `originChannelId/originChannelType` | 既有 parse（本 PR 未改，仅新增消费） | ✅ 仅 WK_GROUP 走 fallback |
| `task_id`（创建成功） | `createSummaryWithParams` transform=nil → 原始 envelope data dict | ✅ 与 regenerate 同口径（`DetailVC.m:1127` 注释互证） |
| `selfUid` | `[WKSDK shared].options.connectInfo.uid` | ✅ 空则不发 |
| `displayName` | `WKLoginInfo.m:154` 实名→昵称 | ✅ 三级兜底到 uid |
| wire `showUnread` | `WKChatManager.m:110` 置 true → `WKMessageQueueManager.m:52` 同步快照 | ❌ 消费点意图(=NO)未流到线上（P1-1） |

## 盲点 checklist（C1–C6）

- **C1 双路径 parity** — clear：create↔regenerate（×3 入口）eligible 对齐；mark↔consume/unmark 成对存在；深链三调用点同口径。
- **C2 control-flow/复用** — clear（含一处残留风险已入 Nit）：`notifyIfNeededWithDetail` 双链路复用由 hasSent 幂等；`lookupFromURLString` 三入口复用行为一致。
- **C3 授权边界** — hit→P2-2：新放开的 `getSummaryDetailByNo` 对任意点击来源的 task_no 未限字符集，认证请求可达 API 源其他路径。
- **C4 容器/成员级联** — N/A：无鉴权层级改动；creator 校验为账号级等值比较。
- **C5 build 过≠运行期对** — 本环境无法 iOS 构建；所有关键行为按运行期路径推演（showUnread 时序即运行期推演所得）。
- **C6 治理文档** — N/A。

## 跨轮 blocker 复检（R6）

N/A — 本 issue 首轮审查，无历史评论/上轮 blocker。

## Verdict: CHANGES_REQUESTED

两条 P1 均为"通知/计数类副作用"且生产可达：showUnread 时序错误让提示对全员冲未读+响铃（与代码自述意图相反）；设备级去重无法兑现"不重发"承诺。建议先修 P1-1（一行时序问题）与 P1-2（幂等 clientMsgNo 或服务端化），P2 可同批处理。

[Octo-Q] verdict: REQUEST_CHANGES — 存在 2 条 P1（showUnread 线上仍为 true 致全员未读/提醒音；同账号多端重复发送），按 R1/R3/R4 不得 APPROVE。

yujiawei:
## Review addendum — corrections to my earlier review (same head SHA `0e618be4`)

Verdict is unchanged (**CHANGES_REQUESTED**), but a second pass over the send path found that **one of my P2s was wrong and understated the impact**, and surfaced **one blocking defect I missed entirely**. Both are corrected below. I'd rather publish the correction than leave an under-rated finding standing.

### Correction 1 — my "P2-2" was wrong: this is P1. The tip bumps unread and rings for *every* group member.

I wrote that `message.header.showUnread = NO` "does achieve the stated goal" and that the only residue was an inconsistent DB row. That was incorrect. There is a **third** copy of the header I did not account for — the wire packet — and it is snapshotted before the assignment runs.

`Modules/WuKongIMiOSSDK/WuKongIMSDK/Classes/manager/WKMessageQueueManager.m:49-52`

```objc
- (void)sendMessage:(WKMessage *)message {
    WKSendPacket *sendPacket = [WKSendPacket new];
    sendPacket.header.showUnread = message.header?message.header.showUnread:0;
```

And the path to it is fully synchronous — the async variant is commented out in `WKChatManager.m:286-318`:

```objc
-(WKMessage*) sendMessage:(WKMessage*)message addRetryQueue:(BOOL)addRetryQueue{
        if(addRetryQueue) {
            [[WKRetryManager shared] add:message];
        }
        [WKMessageQueueManager.shared sendMessage:message];
```

So by the time `sendMessage:tip channel:ch` returns to `OctoSummaryGroupNotifyHelper.m:98`, the `WKSendPacket` already holds `showUnread = 1`. The assignment at `:109` mutates only the in-memory `WKMessage` — it cannot reach a packet that was copied one frame earlier.

Receivers act on the transmitted value, in two places:

- Unread count — `WKChatManager.m:758-771`: `if(message.header.showUnread && ![message isSend]) { ... incUnreadCount += 1; }`
- Remind (sound/vibrate) — `WKSystemMessageHandler.m:136-144`: gated on `message.header.showUnread` and "not currently in this chat", and deliberately *not* on `isSend`.

Net effect: **every member of the source group gets a red-badge increment, and a notification sound if they aren't sitting in that group** — for a system announcement. That directly contradicts the intent stated in the code's own comment at `OctoSummaryGroupNotifyHelper.m:104` ("不该冲未读红点"). The half that *does* work is the narrow one: the sender's own device stays quiet, because the remind branch reads the live in-memory object.

Fix: set the flag before the snapshot. E.g. build the message with `saveMessage:content:channel:...`, set `header.showUnread = NO`, persist, then hand it to the `sendMessage:(WKMessage *)` overload — that path takes the header as-is. Worth confirming the Android sender uses the same convention so the three clients agree.

### Correction 2 — new P1 I missed: the create callback replaces an unrelated view controller

`Modules/OctoContext/OctoContext/Classes/Public/OctoSummaryCreateVC.m:601`

```objc
[[WKNavigationManager shared] replacePushViewController:detail animated:YES];
```

`replacePushViewController:` is unconditional about *what* it removes — `WKNavigationManager.m:74-79`:

```objc
-(void) replacePushViewController:(UIViewController*)viewController animated:(BOOL)animated {
   NSMutableArray *viewControllers = [NSMutableArray arrayWithArray:_rootNavigationController.viewControllers];
    [viewControllers removeLastObject];
    [viewControllers addObject:viewController];
```

The only in-flight guard on submit is `self.submitBtn.enabled = NO` (`:558`), which does nothing about the back button or the swipe-back gesture. So if the user taps submit and immediately goes back — a very ordinary impatient-user action, and the create POST is not instant:

1. Stack is `[…, ChatVC, CreateVC]`; the user pops → `[…, ChatVC]`.
2. The callback lands and calls `replacePushViewController:` → `removeLastObject` removes **`ChatVC`**, not the already-gone `CreateVC`.
3. Result: `[…, DetailVC]`. The chat page the user deliberately returned to is destroyed, and the detail page's back button now leads wherever `ChatVC` used to.

This is new in this PR — the previous code called `popViewControllerAnimated:` here, which was also unguarded but merely popped one page too many. Swapping in `replacePushViewController:` turns that into removing an arbitrary unrelated screen from the stack, so I'd rate it blocking rather than advisory.

Fix: verify ownership before replacing — `if (weakSelf.navigationController.topViewController == weakSelf)` (or `weakSelf.view.window != nil`), and otherwise skip the navigation entirely. Note that skipping it means the poll never runs and the tip never fires for that submission, so this interacts with the P1s about poll lifecycle in my earlier review.

### Positions I'm holding after re-checking

For the record, since these were raised at different severities:

- **`WaitingConfirm` gap — holding at P1, not P2.** It is not one missed tip, it is an entire task class (any confirmation-gated summary) that can never emit one, and the header comment at `OctoSummaryGroupNotifyHelper.h:9-10` claims that transition is covered. Documentation and behaviour must be reconciled either way.
- **`task_no` charset — holding at P1, not P2.** The forged-request surface alone would be P2; what makes it blocking is the composition with `OctoSummaryAPI.m:96-98`, where any 401 from the injected path calls `[[WKApp shared] logout]`. There is no response channel back to an attacker, but there is a side effect on the victim: a crafted chat link can force a logout during the eligibility window.
- **Cross-device duplicate — holding at P2 on likelihood, conceding the guarantee.** It needs two devices on the same account simultaneously parked on the same detail page through the completion window, which I still read as uncommon. But the `SENT` ledger is device-local, so the "宁可漏发不重发" claim in `OctoSummaryNotifyStore.h` is not true across devices at any likelihood. Either make the send idempotent (a deterministic `clientMsgNo` derived from `(taskId, channelId)` is the right shape, if the server dedupes on it — worth verifying before relying on it) or amend the claim to match what the code guarantees.
- **Poll lifecycle — I disagree that this area is clean.** `viewWillDisappear:` invalidates the timer at `:249`, *above* the `trulyLeaving` guard at `:255`, and `viewWillAppear:` never restarts it. So it is not just "stops on close as documented" — it also stops when the page is merely covered by a pushed child, permanently. That is the P1 in my earlier review.

Revised blocking set: the four P1s from my earlier review, plus the two above. Of the six, four are small local fixes (`displayContent`, header ordering, charset validation, nav ownership check); two need a behavioural decision (`WaitingConfirm` coverage, poll restart).

PR #86, opened 2026-09-01
fix(conversation-list): iPad 横屏下消息列表页返回后布局复位

swr614:
Closes #85

## What

iPad 横屏下，消息列表页的头部（搜索栏 / 分类 tab 行 / 导航栏）不能铺满整行，仍然按竖屏宽度显示；从其他页面（如打开一个会话）返回后同样会看到未铺满的状态。

## Why

- `setupFixedHeader` / `layoutFixedHeader` 依赖的 `WKScreenWidth` 宏取的是 `[UIScreen mainScreen].bounds.size.width`，从 iOS 8 起就被冻结在竖屏宽度，不随旋转变化，导致头部宽度计算始终按竖屏算。
- 该 VC 的重新布局原本只能靠 `viewWillTransitionToSize:withTransitionCoordinator:` 触发。但根据 Apple 已知问题 rdar://24277475，该回调在本 VC 不是导航栈最顶层控制器时（例如上面压了一个聊天页）不保证被调用，仅靠这一条路径无法覆盖所有场景。

## How

1. `setupFixedHeader` / `layoutFixedHeader` 改用 `self.view.lim_width`（跟随实际视图宽度），不再用被冻结的 `WKScreenWidth`。
2. 新增 `viewWillTransitionToSize:withTransitionCoordinator:` 覆盖，旋转时通过 `wk_relayoutNavigationBarForWidth:` + `layoutFixedHeader` 重新布局导航栏与固定头部。
3. `viewWillAppear:` 中同样执行一次上述重新布局，保证页面每次可见时都按当前宽度校正，不依赖旋转回调是否被调用。

## Checklist

- [x] Code compiles with no new warnings
- [ ] Linter and formatter pass
- [ ] Tests added/updated for the changes
- [ ] All tests pass locally
- [x] Rebased onto latest `main`
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Self-reviewed the full diff from a reviewer's perspective

## Testing

- 本地执行 `xcodebuild -workspace OctoiOS.xcworkspace -scheme OctoContext -sdk iphonesimulator build`，编译通过。

## Breaking Changes

N/A

PR #92, opened 2026-09-10
fix(im): 信号气泡在非已连接状态时不再展示虚假数据

swr614:
## What

修复信号连接气泡（点击顶部信号图标弹出的悬浮提示）在非「已连接」状态下展示虚假数据的问题。

## Why

`signalTapped` 此前无条件弹出气泡，且气泡内容完全脱离真实连接状态：

- 状态文案硬编码为 `"状态: 已连接"`，不管实际是 `WKDisconnected`、`WKPullingOffline` 还是 `WKConnecting`。
- 延迟展示的是 `currentLatencyMs` 的初始化哨兵值 `-1`（从未 ping 成功过），显示成 `延迟: -1ms`。
- 连接时长用 `connectedAtTime`（未赋值时为 `0`，即 1970-01-01）与当前时间相减，算出约 56.7 年 ≈ `29817069` 分钟的天文数字。

## How

- `signalTapped` 开头增加判断：`connectStatus != WKConnected` 时直接 `return`，非已连接状态不再弹气泡。
- `updateSignalViewForStatus:` 中联动设置 `signalContainerView.userInteractionEnabled`，非已连接状态禁用信号图标点击，避免可点但无响应。
- 断线（`onConnectStatus:` 非 `WKConnected` 分支）时重置 `currentLatencyMs = -1` 和气泡专用的 `signalSessionStart = 0`，并停止 ping 监控，避免重连后展示断线前的过期延迟/时长。`connectedAtTime` 保持原语义不动——它还被 `tryShowPixelHintForMessage:` 用作旧消息过滤截止线，气泡时长改用独立的 `signalSessionStart`，避免过载同一字段。
- `performPing` 的 completion 回调收敛到主线程处理，且仅当状态仍为 `WKConnected` 时才写入延迟并刷新导航栏，杜绝在途请求迟到后覆盖"已断开"状态及跨线程写 nonatomic 属性的竞争。
- 延迟展示兜底：`currentLatencyMs == -1`（哨兵值，未 ping 出过结果）时显示 `--ms`，与顶部导航栏兜底文案一致；时长 `signalSessionStart <= 0` 时显示 `已连接: 0秒`，不再用未初始化的 1970 纪元时间戳做减法。
- `en.lproj` 补充 `"延迟: --ms"` 英文词条。

## Checklist

- [x] Linter and formatter pass
- [ ] All tests pass locally  
## Testing

真机手动验证连接状态切换：
1. 关闭 Wi-Fi/蜂窝数据，等顶部信号图标变红显示"已断开"后点击 → 不再弹出气泡。
2. 恢复网络，在顶部显示黄色"连接中"的瞬间点击 → 不再弹出气泡。
3. 待顶部恢复绿色"已连接"后点击 → 正常弹出气泡，延迟显示真实 ms 或 `--ms`，时长从 0 开始正常计数，不再出现负数延迟或天文数字分钟。

## Breaking Changes

N/A


yujiawei (approved):
# Code Review — PR #92 (octo-ios)

## Summary

This is a tight, well-scoped fix for the signal tooltip showing fake data outside the connected state. The change is confined to `signalTapped` in `WKConversationListVC.m` and does not touch the real connection-status logic (`onConnectStatus:` / `updateSignalViewForStatus:`). Verified against head SHA `ccfdd8ea`.

## Spec compliance

The diff implements exactly what the PR description promises, nothing more:

- **Guard added** — `signalTapped` now returns early when `connectionManager.connectStatus != WKConnected`, so the bubble no longer appears while disconnected / connecting / pulling offline. ✅
- **Latency fallback** — `currentLatencyMs > 0` shows the real ms value, otherwise `--ms`, matching the top nav bar's `WKConnected` fallback (`WKConversationListVC.m:4858`). ✅
- **Duration fallback** — `connectedAtTime <= 0` renders `已连接: 0秒` instead of subtracting the un-set 1970 epoch timestamp. ✅

No missing work, no out-of-scope additions, no deviation from the stated approach.

## Correctness

- The early guard makes the still-hardcoded `状态: 已连接` status label correct, since the method now only runs in the `WKConnected` state.
- `connectedAtTime` is assigned on connect (`:567-568`, `:1477`) and reset to `0` at init (`:224`); the `<= 0` branch is a sound defensive fallback that eliminates the ~56-year duration artifact.
- `currentLatencyMs` is initialized to `-1` and only ever set to `50` / `500` / a real measured value, so `> 0` correctly filters the sentinel without dropping any legitimate reading.

No correctness, security, or data-loss concerns.

## Nits (non-blocking)

- The `状态: 已连接` status text remains hardcoded. It's correct today because of the new guard, but a brief comment noting that the label relies on the guard would help future readers avoid re-introducing the stale-status bug if the guard is ever relaxed.

## Verdict

No P0/P1 issues. Display-only change, no side effects on connection handling. Approving.

Jerry-Xin (dismissed):
This change fixes a real user-facing bug: previously, tapping the signal icon unconditionally popped a tooltip with a hardcoded "connected" status, a `-1ms` latency sentinel, and a ~56-year uptime computed from an unset (epoch-0) timestamp. Verified against the merge base: the old `signalTapped` performed the 1970-epoch subtraction unconditionally, so the fix is new and targets the described defect precisely.

Verification performed:

- Connection-state coverage: the early-return guard (`connectStatus != WKConnected`) covers all four non-connected states of `WKConnectStatus` (`WKNoConnect`, `WKConnecting`, `WKPullingOffline`, `WKDisconnected`), so no transitional state can reach the tooltip.
- Consumer census: `signalTapped` is the only caller of `showTooltip:` for this bubble, and no other code path renders the same latency/duration data into a tooltip, so no sibling display path is left unfixed. The hardcoded "connected" status label is now accurate because the bubble is only reachable in the connected state.
- Consistency with the sibling surface: the new latency fallback (`currentLatencyMs > 0`, else `--ms`) matches the existing top-bar `WKConnected` branch fallback in `updateSignalViewForStatus:`, keeping both surfaces consistent.
- Stale-data check: `connectedAtTime` is refreshed on every `WKConnected` transition (and in `viewDidAppear` when unset), and the tooltip is suppressed entirely while disconnected, so no stale/cached values can leak into the bubble after a disconnect.
- Concurrency: the tap handler and the connection-status delegate callbacks both run on the main thread, and the guard check and tooltip construction happen within the same turn, so there is no check-then-display race.
- Scope: only the independent, side-effect-free display branch is narrowed; the real status handling (`onConnectStatus:reasonCode:` / `updateSignalViewForStatus:`) is untouched.

💬 Non-blocking

- 🟡 `WKConversationListVC.m:4950` — `currentLatencyMs` is not reset on disconnect, so immediately after a reconnect the tooltip (and the top bar, which shares the same value) may briefly show the previous session's latency until the first ping completes. This is a pre-existing pattern unchanged by this PR; consider resetting it to `-1` when leaving `WKConnected` in a follow-up.
- 🔵 Consider adding focused tests for the state guard and the sentinel fallbacks (`connectedAtTime == 0`, `currentLatencyMs <= 0`).

✅ Highlights

- Live connection status is checked before constructing the tooltip.
- Sentinel timestamp/latency values now have safe display fallbacks.
- Narrow, side-effect-free scope; no security, performance, or memory-management concerns.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #92 (octo-ios)
Head: `ccfdd8ea226c9429c6ab2dbeb3186c905f9cf95c` · Base: `5b948cbe082cd0882ca611ad4941d656cbfa0a38` (`main`) · Range: `pr-base...HEAD` · 1 file, +19/-6

## Summary

The PR narrows the signal-tooltip path in `WKConversationListVC.signalTapped`: it returns early unless `[WKSDK shared].connectionManager.connectStatus == WKConnected`, guards the duration against an unset `connectedAtTime` (the 1970-epoch subtraction that produced ~29.8M minutes), and guards the latency line against the `-1` sentinel with a `--ms` fallback. The direction is right and the two numeric guards close the fake-data cases they target; the residual problems are the untranslated new string, the loss of all feedback in the non-connected states, and a latency-freshness gap the new `> 0` guard cannot cover.

## Verification

Static analysis only at head `ccfdd8ea`; build and tests not executed in this environment (no Xcode/macOS toolchain here).

- ✅ **Full diff read hunk-by-hunk** — both hunks are inside `signalTapped`; no other file or method touched (`git diff --name-status pr-base...HEAD`).
- ✅ **Gate semantics vs nav-bar state mapping** — `WKConnectStatus` has exactly 4 values (`Modules/WuKongIMiOSSDK/WuKongIMSDK/Classes/manager/WKConnectionManager.h:17-20`) and `updateSignalViewForStatus:` handles all of them (`:4843-4866`); `!= WKConnected` is exactly the yellow/red set, so there is no status-driven "bar looks connected but tooltip blocked" case.
- ✅ **Early return cannot orphan an overlay** — `showTooltip:atView:` self-dismisses via tap (`:4994-4996`) and a 3s `performSelector` (`:5002`), so returning early never leaves a stale tooltip on the key window.
- ✅ **Single call site** — repo-wide search finds `signalTapped` only as the tap target at `:707` and its definition at `:4916`; no programmatic caller depends on the removed side effect.
- ⚠️ **Locale resolution traced** — `LLang` (`Modules/WuKongBase/WuKongBase/Classes/WuKongBase.h:174`) → `NSString+WKLocalized.m:26-46` → module `Lang/en.lproj`; a missing key returns the Chinese key itself. English is user-selectable in-app (`Modules/WuKongBase/WuKongBase/Classes/Sections/Me/WKLanguageVM.m:33-35`).
- ⚠️ **Repo's own i18n guard executed** — `python3 scripts/i18n_audit.py` reports 177 missing keys at head, then dies with `KeyError: 'OctoContext'` at `scripts/i18n_audit.py:192` (pre-existing script bug, module absent from `MODULE_STRINGS` at `:36-42`). No workflow under `.github/workflows/` invokes it, so CI will not catch a missing translation.

## Findings

One P1 blocker and two P2 items below.

### P1 — New `LLang` key `延迟: --ms` has no English translation (`Modules/WuKongBase/WuKongBase/Classes/Sections/ConversationList/WKConversationListVC.m:4953`)

This hunk introduces a brand-new localization key, and `Modules/WuKongBase/WuKongBase/Assets/Lang/en.lproj/Localizable.strings:706-710` — the `// 连接状态` block that already carries the other four strings of this very tooltip (`状态: 已连接`, `已连接: %ld分钟`, `已连接: %ld秒`, `延迟: %ldms`) — does not contain it. Because `NSString+WKLocalized.m:26-46` returns the key on a lookup miss, an English-locale user gets a mixed-language tooltip: `Status: Connected` / `延迟: --ms` / `Connected: 0 s`.

The branch is reachable, not theoretical: `currentLatencyMs` is initialized to `-1` (`:226`) and only ever written by the async ping (`:4815`, `:4818`), which starts on `WKConnected` (`:1478`) with a 5s request timeout (`:4806`). Any tap in the window between a (re)connect and the first ping response passes the new gate at `:4918` and lands here. Per the diff-scope test this is **new**, and for the `en` locale it is strictly **worse than before**: the pre-PR line rendered through the translated `延迟: %ldms` key as `Latency: -1ms` — wrong value, right language; now it is right placeholder, wrong language. It is also exactly the regression `scripts/i18n_audit.py:5-12` says it exists to block ("新加 LLang 时容易忘了同步补英文翻译，结果英文模式下用户看到中文 key"), and that guard is both unwired from CI and currently crashing, so nothing downstream will catch it.

```
// Modules/WuKongBase/WuKongBase/Assets/Lang/en.lproj/Localizable.strings:710
 "延迟: %ldms"="Latency: %ldms";
+"延迟: --ms"="Latency: --ms";
```

One-line fix: add the key to the `连接状态` block; alternatively compose the line from the already-translated prefix so no new key is needed.

### P2 — Signal icon becomes a silent dead tap in three of four states (`:4918`)

The tap gesture is attached unconditionally (`:707-708`) and the container is force-shown in every handled state (`:4879`, `:4889`), so under `WKConnecting` / `WKDisconnected` / `WKPullingOffline` the icon still looks tappable but now does nothing at all — no toast, no affordance change. The sibling gate for the identical condition at `:897-905` follows the house convention and shows `请等待连接成功后再切换空间` (`:903`); this one returns bare.

There is also a reachable state where the bar *looks* connected and the tap still does nothing: `performPing`'s completion at `:4810-4823` writes `currentLatencyMs` and calls `updateSignalView` without re-checking `connectStatus`, and the in-flight task is never cancelled by `stopPingMonitoring` (`:4768-4777` only invalidates the timer). A ping landing just after a disconnect therefore overwrites the red `已断开` with a green ms value; the user taps a green-looking icon and gets silence. That overwrite is pre-existing, but this PR is what converts the tap into a no-op there.

Disconnected is precisely when a user taps the signal icon to find out what is wrong. Either toast like `:903`, or better: keep the tooltip and drive its status line from the real state (`状态: 已断开` / `连接中...`) instead of hardcoding `状态: 已连接` at `:4941` — the tooltip is the only surface exposing connection detail.

### P2 — `currentLatencyMs > 0` guards presence, not freshness (`:4950`)

The new guard cannot distinguish a live measurement from a stale one, so the tooltip can still print the previous session's latency — the same class of fake data this PR set out to remove. `currentLatencyMs` is never reset when the connection drops: the non-connected branch of `onConnectStatus:` (`:1480-1482`) only calls `stopPingMonitoring`, and that method explicitly leaves the value alone (`:4768-4777`). On reconnect, `:1477-1478` refreshes `connectedAtTime` and starts a new async ping, so a tap before that ping returns passes the gate at `:4918` and shows e.g. `延迟: 45ms` measured against the dead session — right next to a correctly reset `已连接: 0秒`, which makes the pair look self-contradictory.

Reset the latency sentinel in the disconnect branch so the new `--ms` fallback covers this window too:

```objc
} else {
    // 连接中或已断开，停止 ping 监控
    [self stopPingMonitoring];
    self.currentLatencyMs = -1;   // 避免重连后气泡显示上一次会话的 stale 延迟
    self.connectedAtTime = 0;     // 对称清理，防止任何路径读到旧会话起点
}
```

## 数据流回溯 (Data-flow traceback)

- `[WKSDK shared].connectionManager.connectStatus` → consumed at `:4918`; source `WKConnectionManager.m:897-898` returns `connectStatusInner`, written by `changeConnectStatus:` (`:490-495`) under `connectStatusLock`. **Flows correctly**; note the getter reads without the lock while the SDK mutates on global queues (`:507`, `:554`) — a benign pre-existing scalar race shared with `:566`, `:739`, `:897`, not introduced here.
- `self.connectedAtTime` → consumed at `:4924`/`:4927`; declared `:102`, zeroed `:224`, written at `:1477` (on `WKConnected`) and `:567-568` (`viewDidAppear` when already connected). **Flows**: inside the new gate it is normally > 0, so the `<= 0` branch is defensive rather than live; the removed 1970-epoch subtraction was reachable only through that same unset path, so the guard is aimed at the right hole. Also consumed at `:5440` as a message-age gate — untouched.
- `self.currentLatencyMs` → consumed at `:4950`; declared `:104`, init `-1` at `:226`, written only by `performPing` (`:4783`, `:4798`, `:4815`, `:4818`), **never cleared on disconnect** → stale value does flow into the next connected session (Finding 3).
- `LLang(@"延迟: --ms")` → consumed at `:4953`; resolution chain `WuKongBase.h:174` → `NSString+WKLocalized.m:20-46` → module bundle `Lang/en.lproj`. **Does not resolve for `en`** — key absent, fallback returns the Chinese key (Finding 1).
- Tooltip lifecycle → `showTooltip:atView:` `:4970-5003`; dismiss paths `:4994-4996` (tap) and `:5002` (3s). **No orphaned-overlay path** from the early return.

## 盲点 checklist (C1–C6)

- **C1 双路径 parity — HIT (2)**. show↔dismiss pair intact (`:4967` vs `:4994`/`:5002`); connect↔disconnect cleanup is asymmetric (`:1477-1478` sets state, `:1480-1482` clears nothing) → Finding 3. Guard-tool self-coverage variant also hit: the repo's guard for this defect class (`scripts/i18n_audit.py`, regex at `:51` does cover `LLang(@"...")`) is not invoked by any workflow in `.github/workflows/` and crashes at `:192`, so the "audit will catch it" assumption is a blind spot → Finding 1.
- **C2 control-flow ordering / 嵌套复用 — CLEAR**. `signalTapped` has exactly one call site (`:707`); the early return is the first statement, so no downstream ordering or nesting assumption breaks. Non-canonical-input probing (leading whitespace, unicode newlines, case/hyphen variants) is N/A — no parsing, regex, escaping, or sanitization in the diff.
- **C3 授权边界 ≠ 能力边界 — N/A**. No permission, jail, tool-exposure, or credential surface; display-only gate.
- **C4 授权生命周期 / 级联 — N/A**. No auth or container/member status involved.
- **C5 build/note 通过 ≠ 运行期路径正确 — HIT (addressed)**. Cannot build here, so "compiles / linter passes" from the PR checklist was not accepted as mitigation. Runtime reasoning done instead: enum coverage at `:4843-4866`, reachability of the new `--ms` branch (init `-1` + async ping + 5s timeout), locale fallback traced through the actual resolver, and the stale-ping overwrite at `:4810-4823` that makes the bar look connected.
- **C6 治理/策略/文档自洽性 — HIT (1)**. No policy docs in the diff, but the PR description's claim "与顶部导航栏 `WKConnected` 分支的兜底文案保持一致" is only half true: the nav-bar fallback at `:4858` is a raw `@"--ms"` (ASCII, unwrapped, needs no translation), while the new tooltip string wraps Chinese text in `LLang` and therefore does. That asymmetry is the root cause of Finding 1. Test checklist boxes are honestly left unchecked — no test exists for this VC path.

## 跨轮 blocker 复检 (R6)

N/A — this is the first review round for head `ccfdd8ea226c9429c6ab2dbeb3186c905f9cf95c`; the review issue carries no prior comments and no earlier blocker list to re-verify.

## 额外发现 (out of diff scope, non-blocking)

- `updateSignalView` at `:4892` formats `%ldms` with no `> 0` fallback and would render `-1ms`; all current callers set a positive value first (`:4784`, `:4799`, `:4821`) or guard (`:4852`), so it is unreachable today. Pre-existing and untouched by this PR.
- `scripts/i18n_audit.py:192` raises `KeyError: 'OctoContext'` because `MODULE_STRINGS` (`:36-42`) lacks that module — the only i18n guard in the repo is currently dead. Pre-existing; worth its own issue, not a blocker here.
- The `connectedAtTime <= 0` branch duplicates the `已连接: %ld秒` format string (`:4925` vs `:4930`); collapsing it to `seconds = 0` and falling through would remove the duplication. Style only.

- Gate-to-state alignment: no connection state renders a status-driven "connected" bar while blocking the tooltip.
- Both numeric fake-data cases named in the PR description (1970-epoch duration, `-1ms` latency) are genuinely closed by the two new guards.
- Format-specifier/type agreement: `%ld` against `(long)seconds`, `(long)(seconds / 60)`, and `0L` — all correct, including the new zero branch.
- No sibling tooltip implementation elsewhere in the repo still carrying the hardcoded `状态: 已连接` fake-data pattern.

## Verdict: CHANGES_REQUESTED

The behavioral fix is correct and well-scoped, but the new `延迟: --ms` key ships without an English translation, which regresses a previously localized line for English users on a reachable post-connect window and violates the repo's own documented i18n contract — that is a user-visible defect introduced by this diff, so it blocks. It is a one-line addition to `en.lproj:710`; clearing it plus a decision on the two P2s (feedback for the dead tap, latency reset on disconnect) is enough to land this.

[Octo-Q] verdict: REQUEST_CHANGES — 1 个 P1（新增 LLang key `延迟: --ms` 缺 en.lproj 翻译，英文模式下气泡渲染中文 key，属本 PR 新引入且对该 locale 比修复前更糟）+ 2 个 P2（非连接态点击变静默死点击且与 `:903` 同类 gate 的 toast 约定不一致；`currentLatencyMs > 0` 只校验存在不校验新鲜度，重连后仍会显示上一次会话的 stale 延迟）。按 R4，存在 P1 → 不给 APPROVE。P1 修复成本一行，建议连同两个 P2 一起处理后即可 merge。

yujiawei (changes requested):
# Code Review — PR #92 (octo-ios) — updated

Superseding my earlier approval. On a second pass I opened the localization resources, which I had not checked before, and found a reachable regression this diff introduces. Verified against head SHA `ccfdd8ea`.

## Spec compliance

The three behavioral fixes match the PR description and are correct:

- WKConnected guard on `signalTapped` — correct; blocks the bubble in disconnected / connecting / pulling-offline states.
- `connectedAtTime <= 0` duration fallback — correctly eliminates the 1970-epoch subtraction artifact.
- `currentLatencyMs > 0` latency guard with `--ms` fallback — closes the `-1` sentinel display.

No missing or out-of-scope work at the behavioral level.

## Blocking (P1)

### Missing English translation for the new `延迟: --ms` key — `WKConversationListVC.m:4953`

This hunk introduces a brand-new `LLang` key `延迟: --ms`. The `连接状态` block in `Modules/WuKongBase/WuKongBase/Assets/Lang/en.lproj/Localizable.strings:707-710` already carries the other four strings of this exact tooltip (`状态: 已连接`, `已连接: %ld分钟`, `已连接: %ld秒`, `延迟: %ldms`) but does **not** contain the new one.

The resolver (`NSString+WKLocalized.m:42-45`) returns the raw key on a lookup miss, so an English-locale user gets a mixed-language tooltip: `Status: Connected` / `延迟: --ms` / `Connected: 0 s`.

The branch is reachable, not theoretical: `currentLatencyMs` is initialized to `-1` (`:226`) and only written by the async ping (`:4815`, `:4818`), which starts on connect. Any tap in the window between a (re)connect and the first ping response passes the new gate and lands on this line. For the `en` locale this is strictly worse than before the PR — the pre-PR line rendered through the translated `延迟: %ldms` key (`Latency: -1ms`: wrong value, right language); it now renders the Chinese key verbatim.

Fix is one line — add the key to `en.lproj:710`:

```
 "延迟: %ldms"="Latency: %ldms";
+"延迟: --ms"="Latency: --ms";
```

Alternatively, compose the line from the already-translated `延迟: %ldms` prefix so no new key is needed. (Note the nav-bar fallback at `:4858` is a raw ASCII `@"--ms"` that needs no translation — so the PR description's "consistent with the nav-bar fallback" claim is only half true; that asymmetry is the root of this finding.)

## Non-blocking (P2)

### Signal icon is a silent dead tap in the three non-connected states — `:4918`

The tap gesture is attached unconditionally (`:707`) and the container is force-shown in every state, so under `WKConnecting` / `WKDisconnected` / `WKPullingOffline` the icon still looks tappable but does nothing. Disconnected is exactly when a user taps to find out what's wrong. The sibling gate for the identical condition at `:897-905` follows the house convention and shows a toast (`请等待连接成功后再切换空间`, `:903`); this one returns bare. Consider a toast, or better, keep the tooltip and drive its status line from the real state instead of hardcoding `状态: 已连接` at `:4941`.

### `currentLatencyMs > 0` guards presence, not freshness — `:4950`

`currentLatencyMs` is never reset on disconnect: the non-connected branch of `onConnectStatus:` (`:1480-1482`) only calls `stopPingMonitoring`, which just invalidates the timer (`:4768-4777`). On reconnect, a tap before the new ping returns shows the previous session's latency (e.g. `延迟: 45ms`) next to a correctly reset `已连接: 0秒` — the same class of stale/fake data this PR set out to remove. Reset the sentinel in the disconnect branch:

```objc
} else {
    [self stopPingMonitoring];
    self.currentLatencyMs = -1;   // avoid showing last session's stale latency after reconnect
    self.connectedAtTime = 0;
}
```

## Verdict

The behavioral fix is correct and well-scoped, but the new `延迟: --ms` key ships without an English translation and regresses a previously-localized line for English users on a reachable post-connect window. That is a user-visible defect introduced by this diff, so it blocks. The fix is a one-line addition to `en.lproj`; clearing it (and ideally the latency-reset P2) is enough to land this.

Jerry-Xin (dismissed):
**Re-review (fix confirmation) @ 5a4e55ed — adjudication of prior blocking findings**

This round byte-verifies the follow-up commit against the three blocking findings raised on the previous head (ccfdd8e) by @mochashanyao and @yujiawei. Each finding was adjudicated individually (true on old head? fixed on new head?); all three are FIXED, and a fresh full-diff walk of both commits found no new blocking issue.

**1. Missing English translation for "延迟: --ms" (both reviewers) — FIXED**
- TRUE on old head: the fallback key entered the code at `WKConversationListVC.m:4958` (`LLang(@"延迟: --ms")`) with no entry in `en.lproj/Localizable.strings`; the localization helper returns the key itself on a miss, so the English UI showed the raw Chinese string.
- FIXED: `en.lproj/Localizable.strings:711` adds `"延迟: --ms"="Latency: --ms";`. Verified byte-identical key between the code lookup and the strings entry (same colon/space/dash bytes), plain string with no format specifiers matching its non-`stringWithFormat:` use, exactly one occurrence (no duplicate key), and the file passes plist lint.
- Sibling-surface census of every localization key referenced by PR-touched code (both commits): 状态: 已连接 (en:707), 已连接: %ld分钟 (en:708), 已连接: %ld秒 (en:709), 延迟: %ldms (en:710), 延迟: --ms (en:711), 连接中... (en:677), 已断开 (en:323) — all present in `en.lproj` with correct `%ld` specifiers matching the `(long)` casts at the call sites. `zh-Hans.lproj` intentionally carries none of these keys: the helper's key-fallback returns the Chinese source string, the same established pattern as the pre-existing sibling entries 707–710.

**2. Icon clickable but unresponsive when not connected (both reviewers) — FIXED**
- TRUE on old head: the container was created with `userInteractionEnabled = YES` (`:708`) and never updated, while the new early-return guard in `signalTapped` made taps silently inert in every non-connected state.
- FIXED: `updateSignalViewForStatus:` now sets `signalContainerView.userInteractionEnabled = (status == WKConnected)` at `:4843` — placed before the switch, so it executes for every enum value including `WKNoConnect` and unknown values (the `default: return` path cannot skip it).
- Disable/re-enable symmetry traced end to end: the method runs on every status transition (`onConnectStatus:` → `:1460`, the VC's only connection-status delegate entry; delegates are removed only in `dealloc` at `:5396`, so off-screen transitions are covered) and at container creation (lazy getter `:739`, syncing the initial state to live status). On reconnect the SDK runs `handleConnected` → `WKPullingOffline` → sync completion → `WKConnected`, and every `changeConnectStatus:` unconditionally invokes the status delegates (`WKConnectionManager.m:490-495`, `:659-690`) — so the transition back to connected always re-enables the tap. No disable-once-never-re-enable path exists. The live-status guard in `signalTapped` (`:4922-4925`) remains as a second layer.

**3. currentLatencyMs not reset on disconnect → stale latency after reconnect (@yujiawei) — FIXED**
- TRUE on old head: no reset existed; after reconnect the `WKConnected` branch (`:4857-4858`) and the tooltip both rendered the pre-disconnect value.
- FIXED: the else branch of `onConnectStatus:` resets `currentLatencyMs = -1` and `connectedAtTime = 0` before stopping the ping monitor (`:1481-1483`). The else branch covers every non-connected status and is the VC's only state-exit path — reset coverage is complete.
- Post-reset consumers audited: top bar falls back to green `--ms` when latency is `-1` (`:4861-4863`); `updateSignalView` is never invoked with the sentinel (all call sites are gated on fresh ping results or `> 0`); the tooltip's `connectedAtTime <= 0` → "0 s" guard (`:4929-4931`) keeps the epoch-0 uptime guard intact, and the latency fallback (`:4955-4959`) mirrors the top bar. Also traced the non-display consumer of `connectedAtTime` (pixel-hint timestamp filter, `:5445`): offline-backlog delivery lands after the `WKConnected` delegate (SDK sync ordering), so re-arming the value on reconnect strengthens backlog suppression; the brief pulling window where the value is 0 behaves exactly like the pre-existing cold-start semantics (filter inert, hint dedup and visibility checks unchanged) — benign.

**@mochashanyao's 🟡 (latency validity check doesn't cover all staleness cases) — partially addressed, still-open-acceptable:** the disconnect reset removes the primary staleness case. The remaining sources are inherited design rather than new bytes: the `500` failure sentinel (`:4820`), the `50` default when no API base URL is configured (`:4785`, `:4800`), and the 30 s polling cadence.

💬 Non-blocking

- 🟡 In-flight ping race (inherited structure, untouched by this PR): a HEAD request issued up to 5 s before a disconnect can complete after the reset and overwrite `currentLatencyMs` (`:4817-4826`, on the session completion queue). It self-heals on reconnect — `startPingMonitoring` pings immediately (`:4762-4763`) — so a residual value can only flash briefly between the `WKConnected` callback and the first fresh ping. Strictly narrower than the old-head behavior; a follow-up could cancel the task in `stopPingMonitoring` or re-check connection status in the completion before applying the value.
- 🔵 No automated tests cover status transitions or tooltip formatting (acknowledged N/A in the PR checklist); the change is display-only with manual verification steps documented in the PR.

✅ All three prior blocking findings are fixed with correct, symmetric, and fully covered changes; the localization census is clean across all locale files; fresh-eyes re-walk of all PR-owned bytes (both commits, every hunk) surfaced no new blocking issue.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #92 (octo-ios)

## Summary

The PR stops the header signal bubble from presenting invented data whenever the socket is not in `WKConnected`: `signalTapped` early-returns on `[WKSDK shared].connectionManager.connectStatus != WKConnected` (`WKConversationListVC.m:4922`), `signalContainerView.userInteractionEnabled` is driven by the same status (`:4843`), the tooltip falls back to `延迟: --ms` when `currentLatencyMs <= 0` (`:4955`) and to `已连接: 0秒` when `connectedAtTime <= 0` (`:4929`), and the non-connected branch of `onConnectStatus:` now resets `currentLatencyMs = -1` / `connectedAtTime = 0` (`:1481-1482`). One `en.lproj` key is added for the new placeholder.

The UI gating itself is coherent — both gates read the same authoritative status, and the `connectedAtTime <= 0` branch incidentally kills a pre-existing `已连接: 29694912分钟` rendering. The blocker is not in the bubble: `connectedAtTime` has a second consumer, and zeroing it turns that consumer's guard into a no-op.

## Verification

- ✅ **Both gates agree on one source** — `changeConnectStatus:` writes `connectStatusInner` before `callConnectStatusDelegate` (`WKConnectionManager.m:664-691`), so `signalTapped`'s direct read of `connectStatus` cannot contradict the `status` that set `userInteractionEnabled`.
- ✅ **`WKPullingOffline` covered** — the enum (`WKConnectionManager.h:15-21`) includes it and both the existing switch (`:4862`) and the new guard treat it as non-connected, matching the stated intent.
- ✅ **`-1` never reaches the header label** — every `updateSignalView` caller assigns a positive latency first (`:4785`, `:4800`, `:4817-4820`) and the `WKConnected` branch re-checks `> 0` (`:4857`), so `-1ms` cannot be rendered.
- ✅ **Interaction-gate parity** — `rightAddItem` sets `userInteractionEnabled = YES` at creation (`:708`) then re-applies the status gate (`:739`); `hiddenRightItem:` never touches it, so no path re-enables the tap while disconnected.
- ✅ **Main thread** — `callConnectStatusDelegate` hops to main when off-main (`WKConnectionManager.m:682-688`), so the new UIKit write at `:4843` is not a background-thread mutation.
- ✅ **i18n** — `Localized:` returns the key when the table has no entry (`NSString+WKLocalized.m:44-47`) and `zh-Hans.lproj` is a partial stub containing none of the connection keys, so the en-only addition at `Localizable.strings:711` matches the existing pattern; no duplicate key.

Static analysis only at head `5a4e55ed5584df8d067730e60d232ab6923101d6`; build and tests not executed in this environment.

## Findings

One P1 blocker and one non-blocking P2.

### P1 — Zeroing `connectedAtTime` disables the pixel-hint stale-message guard (`Modules/WuKongBase/WuKongBase/Classes/Sections/ConversationList/WKConversationListVC.m:1482`)

`connectedAtTime` is not private to the signal bubble: `tryShowPixelHintForMessage:` uses it as the "only hint for messages newer than this connection" gate at `:5445` — `if (self.connectedAtTime > 0 && message.timestamp > 0 && message.timestamp < self.connectedAtTime) return;`. Zeroing the field makes the whole condition short-circuit to false, so for the entire disconnected / connecting / pulling-offline window that filter does nothing and any `lastMessage` is treated as fresh.

This is reachable offline with the list on screen: long-press menu 清空聊天记录 (`:3981`) → `WKMessageManager clearMessages` → `WKChatManager.m:418` `resetConversationLastMessage:message:nil` → `WKChatManager.m:471` `callOnConversationUpdateDelegate:` → this VC's `onConversationUpdate:` (`:1546`/`:1586`/`:1614`) → `tryShowPixelHintForMessage:` with a DB-read `lastMessage` that can be hours or days old, so the pixel-particle HUD pops that old message (sender + digest) as if it had just arrived. `WKReminderManager.m:110-128` (`done:` → `updateConversations:`) and `WKConversationManager.m:498`/`:534` are the same shape — local-first, DB-read conversations, delegate fired before any network call.

Scope is *amplified*, not pre-existing: before this PR `connectedAtTime` kept the last connect time (`> 0`) through a disconnect and the guard suppressed those messages; the field was only `0` between launch and the first `WKConnected`. The reset therefore widens a cold-start-only hole to every disconnect window for the whole app lifetime, and the window is unbounded when `syncConversations:` fails with a non-404 error — `WKConnectionManager.m:938-940` logs and never emits `WKConnected`, so the status sticks at `WKPullingOffline` with `connectedAtTime == 0` until the heartbeat timeout forces a reconnect.

Smallest fix — stop overloading the field, give the bubble its own session marker:

```objc
// :1481-1482 — reset only the bubble's own state
self.currentLatencyMs = -1;
self.signalSessionStart = 0;   // new property; :4929/:4932 read this instead
// connectedAtTime stays monotonic so :5445 keeps working
```

Or, if the field must stay shared, restore the hard gate that `9890f1b2` dropped at `:5445`:

```objc
if (message.timestamp > 0 && (self.connectedAtTime <= 0 || message.timestamp < self.connectedAtTime)) return;
```

### P2 — The new `> 0` guard covers the tooltip only; the header label can still show live-looking latency after a disconnect (`Modules/WuKongBase/WuKongBase/Classes/Sections/ConversationList/WKConversationListVC.m:4955`)

`stopPingMonitoring` (`:4770`) only invalidates the timer — it never cancels the HEAD request already in flight. That request's completion handler (`:4817-4825`) assigns `currentLatencyMs` and calls `updateSignalView`, which unconditionally writes `latencyLabel.text = "<n>ms"` plus a green tint (`:4897-4917`) over the red 已断开 state that `updateSignalViewForStatus:` rendered a moment earlier. So the "non-connected state must not show false data" contract this PR establishes for the bubble is still violated on the always-visible header label, and the late ping repopulates `currentLatencyMs` right after the new reset at `:1481`, so the next `WKConnected` renders that stale number instead of `--ms`.

Scope is pre-existing — this PR touches neither `performPing` nor `updateSignalView`, and base behaviour is identical, so not a change I'd block on; it is one line to close in the same pass. Fix: return early from the ping completion when `[WKSDK shared].connectionManager.connectStatus != WKConnected`, or cancel and nil the task inside `stopPingMonitoring`.

## Data-flow trace

- `connectedAtTime` — producers: init `0` (`:224`), `viewDidAppear` when already connected (`:567-568`), `onConnectStatus:` `WKConnected` (`:1477`), **new** reset to `0` (`:1482`). Consumers: bubble duration (`:4929`/`:4932`) and the pixel-hint guard (`:5445`). Only the first consumer was accounted for → P1.
- `currentLatencyMs` — producers: init `-1` (`:226`), ping results `50` / measured / `500` (`:4785`, `:4800`, `:4817`, `:4820`), **new** reset `-1` (`:1481`). Consumers: `:4857` (guarded `> 0`), `:4897`/`:4901`/`:4904` (unguarded, only reachable with a positive value except via the in-flight-ping race), `:4955` (new guard). Verified no `-1ms` render path.
- `connectStatus` — `connectStatusInner` is written under `connectStatusLock` before delegates are notified, and `onConnectStatus:` receives the value read at dispatch time (`WKConnectionManager.m:684`), so the rendered state and the tap guard cannot diverge.
- Status machine — `onlyConnect` emits `WKPullingOffline` → `WKConnecting` (`WKConnectionManager.m:165-169`); `handleConnected:` emits `WKPullingOffline` (`:920`) and only emits `WKConnected` from the `syncConversations:` completion (`:927`/`:936`), which is why the non-connected window is entered on every (re)connect and can be entered permanently on a non-404 sync failure.

## Blind-spot checklist

- **C1 double-path parity** — clear. Tap gate (`:4922`) and interaction gate (`:4843`) both use `status == WKConnected`; creation (`:708`) is immediately re-gated (`:739`); `hiddenRightItem:` does not resurrect interactivity.
- **C2 control-flow ordering / gate short-circuit** — **hit → P1**. `:5445`'s `A && B && C` gate is short-circuited by the new `connectedAtTime = 0`, and the reset is ordered before the ping stop, so a late ping can refill the field it just cleared (P2).
- **C3 authorization boundary vs capability boundary** — N/A, no permission, credential, jail, or endpoint surface in this diff.
- **C4 authorization lifecycle / container-member cascade** — N/A, no auth or membership state touched.
- **C5 build/note passing ≠ runtime path correct** — applied: no build was run (no Xcode in this environment), so every claim above is from runtime path reasoning over the SDK sources, not from a green build. The P1 was confirmed by following the real delegate chain into `WKChatManager` / `WKReminderManager` / `WKConversationManager`, not by trusting the reset's comment.
- **C6 governance / policy / security doc self-consistency** — N/A, no docs or policy files changed; the only resource change is one localization key, checked for duplicates and fallback.

## Cross-round blocker recheck

N/A — automated review on head `5a4e55ed5584df8d067730e60d232ab6923101d6`; this leg has no prior-round blocker list in scope.

- `已连接: 0秒` fallback (`:4929`) removes the pre-existing 1970-epoch duration that rendered as tens of millions of minutes.
- `statusLabel.text = LLang(@"状态: 已连接")` is now accurate by construction, since the guard guarantees `WKConnected` before the bubble is built.
- The tooltip auto-dismisses after 3 s (`:5008`), so a bubble open at the moment of a disconnect cannot linger as stale UI.
- `延迟: --ms` placement in the `// 连接状态` block is correct and the key is not duplicated elsewhere in the table.

## Verdict: CHANGES_REQUESTED

The bubble-side gating is right and the `--ms` / `0秒` fallbacks are a real improvement, but `connectedAtTime` is shared state: zeroing it silently switches off the pixel-hint stale-message filter at `:5445` for every non-connected window, which can pop an in-app notification for a message that is hours old — user-visible wrong output that did not happen before this change, and unbounded in the stuck-`WKPullingOffline` case. Splitting the bubble's session marker from `connectedAtTime` (or restoring the `<= 0` hard gate) is enough to clear the P1; the P2 is optional.

yujiawei (changes requested):
# Code Review — PR #92 (octo-ios)

Reviewed at head `5a4e55e`, diff against `main`. The fix goal is sound: the signal tooltip previously popped unconditionally with hardcoded "Connected", `-1ms` latency, and a ~56-year duration computed from an unset `connectedAtTime` (epoch 0). Gating the bubble on `WKConnected` and adding fallback text is the right direction. Two blocking issues below, plus a scope-description mismatch.

## 1. Scope / description accuracy

The PR "How" section states it does **not** modify `onConnectStatus:` / `updateSignalViewForStatus:`. The diff modifies **both**:

- `onConnectStatus:` else-branch now resets `self.currentLatencyMs = -1; self.connectedAtTime = 0;` (`WKConversationListVC.m:1481-1482`). On `main` this branch was only `[self stopPingMonitoring]`.
- `updateSignalViewForStatus:` now sets `self.signalContainerView.userInteractionEnabled = (status == WKConnected)` (`WKConversationListVC.m:4843`).

The changes themselves are reasonable, but the description understates the actual surface area. Please correct the "How" section so reviewers and future archaeology match the diff.

## 2. Blocking findings

### P1-A — Resetting `connectedAtTime = 0` disables the stale-message hint filter

`connectedAtTime` is not display-only. It is also the cutoff in the pixel-hint path:

```objc
// WKConversationListVC.m:5445 (tryShowPixelHintForMessage)
if (self.connectedAtTime > 0 && message.timestamp > 0 && message.timestamp < self.connectedAtTime) return;
```

On `main`, after a disconnect `connectedAtTime` retained the last connection's value (> 0), so messages older than that timestamp were suppressed. This PR now zeroes it on `WKConnecting` / `WKPullingOffline` / `WKDisconnected`, which makes the guard a no-op (`connectedAtTime > 0` is false).

During `WKPullingOffline` the SDK pulls offline messages and forwards synchronized conversations through delegates (`WKConversationManager.m` sync path → `onConversationUpdate` → `tryShowPixelHintForMessage` at `WKConversationListVC.m:1546/1586/1614`). The DB/merge work is dispatched to a background queue and only reaches the main thread later. In the nominal path `WKConnected` fires first and re-stamps `connectedAtTime`, restoring the cutoff. But if the link flaps back to Connecting/Disconnected before that queued update lands on the main thread, the cutoff is 0 and an old offline-pulled message can now trigger a pixel hint that `main`'s behavior suppressed. This is a new regression introduced by this PR. Suggest either keeping a separate persistent hint cutoff, or rejecting hints unless currently `WKConnected`, rather than overloading `connectedAtTime`.

### P1-B — In-flight ping completion overwrites the disconnected state with fake latency

`stopPingMonitoring` (`WKConversationListVC.m:4770`) only invalidates the timer; it does not cancel the outstanding `NSURLSession` HEAD request (up to 5s timeout). On disconnect the main thread sets `currentLatencyMs = -1` and `updateSignalViewForStatus:` renders red "已断开". A ping that was already in flight then completes on the background session queue, writes `currentLatencyMs = 500` (or a stale positive value) at `WKConversationListVC.m:4817/4820`, and calls `updateSignalView` on the main queue, which unconditionally repaints the nav-bar signal label to e.g. "500ms" with a latency color — overwriting the just-set "已断开". That directly contradicts this PR's goal of not showing bogus data when not connected (the tooltip itself is safe behind the `!= WKConnected` guard, but the top-bar indicator is not).

This is also a genuine data race: `currentLatencyMs` is `nonatomic` (`WKConversationListVC.m:104`) and is written from both the main thread (`onConnectStatus:`) and the background session completion handler. Gate the completion handler by a connection/ping generation token and drop late results, and keep the property mutation on the main thread.

## 3. Non-blocking notes

- **Latency `> 0` excludes a legitimate 0ms.** `if (self.currentLatencyMs > 0)` at `WKConversationListVC.m:4955` (and the analogous nav-bar branch at `:4857`) treats a real 0ms measurement as "no data" and shows `--ms`. In practice HEAD latency to a remote server is never 0 and the disconnected sentinel is `-1`, so this is cosmetic — but `>= 0` (or `!= -1`) expresses the intent more precisely. (nit)
- **i18n coverage is correct as-is.** The new `"延迟: --ms"` key is added only to `en.lproj`. This project ships only `en` + `zh-Hans`, and `zh-Hans` intentionally has no entry for these signal strings (nor for the existing `"延迟: %ldms"`), relying on `NSLocalizedStringFromTableInBundle` returning the key (Chinese) verbatim. The addition matches the existing pattern exactly — no other locale needs it. No action needed.
- **Monotonic clock (optional).** Connection duration uses wall-clock `timeIntervalSince1970`; an NTP/user clock adjustment mid-connection can skew it. `CACurrentMediaTime()` would be more robust, but this is pre-existing and out of scope for this fix.

## Verdict

Requesting changes for P1-A (hint-filter regression) and P1-B (in-flight ping overwriting disconnected state + data race). The core tooltip fix is good; these are the reset/threading interactions the reset logic pulls in.

Jerry-Xin (approved):
# Code Review — PR #92 (octo-ios), round 3 @ `c40d4727`

**Verdict: APPROVE.** All four blocking findings raised by peers on the previous head (`5a4e55ed`) are fixed or addressed at byte level; the cumulative diff carries no new blocking issue; one bounded non-blocking residual remains and is recommended as a fast-follow.

## Under-call acknowledgment (round 2)

My previous review approved the disconnect-time reset of `currentLatencyMs` / `connectedAtTime` as full-coverage and graded the pixel-hint consumer in `tryShowPixelHintForMessage:` benign. That call was wrong: zeroing `connectedAtTime` disabled the stale-message cutoff for the entire disconnected window. mochashanyao and yujiawei independently caught this (reviews on `5a4e55ed`), and their data-flow traces — offline-pulled conversations and local-first paths (clear-chat, reminder updates) reaching `tryShowPixelHintForMessage:` while the cutoff is 0 — are confirmed correct at byte level. Credit to both reviewers; this round re-audited the full consumer set rather than the display path alone.

## Per-finding adjudication (new head `c40d4727`)

### 1. mocha P1 / yujiawei P1-A — `connectedAtTime` reset breaks the pixel-hint stale-message filter → FIXED

- The disconnect branch now resets only the bubble-owned fields: `currentLatencyMs = -1` and `signalSessionStart = 0` (`WKConversationListVC.m:1489-1490`). `connectedAtTime` is no longer touched there.
- Full consumer census of `connectedAtTime` at the new head — writes: init (`:225`), `viewDidAppear` lazy fill gated on `WKConnected` (`:569-570`), `onConnectStatus:` connected branch (`:1482`); single read: the hint cutoff (`:5459`). The read site is byte-identical to merge-base, and the write lifecycle now matches merge-base exactly (no reset anywhere). Baseline filter semantics fully restored.
- New `signalSessionStart` is bubble-only: writes at `:226`, `:572-573`, `:1483`, `:1490`; reads only in `signalTapped` (`:4943-4946`). A grep across all modules finds no other consumer. This is exactly the field-separation remedy mocha prescribed, and it also satisfies the finer-grained-control note (avoid affecting unrelated logic): `connectedAtTime` keeps its single legacy meaning.

### 2. yujiawei P1-B / mocha P2 — in-flight ping completion overwrites the disconnected display + cross-thread write to a nonatomic property → FIXED (mechanism), with a non-blocking residual

- The completion handler now computes latency locally, dispatches to the main queue, early-returns unless `connectStatus == WKConnected`, and writes `currentLatencyMs` only on the main thread (`:4823-4841`).
- Threading verified end-to-end: the SDK delivers `onConnectStatus:` on the main thread (`WKConnectionManager.m`, `callConnectStatusDelegate` hops to main when off-main), and `pingTimer` is scheduled from main-thread callers (`viewDidAppear`, `onConnectStatus:`), so every write of `currentLatencyMs` — including the `50ms` default paths (`:4793`, `:4808`) — is main-thread. The data race is eliminated.
- The named symptom (late ping repainting the nav bar over the red disconnected state) is eliminated by the status guard, and the SDK writes `connectStatusInner` before notifying delegates, so the guard cannot read a stale connected status after a disconnect notification.
- **Residual (non-blocking):** there is no generation token / task cancellation, so a ping issued in session A that lands after a quick A→disconnect→B reconnect (within the ~5s request window) passes the status guard and writes a seconds-old measurement. Graded non-blocking because: (a) the only affected value is the displayed latency number — a real measurement to the same server, or the `500ms` poor-network sentinel — never the connection state, session start, or hint cutoff; (b) `startPingMonitoring` performs an immediate fresh ping on reconnect (`:4771`), bounding staleness to ~one RTT in the common case and one 30s interval worst-case, after which it self-heals; (c) the class is inherited and strictly narrowed: at merge-base the same completion wrote `currentLatencyMs` unconditionally from the session queue thread — including while disconnected — so this PR removes the harmful members of the class (disconnected-state overwrite, cross-thread race) and leaves only the cosmetic one; (d) mocha graded this same mechanism P2 non-blocking on the previous head. **Recommended fast-follow:** capture a generation counter incremented in `startPingMonitoring` and drop mismatched completions, or retain the task and cancel it in `stopPingMonitoring`.

### 3. yujiawei #1 — PR description does not match the actual changes → FIXED

- The live PR body's How section now describes exactly the implemented surface: the `onConnectStatus:` reset with `connectedAtTime` semantics preserved, the `userInteractionEnabled` gating in `updateSignalViewForStatus:`, the ping main-thread + status-guard convergence, the `--ms` / `0秒` fallbacks, and the `en.lproj` entry. Each of the six claims was checked against the diff — all match. The earlier statement that these two methods were untouched is gone.

### 4. Non-blocking notes from the previous round

- yujiawei's `> 0` sentinel nit: adopted in the tooltip (`!= -1` at `:4969`), which renders the full value domain `{-1, 50, 500, measured}` truthfully. The nav-bar `> 0` branch (`:4871`) is pre-existing untouched code.
- i18n: confirmed no action needed (see census below).

## Fresh-eyes walk of the cumulative diff (merge-base → `c40d4727`)

- **State machine:** both timestamps are set together on `WKConnected`; only the bubble field is cleared on non-connected states; the `viewDidAppear` lazy fill is double-gated (`WKConnected` and `== 0`). No path can display the 1970-epoch duration or `-1ms`, and no path lets a session marker survive across a disconnect.
- **Interaction gate:** `userInteractionEnabled = (status == WKConnected)` (`:4857`) runs on every status update after the container-nil guard; the initial setup applies the same gate (`:744`); `signalTapped`'s early return (`:4937`) is defense in depth. No path re-enables tapping while disconnected.
- **Tooltip status line:** the hardcoded connected label is now only reachable behind the `WKConnected` guard, so it is truthful by construction.
- **Localization census:** `"延迟: --ms"` added to `en.lproj` only, matching its sibling signal keys (`"延迟: %ldms"`, `"已连接: …"`); `zh-Hans.lproj` carries no entries for this family and falls back to the Chinese key by design. No duplicate keys.
- **Scope hygiene:** the delta commit touches only `WKConversationListVC.m` (23+/9−); cumulative changes are that file plus one `en.lproj` line. No unrelated edits.

## Non-blocking observations

1. Cross-session stale-latency residual above — generation-token fast-follow recommended.
2. No automated tests cover the now-central state transitions (tap while disconnected, sentinel rendering, late completion after disconnect, quick reconnect). The checklist declares manual device verification; adding unit coverage for `signalTapped` gating and the ping-completion guard would lock the fix in.

## Conclusion

APPROVE at `c40d47278e7a4d8aa34c72fe4a39f78c6b569edf`. The fix adopts the peers' prescribed remedies (field separation; main-thread + status-guarded ping completion), restores merge-base semantics for the pixel-hint cutoff, and the description now matches the diff. The remaining residual is bounded, cosmetic, self-healing, and strictly narrower than base behavior.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #92 (octo-ios)

## Summary

本 PR 修的是会话列表导航栏「信号气泡」在非已连接状态下展示虚假数据的问题：`signalTapped` 增加 `connectStatus != WKConnected` 提前返回，`updateSignalViewForStatus:` 在非已连接态关掉容器交互，气泡延迟行在 `currentLatencyMs == -1` 时回退到新增的 `延迟: --ms` 文案，断线分支重置 `currentLatencyMs` 与新增的 `signalSessionStart`，并给 `performPing` 的在途回调补上「非已连接不落值」的守卫。第三个 commit 把原先复用 `connectedAtTime` 的写法拆出独立的 `signalSessionStart`，避免动到旧消息过滤的截止线。改动共 2 文件 +44/-11，范围收敛、方向正确。

整体没有发现 P0/P1：核对下来「已连接时长」不再可能算出以 1970 纪元为起点的天文数字，断线后导航栏也不会被迟到的 ping 回调刷回延迟数字。余下两项 P2 都属于「修复没覆盖到的同类残留路径」，非阻塞。

## Verification

Static analysis only at head `c40d4727`; build and tests not executed in this environment.

- ✅ **审查范围与 PR 元数据一致** — 本地 checkout `HEAD=c40d47278e7a4d8aa34c72fe4a39f78c6b569edf`，`pr-base=5b948cbe082cd0882ca611ad4941d656cbfa0a38`（同时是 merge-base），`git diff pr-base...HEAD` 为 2 文件 +44/-11。
- ✅ **旧消息过滤截止线未被破坏** — `tryShowPixelHintForMessage:` 的唯一消费点 `:5459` 仍读 `connectedAtTime`；断线 else 分支(`:1486-1491`)只重置 `currentLatencyMs` / `signalSessionStart`，`connectedAtTime` 全程不清零，断线窗口内过滤仍生效。
- ✅ **`-1` 哨兵不会漏成 `-1ms`** — `updateSignalView`(`:4911`) 的三个调用点都保证 `currentLatencyMs >= 0`：`:4793`/`:4808` 先写 50，`:4838` 先写实测值，`:4871` 由 `currentLatencyMs > 0` 门控。
- ✅ **新守卫对枚举穷尽** — `WKConnectStatus` 实为 5 值（`WKNoConnect`/`WKConnecting`/`WKPullingOffline`/`WKConnected`/`WKDisconnected`，见 `Modules/WuKongIMiOSSDK/WuKongIMSDK/Classes/manager/WKConnectionManager.h:15`），`!= WKConnected` 与 `== WKConnected` 两个新判定连带覆盖了 `default: return;`(`:4888`) 漏掉的 `WKNoConnect`。
- ✅ **关闭容器交互不误伤兄弟控件** — `pcOnlineBtn` 与加号按钮都是 `_rightAddItem` 的直接子视图(`:739-742`)、由 `relayoutRightItems`(`:789`) 按 x 递增排布互不重叠，并非 `signalContainerView` 的子视图，故 `userInteractionEnabled = NO` 不会连带禁用 PC 在线图标或加号。
- ✅ **英文翻译补齐且符合仓库约定** — `zh-Hans.lproj/Localizable.strings` 不含任何连接状态类 key（中文原文即 key，缺失时回落自身），故只需在 `en.lproj/Localizable.strings:711` 增补 `延迟: --ms`；本 PR 已补，无遗漏语种。

## Findings

无 P0/P1；两项 P2 非阻塞建议如下。

### P2 — 在途 ping 守卫只认连接状态、不认连接会话 (`Modules/WuKongBase/WuKongBase/Classes/Sections/ConversationList/WKConversationListVC.m:4835`)

新守卫判断的是「回调落地这一刻是否已连接」，而不是「这次 ping 属于哪一次连接会话」。断线→重连若发生在 ping 的 5s 超时窗口内（移动网络抖动下并不罕见），迟到回调照样通过 `connectStatus == WKConnected` 检查，把跨断线窗口测得的延迟、或失败分支硬编码的 `500`，写进**新会话**，再由 `updateSignalView`(`:4911`) 当作实时延迟刷上导航栏。本 PR 刚引入的 `signalSessionStart` 正是会话标识，却没有用在这里。

同一方法内另有两条路径完全绕过该守卫：`:4793` 与 `:4808` 在 `dispatch_async` 之外直接写 `self.currentLatencyMs = 50`（未经任何测量的伪造值），随后无条件刷新 UI——这与新注释宣称的「断线后迟到的回调不能把导航栏刷回延迟数字」不变量相矛盾。已核对该分支的实际可达性：`apiBaseUrl` 由 `Modules/WuKongLogin/WuKongLogin/Classes/Login/Src/WKLoginSettingVC.m:238` 用 `http(s)://ip:port/v1/` 拼装、恒带 scheme，`Octo/AppDelegate.m:126` 走编译期常量，故这两条 fallback 现网几乎不可达，属一致性缺口而非现网故障，不单独定级。

```objc
// [task resume] 之前捕获会话标识
NSTimeInterval session = self.signalSessionStart;
// 回调主队列块内改为按会话判定（可叠加 connectStatus）
if (weakSelf.signalSessionStart != session || session <= 0) { return; }
weakSelf.currentLatencyMs = latency;
[weakSelf updateSignalView];
```

修复方向：用 `signalSessionStart` 做会话比对替代裸状态判定，并把 `:4793`/`:4808` 的赋值一并移进主队列块、套用同一守卫。

### P2 — 已经在屏上的气泡不会随断线关闭 (`Modules/WuKongBase/WuKongBase/Classes/Sections/ConversationList/WKConversationListVC.m:1490`)

本 PR 只堵住了「打开气泡」这一条路径。`showTooltip:atView:`(`:4990`) 把气泡 addSubview 到 `keyWindow`(`:4997`)，靠 `performSelector:withObject:afterDelay:3.0`(`:5023`) 延时关闭；`afterDelay:` 注册在 `NSDefaultRunLoopMode`，用户点开气泡后立刻滑动会话列表时 runloop 处于 tracking mode，3 秒定时不会触发，气泡会一直挂着（且挂在 window 而非 `self.view`，push 新页面也不会带走它）。这期间若连接断开，导航栏已被 `updateSignalViewForStatus:` 刷成红色「已断开」，屏幕上却仍浮着「状态: 已连接 / 延迟: N ms / 已连接: N分钟」的旧数据——正是 PR 标题要消除的那类假数据，只是从「打开时」挪到了「已打开后」。

```objc
// onConnectStatus: 的 else 分支内
[NSObject cancelPreviousPerformRequestsWithTarget:self
                                         selector:@selector(dismissTooltip:)
                                           object:nil];
[self dismissTooltip:self.visibleSignalTooltip]; // 需把 tooltipView 存成属性
```

修复方向：把 tooltipView 提为属性，在断线分支取消延时并立即 dismiss。

## 数据流回溯

逐个消费点回溯上游生产点，确认运行时数据真的流到消费处：

- **`signalSessionStart`** → 消费于 `:4943`（`<= 0` 兜底）与 `:4946`（时长计算）。生产点：`:226` init=0、`:573` viewDidAppear（仅当已连接且为 0）、`:1483` onConnectStatus 已连接分支（= `connectedAtTime`）、`:1490` else 分支归零。已连接态下两个生产点（delegate + viewDidAppear 兜底）覆盖完整，数据确实流到消费点 ✅。唯一残留窗口：容器在 `viewWillAppear`(`:520` → `rightAddItem` → `:744`) 就已建好并可点，而兜底赋值在 `viewDidAppear`(`:572-574`) 才跑；若 SDK 在 `addDelegates`(`:262`) 之前就已连接，这段出现动画窗口内 `signalSessionStart` 仍为 0 → 气泡显示「已连接: 0秒」。窗口极窄，且严格优于 base 的「1970 纪元 → 约 2900 万分钟」，不构成 finding。
- **`currentLatencyMs`** → 消费于 `:4871`（门控）、`:4911`（导航栏文案）、`:4969-4970`（气泡延迟行）。生产点：`:228` init=-1、`:1489` 断线=-1、`:4793`/`:4808` 伪造 50、`:4838` 实测值。三条 UI 刷新路径均保证非负，`-1` 不会漏成 `-1ms` ✅；残留问题是跨会话陈旧写入（见 P2 第 1 项）。
- **`connectedAtTime`** → 消费点仅 `:5459`（旧消息过滤截止线）。生产点 `:225`/`:570`/`:1482`，初始化后不再归零，断线窗口内过滤契约保持 ✅。这正是本轮拆出 `signalSessionStart` 要保住的语义。
- **`connectStatus`** → 新增消费点 `:4835`、`:4937`；既有消费点 `:566`、`:744`。来源为 SDK 单例的 readonly 属性，运行期确有数据流入 ✅。
- **`LLang(@"延迟: --ms")`** → `en.lproj:711` 命中；zh-Hans 无该 key，回落为中文原文 ✅。

## 盲点 checklist

- **C1 双路径 parity — HIT（见 P2 第 1 项）**：`performPing` 内三条 `updateSignalView` 路径只有网络回调那条加了守卫，另两条 fallback 未加。`signalSessionStart` 的三处 set ↔ 一处 reset 配对完整；`userInteractionEnabled` 的开/关由同一表达式驱动，无单边分支。本 PR 不含守卫/lint 工具自身，等价写法覆盖 N/A。
- **C2 control-flow ordering / 嵌套复用 — CLEAR**：`updateSignalViewForStatus:` 仅两个调用点(`:744` 容器创建、`:1465` onConnectStatus)，均已核对。`onConnectStatus:` 先刷 UI(`:1465`) 再改状态字段(`:1482-1491`)，因此重连瞬间 `currentLatencyMs` 仍是 -1 → 走 `:4877` 的 `--ms` 占位而非上一会话旧值，随后 `startPingMonitoring` 内同步调用 `performPing`(`:4771`) 在约一个 RTT 内补上真值，不会长时间停在占位文案。本 PR 无正则/escape/sanitize 类安全控件，非 canonical 形式试穿 N/A。
- **C3 授权边界 ≠ 能力边界 — N/A**：纯客户端 UI 状态展示，无权限、jail、tool 暴露或凭证改动。
- **C4 授权生命周期 / 容器-成员状态级联 — N/A**：无鉴权改动，无父容器 status 字段参与查询或中间件。
- **C5 build/note 通过 ≠ 运行期路径正确 — HIT（已按要求做运行期推演）**：本环境为 Linux、无 iOS toolchain，未执行 xcodebuild，故不以「build 过」作缓解；改以运行期推演覆盖——5 值枚举穷尽性、`-1` 哨兵不外漏、`performSelector:afterDelay:` 的 runloop mode 语义（P2 第 2 项）、`apiBaseUrl` 的真实拼装来源（据此把 fallback 分支降级为一致性缺口）。新注释宣称的不变量一律不当作已成立，逐路径 grep 复核后才定级。
- **C6 治理/策略/安全文档自洽性 — N/A**：无治理或安全策略类改动，唯一的资源文件变更是 1 行 i18n 增补。

## 跨轮 blocker 复检

本 PR 为第 3 轮。按本轮 leg 约束未查询 GitHub 历史评论，轮次与上轮 blocker 由本地 commit 序列还原：`ccfdd8ea`（首版修复）→ `5a4e55ed`（未连接态禁用交互 + 补英文翻译 + 断线重置过期数据）→ `c40d4727`（「修复 review 指出的 connectedAtTime 过载与在途 ping 回调问题」）。当前审查对象为 head `c40d47278e7a4d8aa34c72fe4a39f78c6b569edf`。

1. **`connectedAtTime` 语义过载（气泡会话起点 vs 旧消息过滤截止线）→ 已修**。证据：新增 `signalSessionStart`(`:103`)，气泡改读 `:4943`/`:4946`；断线 else 分支(`:1486-1491`)只重置 `signalSessionStart` 与 `currentLatencyMs`，`connectedAtTime` 不清零，消费点 `:5459` 的过滤契约保持。
2. **在途 ping 回调把导航栏刷回延迟数字、覆盖「已断开」→ 部分修**。已修部分：`:4835` 新增 connectStatus 守卫，且赋值移入主队列块(`:4838`)，消除与 `onConnectStatus:` 重置之间的写竞争。未修部分：守卫只认状态不认会话，跨 5s 超时窗口的断线→重连仍会被旧回调污染；`performPing` 另两条 fallback 路径(`:4793`/`:4808`)完全绕过守卫。详见 Findings P2 第 1 项。

无 P0/P1 级 blocker 未修，故本轮不因跨轮复检强制 REQUEST_CHANGES；轮次为第 3 轮，未触及「第 4 轮及以上应停打地鼠、出整体方案重审」的升级线。

- 禁用交互不会把点击漏给加号按钮或 PC 在线图标：三者是 `_rightAddItem`(`:739-742`) 的同级子视图，frame 由 `relayoutRightItems`(`:789`) 递减排布且互不重叠；`_rightAddItem` 自身是无手势的普通 UIView，禁用后点击区确实「什么都不发生」，与 `signalTapped`(`:4937`) 的提前返回行为一致，不存在双重响应。
- 顺带一提：`:4857` 的 `userInteractionEnabled` 切换在用户侧不可观察（图标与文字外观不变），注释里「避免图标可点但无响应」的效果实际没有达成——改动安全，但若想给出可感知的禁用态，可同时下调 `signalImageView`/`latencyLabel` 的 alpha，或保留可点并在非连接态弹一条「当前未连接」说明。非阻塞。
- 容器创建时序安全：`:713` 先置 `userInteractionEnabled = YES`，紧接 `:744` 用真实 `connectStatus` 校正；`updateSignalViewForStatus:` 的 `!signalContainerView` 提前返回(`:4852`) 位于新赋值之前，不存在「已断开但可点」的初始化窗口。
- `signalSessionStart <= 0` 的兜底文案「已连接: 0秒」(`:4943-4945`) 仅在上述 appear 动画窗口内可达，是诚实的未知值，明显优于 base 的天文数字。
- `NSInteger latency;`(`:4823`) 在 if/else 两支均完成赋值后才被主队列块读取，无未初始化读取；`weakSelf` 为 nil 时整段静默 no-op。
- `WKNoConnect` 全仓无任何赋值点（仅枚举声明 `WKConnectionManager.h:16` 与日志映射 `Modules/WuKongBase/WuKongBase/Classes/Services/Logs/WKMessageSendTracer.m:169`），因此 `updateSignalViewForStatus:` 的 `default: return;`(`:4888`) 不会在运行中把陈旧延迟文案长期留在导航栏；冷启动首帧的空标签属既有行为且随即被 `WKConnecting`/`WKConnected` 覆盖，不在本 diff 范围。
- delegate 生命周期：`addDelegates` 在 `:262`(viewDidLoad)、`removeDelegates` 只在 `dealloc`(`:5410`)，故离屏期间 `onConnectStatus:` 仍会驱动 `signalSessionStart`/`currentLatencyMs` 的重置与恢复；回前台不依赖 `viewDidAppear` 兜底也不会出现陈旧会话起点。`viewDidDisappear`(`:636`) 只停 ping、不清状态，与上述恢复路径不冲突。
- i18n 一致性：导航栏占位 `@"--ms"`(`:4877`) 是既有硬编码、无语种语义，未被本 diff 触碰；气泡侧走 `LLang(@"延迟: --ms")` 并已补英文，两者并存不影响正确性。

## Verdict: APPROVED

无 P0/P1。改动方向正确、范围收敛，上一轮的 `connectedAtTime` 语义过载已真正修好，`-1` 哨兵与 5 值枚举的穷尽性都经得起回溯，禁用交互也未误伤兄弟控件。两项 P2（在途 ping 守卫只认状态不认会话、且两条 fallback 分支绕过守卫；已在屏气泡不随断线关闭）都属于「修复未覆盖到的同类残留路径」，最坏后果是短暂展示一个跨会话的陈旧延迟数字，或一条延时关闭的气泡多挂几秒，不会让任何既有可用路径失效，也不产生持久化的错误数据，按 R1/R4 不构成阻塞。建议合并前顺手处理第 1 项（约 2 行，且可直接复用本 PR 刚引入的 `signalSessionStart`），第 2 项可随后跟进。

yujiawei (approved):
# Code Review — PR #92 (octo-ios)

Reviewed at head `c40d4727`, diff against merge-base `5b948cbe` (2 files, +44/-11). This is the third round; both blocking findings from round 2 are verified fixed. Static analysis only — no macOS/Xcode toolchain in this environment, so nothing below rests on a green build.

## 1. Spec compliance — ✅

**Missing work: none.** Both round-2 blockers are genuinely closed, not papered over:

- **`connectedAtTime` semantic overload → fixed.** A dedicated `signalSessionStart` property is introduced (`WKConversationListVC.m:103`) and the bubble now reads it (`:4943`, `:4946`). Critically, I diffed every *write* site of `connectedAtTime` against `main`: the only additions are `signalSessionStart` lines, and the disconnect branch (`:1486-1491`) no longer zeroes `connectedAtTime`. Its write sites are therefore byte-identical to `main`, so the stale-message cutoff at `:5459` behaves exactly as before the PR. That is the correct fix — the round-2 regression was introduced by zeroing a field with a second consumer, and the second consumer's contract is now provably untouched.
- **In-flight ping overwriting the disconnected state + data race → fixed.** The completion handler now converges on the main queue and returns early unless still `WKConnected` (`:4834-4840`). I verified the race is fully closed rather than narrowed: `callConnectStatusDelegate` (`Modules/WuKongIMiOSSDK/.../WKConnectionManager.m:673-691`) explicitly hops to the main queue when `![NSThread isMainThread]`, so `onConnectStatus:` always runs on main; the timer-driven `performPing` also runs on main. With the ping write moved to main, `currentLatencyMs` / `signalSessionStart` / `connectedAtTime` are now main-thread-confined, so the `nonatomic` declarations are safe.

**Out-of-scope additions: none.** `signalSessionStart` is precisely the remedy the previous round recommended. No new flags, fields, or public interfaces.

**Deviations: none.** The round-2 complaint that the "How" section understated the touched surface is resolved — the description now enumerates `onConnectStatus:`, `updateSignalViewForStatus:`, and `performPing` explicitly, and matches the diff.

One wording quibble, not a spec violation: "consistent with the top nav-bar fallback" is true of the *text* (both render `--ms`) but not of the *predicate* — see P2-3.

## 2. Code quality — Approved (no P0/P1)

Things I verified as sound:

- **Enum exhaustiveness.** `WKConnectStatus` has five values (`WKConnectionManager.h:15-20`). Both new predicates are `== WKConnected` / `!= WKConnected`, so they cover `WKNoConnect`, which the pre-existing `switch` drops through `default: return` (`:4888`). The `userInteractionEnabled` assignment is placed *before* the switch (`:4857`), so it executes for every value including the drop-through case — the cold-start state (`connectStatusInner` defaults to `0` == `WKNoConnect`) correctly yields a non-interactive icon.
- **No tap fall-through.** Disabling interaction on a container can silently reroute taps to whatever sits underneath. It does not here: `relayoutRightItems` (`:779-796`) places `signalContainerView` at x=0 w=60 and the next item at x=74, and `pcOnlineBtn` / the plus button are siblings under `_rightAddItem` (`:739-742`), not subviews of the signal container. No overlap, no newly-reachable button.
- **No path re-enables the tap while disconnected.** Only two writers of `userInteractionEnabled` exist (`:713` at creation, `:4857`), and `:744` re-applies the live-status gate immediately after creation. `hiddenRightItem:` does not touch it.
- **The `-1` sentinel cannot reach the nav bar.** All three `updateSignalView` call sites assign a non-negative value first (`:4793`, `:4808`, `:4838`) and the `WKConnected` branch re-checks (`:4871`).
- **i18n is correct and complete.** `延迟: --ms` is byte-identical between the call site and `en.lproj/Localizable.strings:711` (verified at the UTF-8 byte level), appears exactly once, and carries no format specifier — matching its non-`stringWithFormat:` use. `zh-Hans.lproj` deliberately contains none of the connection keys (nor the pre-existing siblings at 707-710); the resolver returns the key itself on a miss, so Chinese renders from the source string. The en-only addition is the established pattern, not a gap.
- **Ordering on reconnect.** `onConnectStatus:` repaints (`:1465`) before restamping the fields (`:1482-1491`), so at the moment of reconnect `currentLatencyMs` is still `-1` and the bar shows the `--ms` placeholder rather than the previous session's number, with `startPingMonitoring` filling in a real value within roughly one RTT.
- **No sibling-PR conflict.** #86 also touches this file but at `:354` / `:514` / `:737` — no hunk overlap and no semantic interaction. #89 touches `WKConnectionManager` but adds liveness probing without altering the status machine this PR depends on.

### P2 — In-flight ping guard keys on status, not on connection session (`WKConversationListVC.m:4835`)

The guard asks "are we connected *now*", not "does this reply belong to *this* connection". A disconnect→reconnect inside the request's 5 s timeout — unremarkable on mobile — lets a reply measured across the outage pass the check and land in the new session, after which `updateSignalView` (`:4911`) presents it as live. This PR already introduced the right discriminator; it just isn't used here:

```objc
// capture before [task resume]
NSTimeInterval session = self.signalSessionStart;
// inside the main-queue block
if (session <= 0 || weakSelf.signalSessionStart != session) { return; }
```

Worth noting this is materially milder than the round-2 blocker it descends from: that one repainted a *disconnected* bar as connected, whereas the residual shows a real measurement a few seconds stale while genuinely connected, self-corrected by the ping the reconnect issues. Non-blocking.

Two sibling paths in the same method bypass the guard entirely: `:4793` and `:4808` write a fabricated `currentLatencyMs = 50` outside the `dispatch_async` and refresh unconditionally, which contradicts the invariant the new comment at `:4832` asserts. I checked reachability before rating this: `apiBaseUrl` is assigned only at `WKLoginSettingVC.m:238` (built with an explicit `http://` / `https://` prefix at `:229-234`) and `Octo/AppDelegate.m:126` (compile-time constant), so both fallbacks are effectively dead. Consistency gap, not a live defect — but if you touch this method, fold them into the same guarded block.

### P2 — An already-visible tooltip is not dismissed on disconnect (`WKConversationListVC.m:1490`)

The PR closes the "open the bubble" path only. `showTooltip:atView:` adds the view to the **key window** (`:4997`) — not `self.view` — and relies on `performSelector:withObject:afterDelay:3.0` (`:5022`). That schedules in `NSDefaultRunLoopMode`, so while the user drags the conversation list the run loop is in tracking mode and the timer does not fire. Tap the icon, then scroll, and the bubble outlives its nominal 3 s; if the socket drops in that window the nav bar turns red "disconnected" while a "Connected / N ms / connected N min" card still floats above it — the exact class of false data this PR exists to remove, relocated from "on open" to "after open".

To be precise about the blast radius: the timer fires as soon as the run loop returns to default mode, i.e. when scrolling stops — so this extends the window by seconds, it does not leak the view indefinitely. Display-only, hence P2. Fix: hold the tooltip in a property, then `cancelPreviousPerformRequestsWithTarget:` and dismiss it in the non-connected branch.

### P2 — Sentinel predicate now differs between the two surfaces (`:4871` vs `:4969`)

The nav bar tests `currentLatencyMs > 0`; the bubble now tests `!= -1`. Two consequences. A genuine sub-millisecond reading truncates to `0` (plausible against a LAN-deployed server, which this product supports) and renders `--ms` on the bar but `延迟: 0ms` in the bubble. More to the point, `!= -1` re-admits *negative* values that `> 0` filtered: if the wall clock steps backward mid-request, `latency` at `:4826` goes negative and the bubble will happily print `延迟: -3000ms`. Prefer `> 0` (or `>= 0`) on both surfaces so the predicate matches the intent rather than one specific sentinel.

### Nits

- The duration uses wall-clock `timeIntervalSince1970`, so a backward clock adjustment mid-session yields a negative duration and a `Connected: -10 s` string. `MAX(0, ...)` clamps it; `CACurrentMediaTime()` would be structurally right. Pre-existing.
- `:4857` toggles `userInteractionEnabled` but nothing about the icon's appearance changes, so the comment's stated goal — "avoid an icon that looks tappable but does nothing" — isn't actually achieved from the user's side. The change is safe either way; if you want a perceivable disabled state, dim `signalImageView` / `latencyLabel`, or keep the tap live and show the real status in the bubble instead of the hardcoded "Connected" at `:4961`.
- `viewDidAppear` calls `[NSDate date]` twice (`:570`, `:573`) for two fields meant to be the same instant; one local would keep them identical and read better.

## 3. Verdict: APPROVED

No P0/P1. The two prior blockers are fixed for the right reasons — the shared-field overload is undone rather than worked around, and the ping race is closed by main-thread confinement rather than narrowed. The four P2s are all residual paths of the same class the PR is fixing; their worst case is a briefly stale latency number or a card that lingers a few extra seconds, with no persisted bad state and no existing path broken. Worth folding P2-1 in before merge — it is about two lines and reuses the `signalSessionStart` this PR just added — and the rest can follow up.

PR #94, opened 2026-09-14
fix(voice): 语音输入+图片托盘场景下未发送文本丢草稿

swr614:
## What

修复语音输入 + 图片托盘场景下，未发送的输入框文本没有被保存到草稿的问题；并修复该场景下草稿文本还原时，其依赖的 @ 提及状态被一并清空的问题。

## Why

用户先输入文本、选中图片（图片进入待发送图片托盘，尚未点发送），再用语音输入说一段话并发送：图片和语音识别文本会正常作为一条消息发出，但用户此前打在输入框里的原始文本（草稿）会被静默清空且从未保存过，切出会话页后再回来草稿丢失。
根因：WKConversationInputPanel.m 的 _commitPendingWithCaption:（有 pending 图片时的统一发送出口）发送后无条件把 self.textView.text 清空成 @""。手动点发送键路径下这是对的（caption 本来就是 textView 全部内容）；但语音发送路径（holdToTalkManager:sendText: / sendText:mentions:）传入的 caption 是语音识别文本，跟 textView 里用户选图后打的原草稿是两份不同的东西，清空逻辑没有区分二者。草稿的保存/恢复（WKConversationView.m 的 saveDraftOrKeepPosition / recoveryDraft）唯一数据源就是 textView.text，所以草稿随这次误清空一并丢失。

第一版修复让文本正确还原了，但 review 指出：图文聚合分支里还有一处无条件的 [ctx cleanMentionCache]，会把还原出来的草稿依赖的 @ 提及缓存也一并清空——草稿发出去后看得见 @，但 entities/mentionedInfo 是空的，@ 提醒静默失效，@所有人/@所有AI 的广播和 bot 路由也会跟着丢。这是第一版修复引入的新问题（base 分支上文本和缓存本来是一起清的，这种"文本还原了但缓存没还原"的不一致状态根本不可达）。

## How

第一步：还原草稿文本

给 _commitPendingWithCaption: 加 restoreText: 参数：

- 手动发送键路径（inputSendFinished）传 nil，行为不变（nil ?: @"" 仍是硬清空）。
- 两处语音发送回调（holdToTalkManager:sendText: / sendText:mentions:）在调用前先读一份 self.textView.text（用户原草稿）传入。
- 清空动作从硬编码 self.textView.text = @""; 改为 self.textView.text = restoreText ?: @"";，语音路径下还原原草稿而不是清空。
- 异步发送失败的合并恢复逻辑（原有代码，L1266 附近）无需改动——此时读到的 textView.text 已经是还原后的草稿，自动走现有"空则写回 caption / 非空则合并"分支，失败时语音文字与原草稿会合并显示，不会丢内容。

第二步：还原草稿依赖的 mention 缓存

- WKInputMentionCache.h：给 matchString:（正则提取文本里的 @name）补上头文件声明——原来只在 .m 里实现，外部调不到。
- WKConversationContext.h / WKConversationContextImpl.m：新增协议方法 removeMentionItemsMatchingText:，只移除某段文本里实际出现的 @ 项并返回被移除的 item；复用现成的 matchString: + removeName:，没有新写匹配逻辑。
- WKConversationInputPanel.m：_commitPendingWithCaption:restoreText: 里把无条件 cleanMentionCache 改成按 restoreText 分流——restoreText 为空（手动发送）走原有整体清，零回归；非空（语音还原了草稿）时改用 removeMentionItemsMatchingText:captionRaw，只清 caption 实际消费掉的 mention 项，草稿里的 @ 保留在缓存里。

## Checklist

- [ ] All tests pass locally

## Testing

1. 输入框打字（不发）→ 选图 → 语音说话并发送 → 输入框应恢复为原来打的文本，退出会话页再进入草稿仍保留。
2. 回归：有图 + 有文字，手动点发送键发送 → 发送后输入框正常清空（确认未受本次改动影响）。

## Breaking Changes

N/A


yujiawei (approved):
# Code Review — PR #94 (octo-ios)

## Summary

Fixes a draft-loss bug: when a user types text, attaches an image (into the pending tray), then uses voice input to send, the original typed draft was silently wiped because `_commitPendingWithCaption:` unconditionally cleared `textView.text` after send. The fix threads a `restoreText:` parameter through so the voice paths restore the pre-existing draft while the manual send path keeps its old hard-clear behavior. Scope is minimal (1 file, +20/-9) and well-targeted.

## Spec compliance ✅

The diff matches the PR description exactly:
- Adds `restoreText:` to `_commitPendingWithCaption:`.
- Manual send path (`inputSendFinished`) passes `nil` → `nil ?: @""` preserves the previous hard-clear.
- Both hold-to-talk STT callbacks (`sendText:` and `sendText:mentions:`) capture `self.textView.text` before the call and pass it as `restoreText`.
- The clear becomes `self.textView.text = restoreText ?: @"";`.

No over-scope, no missing pieces, no divergence from the stated approach.

## Correctness

- **Core logic is sound.** `restoreText ?: @""` gives `@""` on the manual path (behavior unchanged) and the captured draft on the voice paths.
- **Async failure path holds without modification** (author's claim verified). On `onFailure`, `currentText = self.textView.text` is now the restored draft, so `textOverwriteOnly` is false and the STT caption is merged in front of the draft with the "content merged into draft" HUD — no content is lost. When the draft was empty, `currentText` is `""` and it falls back to `setText:captionRaw`, identical to prior behavior.
- The only remaining caller of the renamed selector is updated; the method is private (no header declaration), so the signature change is build-safe.

## Non-blocking observations

- **Send-button / input-height state after restore (nit).** After restoring a non-empty draft, the method still forces `sendButton.show = NO; hidden = YES;` and `resetInputHeight` (single-line). This is harmless because STT only fires in voice mode (textView hidden, send button irrelevant), and toggling back to keyboard mode recomputes both the send-button visibility and the real text height via `resetCurrentInputHeight`. It self-heals, but the implicit dependency on voice mode is worth a brief comment for future maintainers.
- **Draft @mentions (pre-existing, not introduced here).** The aggregate branch calls `cleanMentionCache`, so a restored draft containing `@name` keeps the literal text but loses its mention-cache entry. This is consistent with the existing draft save/restore limitation (`recoveryDraft` → `inputSetText` never rebuilds the mention cache), so it is not a regression from this change.
- **No automated test** (checklist acknowledges this). The behavior is a UIKit interaction that is awkward to unit test; the two manual test steps in the PR cover the fix and the manual-send regression. Acceptable for this change.

## Verdict

No P0/P1 issues. The fix is correct, minimal, and the tricky failure/merge path was reasoned through correctly. Approving.

Jerry-Xin (dismissed):
**Verdict: APPROVE** (head `1aa26b03360a`, verified against merge-base `f5e7a010`)

The fix adds a `restoreText:` parameter to `_commitPendingWithCaption:restoreText:`. Both hold-to-talk STT send callbacks snapshot `textView.text` before committing and pass it as `restoreText:`, so the user's pre-existing draft survives a voice send while the pending-image tray is populated; the manual send path passes `nil`, preserving the previous hard-clear semantics exactly (`nil ?: @""` ≡ merge-base behavior at base line 1235).

## Verified (byte-level + repo-wide census)

1. **Call-site census complete.** The renamed selector has exactly three call sites (`WKConversationInputPanel.m:687`, `:1536`, `:1552`), all updated; a repo-wide grep finds no caller of the old `_commitPendingWithCaption:` selector. The CI `build` job at this head SHA is green — and it is a real compile (`pod install` + `xcodebuild build-for-testing` on the full workspace, with an explicit iOS-SDK ≥ 26 assertion), so whole-target compilability of the rename is machine-verified.

2. **Path census — every writer that can clear/overwrite unsent text audited:**
   - Manual send with no pending images (`inputSendFinished`, `:697`) and the text-to-file fallback (`sendTextAsFile`, `:749`): clear content that was just sent; unchanged.
   - Image tray flows (`WKMoreItemClickEvent.m`, incl. the atomicity-gate undo paths): write only the pending-image bar, never the text view.
   - Voice-message path (`holdToTalkManager:sendVoiceData:...`, `:1476`): sends voice (+ pending images separately), never touches the text view.
   - Transcribe path (`didTranscribeText` → `inputInsertText:`): inserts at cursor, preserves the draft.
   - `toggleVoiceMode` (enter and exit): hides/shows the text view without writing its text.
   - Voice sends with no pending images go through delegate `inputPanelSend:text:` (`WKConversationView.m:718`), which does not touch the text view.
   The touched commit path was the only writer that destroyed the unsent draft, and both of its voice-mode entry points are now covered.

3. **Restore cannot clobber newer input, resurrect stale text, or double-fire.** The STT send delegate fires synchronously from the hold-to-talk result overlay's send button (`WKHoldToTalkManager.m:1170-1188`), which lives in its own window — the input panel underneath is not interactive until `hideOverlay`, and in voice mode the text view is hidden with editing ended. Snapshot → commit → restore is one synchronous main-thread sequence, so no interleaved edit exists. Re-entrancy is guarded by `inFlightPendingSend` plus `hasPendingImages` (the bar is cleared in the same call, so a second callback degrades to the plain-text delegate path). On the early-return branches (in-flight, empty bar, missing context) nothing is cleared and the snapshot is harmlessly discarded.

4. **Async failure recovery composes correctly with the restore.** The `onFailure` merge logic (`:1284-1305`, merge-base bytes unchanged) reads `textView.text` at failure time — now the restored draft — and takes the existing "empty → write caption back / non-empty → prepend caption + separator" branch, so on upload failure both the voice caption and the original draft survive, with the HUD merge notice. The pure-image branch failure path restores only the bar, never the text.

5. **UI state after restore is consistent.** The restore happens while in voice mode, where the send button is deliberately hidden. Exiting voice mode recomputes `sendButton.show/hidden` and input height from the current text (`:1449-1458`), so a restored non-empty draft correctly re-arms the send button and multiline height on return to text mode.

6. **Draft persistence verified end-to-end.** `saveDraftOrKeepPosition` (`WKConversationView.m:356`) stores `inputText` when leaving the conversation and `recoveryDraft` (`:508`) restores it on entry — the restored draft now survives exactly the leave/return round trip described in the PR.

7. **No accidental mention side-effects on the outgoing voice message.** `allMentionUid:` filters cached items by presence in the sent text (`WKInputMentionCache.m`), so draft-owned mentions are not pulled into the voice message's `mentionedInfo`; the known `name=="all"` sentinel leak is guarded by the pre-existing `cleanMentionCache` on this branch.

## 🟡 Advisories (non-blocking)

- **A-1 — Restored draft returns without its mention-cache backing (mechanism verified true; inherited family gap, not a regression).** On the aggregate branch, `cleanMentionCache` (`:1257-1259`, merge-base bytes) wipes all mention items — including those backing the user's draft — and the restore at `:1242` writes plain text through a property setter, which does not re-run the mention pipeline. A draft containing real @mentions comes back visually intact but, if later sent unedited, goes out as literal "@Name" text with no mention entity/notification. Attribution: this exact degraded state is already reachable at merge-base through the no-pending-images voice path (`inputPanelSend:` → `sendTextMessage` → unconditional `mentionCache clean`, `WKConversationContextImpl.m:509`), which never touched the text view; on the path this PR changes, the merge-base outcome was strictly worse (the entire draft was destroyed). The fix is therefore a net improvement and creates no new failure class — the residual gap is that the restore is text-only. Suggested fast-follow: snapshot the draft-owned mention items alongside the text and re-add them after the caption's `cleanMentionCache`, coherently across both voice paths, taking care not to reintroduce the "@all" sentinel leak that the clean guards against.
- **A-2 — No automated regression test** for this stateful voice + tray + draft interaction (the PR checklist acknowledges this), and CI is compile-only (`build-for-testing`; no test-execution step in `ci.yml`). A unit/UI test pinning "voice send with pending images restores the prior draft" would lock the contract. Also noted for merge awareness only: the `check-sprint` status check (org-level project-metadata workflow, non-code) is currently red on this PR.

## Evidence base / scope

No local full-workspace build or simulator run was performed for this review; conclusions rest on byte-level verification against the merge-base, the repo-wide path census above, and the head-SHA CI compile (`build` green at `1aa26b03360a`). CI executes no tests, so runtime behavior claims are code-evidence-based.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #94 (octo-ios)

## Summary

单文件 +20/−9。给 `_commitPendingWithCaption:` 增加 `restoreText:` 参数：手动发送传 `nil`（保持原清空语义），按住说话的两个 STT 入口（`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKConversationInputPanel.m:1536` 与 `:1552`）在提交前抓一份 `textView.text`，发送后把这份草稿还原回输入框，修掉「选图 + 语音输入」场景下未发送草稿被连带清空的问题。方向是对的：STT 文本走的是 `WKHoldToTalkManager` 自己的 result overlay（`onResultSendText`，`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKHoldToTalkManager.m:1170`），面板侧唯一会把识别文本写进输入框的 `holdToTalkManager:didTranscribeText:`（`WKConversationInputPanel.m:1472`）在 manager 里没有任何调用方，所以 `draft` 抓到的确实是用户语音前打的草稿，不会和 caption 混淆。问题是这次只还原了文本，没有还原这段文本依赖的 mention 状态。

## Verification

Static analysis only at head `1aa26b03`; build and tests not executed in this environment（`OctoiOS.xcodeproj/project.pbxproj` 里没有 XCTest target，本仓无对应单测可跑）。

- ✅ **选择器改名无遗漏调用方** — 全仓 `_commitPendingWithCaption` 只有 `:687` / `:1536` / `:1552` 三处调用，均已改成两参形式；`WKMoreItemClickEvent.h:44`、`WKMoreItemClickEvent.m:128`、`:297` 只是注释引用，也不存在 `performSelector` / 字符串选择器调用。
- ✅ **手动发送路径行为等价** — `:687` 传 `nil`，`restoreText ?: @""`（`:1242`）与 base 的 `self.textView.text = @""` 完全一致，零回归。
- ✅ **草稿持久化不受影响** — `saveDraftOrKeepPosition` 在保存时读 `[self.input inputText]`（`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/WKConversationView.m:356`），还原后的草稿会正常写入 `remoteExtra.draft`，退出会话也不丢。
- ✅ **输入框高度 / send 按钮会被重算** — 语音→键盘切换时 `toggleVoiceMode` 按 `textView.text` 重算 `show`/`hidden` 并调 `resetCurrentInputHeight`（`:1454`、`:1457`、`:183`），`WKGrowingTextView setText:` 另外 dispatch 一次 `refreshHeight:`（`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKGrowingTextView.m:207`），所以还原非空草稿不会留下「有字但没有发送按钮 / 高度塌成一行」的可见状态。
- ✅ **pending 图闸门同源** — `hasPendingImages` → `imageCount` → `mutableImageDatas.count`（`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKConversationPendingImageBar.m:61`）与 `:1234` 的 `images.count == 0` 防御同一数据源，不存在 gate 通过但 `images.count == 0` 把 STT 文本静默吞掉的新口子。
- ❌ **mention 状态没有随文本一起还原** — 见下方 P1。

## Findings

一个 P1 阻塞项。

### P1 — 还原草稿只还原了文本，@ 提及静默失效（`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKConversationInputPanel.m:1242`）

`:1242` 把 textView 还原成 `restoreText`，但同一条聚合分支在 `:1258` 仍然无条件执行 `[ctx cleanMentionCache]`，而 `mentionCache` 是 mention 实体的**唯一**来源：`entities:` / `mentionedInfo:`（`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/WKConversationContextImpl.m:394`、`:428`）只遍历 `mentionCache.items` 并在待发文本里匹配 `@name`，`sendTextMessage:` 也是先取 entities 再 `[self.mentionCache clean]`（`:506`、`:509`），而能从文本反推 item 的 `matchString:`（`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/WKInputMentionCache.m:106`）全仓无调用方。草稿里的 @ 是用户点选人时写进缓存的（`WKConversationContextImpl.m:1074` → `addMentionToCache:` `:1088`），`toggleVoiceMode` 和 `hideMentionUsers`（`:299`）都不清缓存，所以 STT 提交那一刻草稿的 mention item 确实在缓存里，然后被 `:1258` 连同 caption 消费掉的 item 一起抹平。

可达后果：草稿 `@Alice 看下这个` + 待发送图 → 按住说话发出一条 STT caption → 图文消息正常发出、草稿文本被还原 → 用户切回键盘把这条草稿发出去，消息里带着肉眼可见的 `@Alice`，但 `entities` 为空、`mentionedInfo.uids` 为空，Alice 收不到提醒、客户端也不做 mention 高亮。`@所有人` / `@所有AI` 更严重：`humans=1` / `ais=1` 标记丢失，`WKConversationContextImpl.m:471` 起的 robot uid 展开也不会执行，全员广播和 bot 路由静默失效。`onFailure` 合并分支（`:1284`）把 caption + 草稿写回输入框时，缓存同样已经是空的。

diff-scope：new / newly-exposed。base 分支上同一次操作会把草稿文本一并清空，「带死 @ 的消息」根本发不出去；是本 PR 的 `:1242` 让这份文本活下来并变得可发送，缺陷由本次改动新暴露（`:1258` 的 blanket clean 本身是 `c6547b93` 为防 `@all` 泄漏刻意加的，手动路径 caption == textView 全量内容，那里依然正确，不要一起回退）。

修复方向：把清理范围收敛到 caption 实际消费掉的 item —— 在 clean 前快照 `mentionCache.items`，clean 后把 `@name` 仍出现在 `restoreText` 里的 item 用 `addMentionItems:` 写回；或改用 `WKInputMentionCache.m:90` 的 `removeName:` 只删 `captionRaw` 里匹配到的名字，`restoreText == nil`（手动路径）时保留现有 blanket clean。`onFailure` 合并路径需要同样保住这部分状态。

- 兄弟语音消息路径 `holdToTalkManager:sendVoiceData:`（`:1476`）全程不碰 `textView.text`，草稿天然不丢，与本 PR 的还原语义一致，无对称性缺口。
- `restoreText` 的形状边界安全：`nil` 由 `?: @""` 兜住；纯空白草稿不会让 send 按钮误显示（`:842` 与 `:1452` 都先 trim）；超长草稿还原时不校验 `messageTextMaxBytes` 也没问题，发送时 `:690` 仍会弹「转文档」提示。
- 还原路径不调用 `handleTextViewContentDidChange`，因此不会误触发 @ 候选列表或 typing 上报（`:848`），send 按钮与高度由上面 Verification 里的两条机制兜回；今天两个 STT 调用方都只在语音模式可达（`holdToTalkBtn` 仅语音模式可见），不存在「文本模式下有草稿但发送按钮隐藏」的可达组合。
- `draft` 抓取在 `addMentionItems:`（`:1546`）之后，但那只写缓存不写文本；`inFlightPendingSend` 在聚合分支内同步置位/复位（`:1264`、`:1308`），STT 路径撞不上这道闸。

## Verdict: CHANGES_REQUESTED

修复本身方向正确、调用方全覆盖、手动发送路径零回归，草稿持久化与输入框高度/发送按钮状态都已验证能自洽恢复。阻塞点只有一个：文本被还原了，但支撑这段文本的 mention 缓存已经在 `:1258` 被清空，用户随后发送这条还原草稿会得到一条「看得见 @、实际不通知任何人」的消息，`@所有人` / `@所有AI` 还会丢掉广播与 bot 路由。按上面的方式把清理范围收敛到 caption 消费掉的 item（并在 `onFailure` 合并路径同样处理）即可解除阻塞。

[Octo-Q] verdict: REQUEST_CHANGES — 存在 1 个 P1：还原草稿未还原 mention 状态（`WKConversationInputPanel.m:1242` 还原文本 / `:1258` 清空 mentionCache），导致还原后的草稿发出时 @ 静默失效、`@所有人`/`@所有AI` 广播与 bot 路由丢失，属用户可见错误数据；无其它 P2 阻塞项。

yujiawei (changes requested):
# Code Review — PR #94 (octo-ios)

## Summary

Single file, +20/−9. Adds a `restoreText:` parameter to `_commitPendingWithCaption:`: the manual send path passes `nil` (unchanged clear semantics), and the two hold-to-talk STT entry points (`WKConversationInputPanel.m:1535`, `:1551`) capture `self.textView.text` before committing and restore that draft into the input field after send. The direction is correct — STT text flows through `WKHoldToTalkManager`'s own result overlay (`onResultSendText`, `WKHoldToTalkManager.m:1170`), and the panel's only method that would write recognized text into the input field, `holdToTalkManager:didTranscribeText:` (`WKConversationInputPanel.m:1472`), has no caller anywhere in the manager. So the captured `draft` really is the user's pre-voice text and is not confused with the caption.

The problem: this change restores only the **text**, not the **mention state** that text depends on.

## What is correct (verified at head 1aa26b03, static analysis only — repo has no XCTest target)

- **Selector rename has no missed callers.** `_commitPendingWithCaption` appears only at `:687` / `:1535` / `:1551`, all updated to the two-arg form; the references in `WKMoreItemClickEvent.{h,m}` are comments only, no `performSelector`/string-selector dispatch. Method is private (no header decl), so the signature change is build-safe.
- **Manual send path is behavior-equivalent.** `:687` passes `nil`; `restoreText ?: @""` (`:1242`) is identical to base's `self.textView.text = @""`. Zero regression.
- **Draft persistence and input-height/send-button state recover correctly.** `saveDraftOrKeepPosition` reads `[self.input inputText]`; voice→keyboard `toggleVoiceMode` recomputes `sendButton.show/hidden` and calls `resetCurrentInputHeight`; `WKGrowingTextView setText:` re-dispatches `refreshHeight:`. No "has text but no send button / collapsed height" state.
- **Async failure/merge path preserves text content.** On `onFailure`, `currentText = textView.text` is now the restored draft → merge branch prepends the caption with a HUD notice; content is not lost.

## Blocking finding

### P1 — Restored draft loses its @mention state; the mention silently stops working (`WKConversationInputPanel.m:1242` restores text, `:1258` wipes the cache)

The aggregate branch restores the textView to `restoreText` at `:1242`, but the same branch unconditionally executes `[ctx cleanMentionCache]` at `:1258`. The `mentionCache` is the **sole** source of mention entities:

- `entities:` / `mentionedInfo:` (`WKConversationContextImpl.m:323`, `:397`) iterate `mentionCache.items` and match `@name` against the outgoing text.
- `sendTextMessage:` computes entities from the cache and then calls `[self.mentionCache clean]` (`:505`, `:509`).
- The only method that could reverse-derive an item from text, `matchString:` (`WKInputMentionCache.m:106`), has **no callers** in the repo.

Mention items are written into the cache when the user taps a person (`WKConversationContextImpl.m:1074` → `addMentionToCache:`). Neither `toggleVoiceMode` nor `hideMentionUsers` cleans the cache, so at the moment of STT commit the draft's mention item is still in the cache — and then `:1258` wipes it alongside the caption's own items.

**Reachable consequence:** draft `@Alice look at this` + a pending image → hold-to-talk sends an STT caption → the image+text message goes out and the draft text is restored → the user switches back to the keyboard and sends that restored draft. The message shows a visible `@Alice`, but `entities` is empty and `mentionedInfo.uids` is empty — Alice is not notified and the client does no mention highlight. `@所有人` / `@所有AI` is worse: the `humans=1` / `ais=1` flags are lost, so the broadcast and the robot-uid expansion at `WKConversationContextImpl.m:471` never run — all-broadcast and bot routing silently fail. The `onFailure` merge branch (`:1284`) rewrites caption + draft into the field when the cache is already empty, so it has the same defect.

**Newly-exposed by this change (verified against `pr94^`):** on base, this same operation cleared the draft text (`self.textView.text = @""`), so a "dead-@" message could not be sent at all. It is precisely `:1242` keeping this text alive and sendable that newly exposes the mention loss. Note `:1258`'s blanket clean was added deliberately (commit `c6547b93`) to prevent `@all` leakage; on the manual path (`restoreText == nil`) caption == full textView content and the blanket clean is still correct there — do not revert it wholesale.

**Suggested fix direction:** narrow the cleanup to the items the caption actually consumed. Snapshot `mentionCache.items` before the clean and re-add (via `addMentionItems:`) the items whose `@name` still appears in `restoreText`; or use `WKInputMentionCache.m:90` `removeName:` to drop only the names matched in `captionRaw`, keeping the existing blanket clean when `restoreText == nil` (manual path). Apply the same preservation on the `onFailure` merge path. (Note the `@所有人`/`@所有AI` sentinel items carry localized names `LLang(@"所有人")` / `LLang(@"所有AI")`, so the "re-add if `@name` still in restoreText" rule covers them too.)

## Things checked that are fine

- Sibling voice path `holdToTalkManager:sendVoiceData:` (`:1475`) never touches `textView.text`; no symmetry gap.
- `restoreText` shape boundaries are safe (`nil` → `?: @""`; pure-whitespace drafts don't falsely show the send button because `handleTextViewContentDidChange` and `toggleVoiceMode` both trim; oversized draft still hits the `messageTextMaxBytes` prompt on the eventual send at `:690`).
- `hasPendingImages` and the `images.count == 0` guard at `:1234` share the same data source, so no new silent-swallow path for the STT text.
- No automated test (checklist acknowledges this); the fix is a UIKit interaction that's awkward to unit test, but the mention-state case above should be exercised manually before merge.

## Verdict

The fix's direction is right, caller coverage is complete, and the manual-send path is a zero regression. The single blocker: the text is restored but the mention cache backing it was already cleared at `:1258`, so sending that restored draft produces a message with a visible-but-dead `@`, and `@所有人`/`@所有AI` lose broadcast + bot routing. Narrow the cleanup to the caption-consumed items (and handle the `onFailure` merge path the same way) to unblock.

Jerry-Xin (approved):
**Verdict: APPROVE** (head `295d54f58798`, verified against merge-base `f5e7a010`; supersedes my prior approval of `1aa26b03`)

Remediation round. The delta `1aa26b03..295d54f5` is exactly the four-file mention-restore step (conditional cleanup branch + new `removeMentionItemsMatchingText:` protocol method + impl + `matchString:` header exposure), classified hunk-by-hunk; in addition, the PR's full own bytes vs merge-base were re-walked with fresh eyes, not just the delta.

## Under-call ownership and severity re-calibration

My previous round verified the mention-loss mechanism as true but filed it as a non-blocking advisory ("A-1: net improvement + inherited family gap"). That severity call was wrong and is publicly corrected here.

- The inherited-gap attribution was byte-true but scoped to the wrong flow. At merge-base the no-pending-images voice path (`inputPanelSend:` at `WKConversationView.m:718` → `sendTextMessage` → unconditional `mentionCache clean` at `WKConversationContextImpl.m:509`) did leave an untouched textView draft in the visible-but-dead-@ state — verified on merge-base bytes. But on the tray flow this PR fixes, merge-base destroyed the draft entirely, so a sendable dead-@ draft was NOT reachable there. This PR's restore is what newly creates that state in exactly the flow it touches.
- "Net improvement" does not excuse shipping a state where a visible @ sends with no notification — and where the all-member / all-AI sentinels silently lose broadcast and bot routing — when the coherent fix is this cheap.
- Credit where due: mochashanyao (review 5195157402) and yujiawei (review 5195353158, after self-flipping their earlier approval) independently escalated this same mechanism to blocking and prescribed the exact fix direction now implemented. Two independent blocking escalations against my advisory classification was the correct outcome; my calibration under-weighted a user-visible silent-failure state in the PR's own shipped flow.

## Adjudication of the peer findings at this head

1. **mochashanyao blocking / yujiawei blocking #1 — restored draft loses its mention backing → FIXED.** The unconditional `[ctx cleanMentionCache]` is now branched (`WKConversationInputPanel.m:1259-1266`): when `restoreText.length > 0` (voice path), only items whose `@name` actually occurs in `captionRaw` are removed via the new `removeMentionItemsMatchingText:` (`WKConversationContextImpl.m:1164-1174`, composing the pre-existing `matchString:` + `removeName:` — no new matching logic); draft-backed items survive and back the restored draft on its eventual manual send (entities + `mentionedInfo` incl. `humans=1` / `ais=1` + robot-uid expansion). When `restoreText == nil` (manual path) the original blanket clean is retained verbatim → zero regression to the deliberate "@all" leak guard (commit c6547b93) where its premise holds. This is precisely the fix both reviewers prescribed.
   - Extraction reliability: `onResultSendText` always runs `parseMentionMarkers` before delegating (`WKHoldToTalkManager.m:1170-1186`), normalizing every parsed mention to the `@Name` + U+2004 terminator form, and `matchString:`'s pattern `@([^U+2004]+)U+2004` extracts exactly those names. A caption with no parsed mentions yields no matches → nothing removed → draft items untouched.
   - Sentinel coverage (yujiawei's note): the all-member / all-AI sentinel items carry the same display name in the item and in the normalized caption text within each producer path (`addMentionToCache:` LLang names for the keyboard path; `parseMentionMarkers` LLang/hardcoded names for STT), so `removeName:` by display name hits them; draft-owned sentinels survive and still fire broadcast/routing on later sends (`allMentionUid:` text-gates them by `@name` presence).
2. **yujiawei blocking #2 — cleanMentionCache timing wipes mention data at restore → FIXED.** The timing claim was byte-true at the old head (restore at `:1242`, blanket clean later in the same synchronous branch at `:1258`). At this head the wipe on the voice path is replaced by the caption-scoped removal, making the ordering moot. Full census of cache cleaners at this head: `cleanMentionCache` is invoked only from the manual aggregate branch (`WKConversationInputPanel.m:1265`); `sendTextMessage`'s internal clean (`Impl.m:509`) is unchanged; no other cleaner exists. Restore/cleanup ordering verified on every path: STT success (synchronous branch, scoped removal), STT/upload failure (`onFailure` never touches the cache — draft items survive the failure), overlay cancel (`onResultCancel` touches neither textView nor cache), mode switch (`toggleVoiceMode` has no cache interaction, re-verified at this head).
3. **Overlap and duplicate edges resolve benignly.** `addMentionItems:` appends without dedupe (`Impl.m:1082-1084`), so a caption mention that also appears in the draft means two identical items at commit time; `removeName:` removes one and the identical survivor keeps the draft's `@name` backed. Multiple mentions, repeated occurrences and duplicate display names all resolve the same way. Caption and draft are disjoint strings (they merge only in `onFailure` with a `"\n\n"` separator), so no mention can span the caption boundary. A whitespace-only caption routes to the pure-image branch, which never touches the cache.
4. **Protocol surface is build-safe.** `removeMentionItemsMatchingText:` is added under the header's `@optional` section (no `@required` in `WKConversationContext.h`), the sole conformer `WKConversationContextImpl` implements it, and the call site is `respondsToSelector:`-guarded. The CI `build` job (real compile: pod install + xcodebuild with the iOS SDK >= 26 assertion) is green at this head.
5. **Test requests (both reviewers, non-blocking) → NOT ADDRESSED, stated plainly.** No test file is added at this head; the project has no XCTest target and CI is compile-only (`build-for-testing` with no test-execution step), so there is no coverage illusion — nothing was added that CI would silently skip. The ask remains open and cannot be met inside this PR's scope without standing up test infrastructure. The checklist honestly leaves tests unchecked and the PR body documents a manual test plan (restore case + manual-send regression case); the manual all-member sentinel case should be exercised before merge.

## 🟡 Advisories (non-blocking)

- **A-1 — Failure-path backfill is declared but unwired; mandatory fast-follow.** `removeMentionItemsMatchingText:` returns the removed items and the new header doc says the return value exists for backfill on send failure (`WKConversationContext.h:155-157`), but the call site discards it (`WKConversationInputPanel.m:1262`). On upload failure, `onFailure` merges `captionRaw` back into the textView while the caption's mention items are already removed → the caption-side `@name`s (including sentinels) in the merged text are unbacked if re-sent. Classification, with byte evidence: inherited, NOT TRUE-NEW — at merge-base the identical sequence (blanket clean → `onFailure` writes `captionRaw` back into the cleared slot) left the caption-side @ exactly as dead, and the `onFailure` block itself is untouched by this PR (the delta changes only a comment line above it). At this head the state is strictly improved: the draft-side @s now survive the failure, which is the preservation both reviewers asked the failure path to have. The fix is three lines — capture the returned array and re-add via `addMentionItems:` inside the failure block (re-added items are text-gated at send time, so no leak risk) — and would make the implementation match the API contract this PR itself introduces.
- **A-2 —** The `name=="all"` no-text-check branch in `allMentionUid:` has no producer anywhere in the repo at this head (sentinel items carry display names with uid `all`), so narrowing the cleanup does not reopen the leak the blanket clean guards against; real sentinels are text-gated. Noted so future edits do not reintroduce a `name=="all"` producer without restoring a guard.
- **A-3 —** Merge awareness only: the `check-sprint` status is red at this head — the org-level project-metadata gate (no `Closes #<issue>` reference), non-code; `build` and all scan jobs are green.

## Note on the live CHANGES_REQUESTED reviews

mochashanyao's and yujiawei's change requests (both pinned to `1aa26b03`) remain live and are not mine to dismiss. I adjudicate their blocking findings FIXED at this head with the byte evidence above; whether to re-review is theirs to grant. If either considers the A-1 caption-side failure path part of the blocking scope rather than the inherited family I classified it as, the three-line wiring closes it immediately.

## Evidence base

Byte-level verification against merge-base `f5e7a010` and prior head `1aa26b03` (delta isolated via git diff; merge-base behavior of `sendTextMessage` / `inputPanelSend:` / `onFailure` confirmed via git show); repo-wide censuses (cache cleaners, mention-item producers, call sites, protocol conformers); head-SHA CI compile green. No simulator run; runtime behavior claims are code-evidence-based.

yujiawei (approved):
# Code Review — PR #94 (octo-ios)

**Head SHA reviewed:** `295d54f58798648f9b55da461ffc9725e478c021`

Independent review, including a cross-model adversarial pass. The fix is correct and well-scoped. No blocking issues. Two non-blocking polish items below.

## 1. Spec compliance

**Spec: ✅**

The stated bug: user types a draft, selects images (pending tray), then sends via voice — the STT text + images go out, but the typed draft in the text view is silently cleared and never saved. The fix threads a `restoreText:` parameter through `_commitPendingWithCaption:`:

- Manual send path (`inputSendFinished`, L687) passes `nil` — behavior unchanged (`nil ?: @""` still hard-clears, which is correct since the caption *is* the whole textView content).
- Both hold-to-talk voice paths (`sendText:` L1541, `sendText:mentions:` L1557) snapshot `self.textView.text` and pass it as `restoreText` before committing.
- The clear becomes `self.textView.text = restoreText ?: @""` (L1240), restoring the draft on the voice path instead of wiping it.

No missing scope, no extra features, no deviation. The `removeMentionItemsMatchingText:` / `matchString:` additions are exactly what's needed to avoid over-clearing the mention cache when a draft is restored.

## 2. Code quality

**Quality: Approved**

Data-flow trace confirms the mention handling is safe. The message payload is built strictly from `captionRaw` (the STT text): `entities:` matches each cached item by `@<name>` range within the passed text, and `mentionedInfo:` → `allMentionUid:` gates each item by `containsString:`. Draft-only `@`-items that remain in the cache therefore produce **no** entity and **no** mentionedInfo for the sent message — they cannot leak into the outgoing STT message. The new `restoreText.length > 0` branch then removes only the mention items the STT caption actually consumed and leaves the draft's `@`-items in the cache so the restored draft, if later sent from the keyboard, keeps its mentions. This is the right call.

Build safety is fine: `removeMentionItemsMatchingText:` is declared under the protocol's `@optional` section (`WKConversationContext.h:157`, `@optional` at L27, no intervening `@required`), and the call site guards with `respondsToSelector:`. No conformer is forced to implement it.

### P2 (non-blocking) — send button / input height not refreshed after programmatic restore

`WKConversationInputPanel.m:1240` sets `self.textView.text = restoreText` directly, then forces `sendButton.show = NO; sendButton.hidden = YES;` and `resetInputHeight` (single line). Assigning `.text` programmatically does **not** fire `textViewDidChange:`, so `handleTextViewContentDidChange` never runs for the restored content. Every other text-setter in this class (`inputSetText:` L1010, `inputDeleteText:`) calls `handleTextViewContentDidChange` immediately afterward for exactly this reason.

In practice this is masked: the voice paths only fire in voice mode, where the text view is hidden, and `toggleVoiceMode` (L1453) recomputes button visibility and height when the user switches back to the keyboard. So the wrong state is not visible today. It is still a latent trap — if a future change surfaces the text view without going through `toggleVoiceMode`, a non-empty draft would render with the send button hidden and the height collapsed. Suggest routing the restore through `inputSetText:` (or calling `handleTextViewContentDidChange` after the assignment) so the visible state always matches the text.

### P2 (non-blocking) — removed mention items are discarded

`removeMentionItemsMatchingText:` is documented as returning the removed items "for failure restoration," but the caller at L1262 discards the return value. On the async `onFailure` path, the STT caption is merged back into the (already-restored) draft, but the mention metadata for those STT `@`-items is gone from the cache, so a re-send of the merged text would drop those mentions' entities. This is a narrow, best-effort recovery path (upload failure) and not a regression from this PR, but if you want the failure-merge to be lossless, capture the returned items and re-add them when merging the caption back.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. The two P2s are polish, not merge blockers.

## 4. Suggestions

- Route the draft restore through `inputSetText:` (or call `handleTextViewContentDidChange` after the direct `.text` assignment) to keep the visible send-button/height state consistent, and to avoid a latent trap for future callers.
- If you want the async-failure merge to preserve STT mentions, capture the array returned by `removeMentionItemsMatchingText:` and re-add it in the `onFailure` merge block.
- Consider a lightweight test for the reported repro (type → pick image → voice-send → assert draft restored) since the PR checklist leaves tests unchecked.

## 5. Additional notes

While tracing, I confirmed the `[item.name isEqualToString:@"all"]` branch in `allMentionUid:` (WKInputMentionCache.m) is effectively dead — no code sets a mention item's `name` to the literal `"all"` (the @所有人 item carries `uid=="all"` but `name==所有人`/`Everyone`). Not introduced by this PR and not a concern for this change, but worth a cleanup pass someday so the special-case doesn't mislead future readers into thinking a draft `@all` could bypass the text-scoping.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #94 (octo-ios)

## Summary

This PR fixes a draft-loss bug in the voice-input + pending-image-tray flow. `_commitPendingWithCaption:` used to hard-clear the input box after every send, which is correct for the manual send key (the caption *is* the whole text view) but wrong for the hold-to-talk path, where the caption is STT text and the user's typed draft is a separate thing. The fix adds a `restoreText:` parameter — `nil` from `inputSendFinished`, a pre-send snapshot of `textView.text` from both voice callbacks — and restores it instead of clearing. The second commit closes the follow-up hole that created: the aggregate branch's unconditional `cleanMentionCache` would have wiped the mention items the restored draft depends on, so it now branches on `restoreText.length` and calls the new `removeMentionItemsMatchingText:` to evict only the items the caption actually consumed.

The change is small, well-targeted and the core logic holds up under tracing: the restored text really does reach the draft persistence path, the outgoing message still gets its own entities, and the voice caption's mention items are always removable by the new text-scoped matcher. No P0/P1 issues; four P2 items and one nit below, all non-blocking.

## Verification

Static analysis only at head `295d54f5`; build and tests not executed in this environment. Reviewed against merge-base `f5e7a010` with `main`.

- ✅ **Manual send path is byte-identical** — `Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKConversationInputPanel.m:687` passes `nil`, and `restoreText ?: @""` at `:1242` reproduces the old hard clear, including the `cleanMentionCache` branch at `:1264`.
- ✅ **All call sites migrated** — `_commitPendingWithCaption:restoreText:` has exactly three callers (`:687`, `:1543`, `:1559`); no stale single-argument selector remains anywhere in the repo.
- ✅ **Restored text reaches the draft store** — `inputText` returns `textView.text` (`:1040`), `saveDraftOrKeepPosition` reads it on `viewWillDisappear` (`WKConversationView.m:168`, `:356`), and `recoveryDraft` writes it back on load (`:86`, `:508`). The consumer is live, not a dead assignment.
- ✅ **Restored draft cannot leak into the outgoing message** — `sendRichTextMixedImageDatas:` only ever reads its `extraText:` argument (`WKApp.m:2661`), and the aggregation gate is computed from `captionRaw`, not from the text view (`WKApp.m:2633`).
- ✅ **The snapshot really is the user's draft** — `holdToTalkManager:didTranscribeText:` (the only callback that would insert STT into the input box, `:1479`) is never invoked by `WKHoldToTalkManager`; recognized text stays in the overlay's own `resultTextView`. So `textView.text` at `:1542`/`:1558` contains typed draft only, with no voice text to duplicate.
- ✅ **Message mentions still built before eviction** — `entities:`/`mentionedInfo:` at `:1251-1252` run before the removal branch at `:1260`, so the sent RichText keeps its own `@` payload.
- ✅ **Voice sentinels don't survive the send** — `parseMentionMarkers:` appends an item and its canonical `@name\u2004` marker as a pair, including the `@All AIs` → `@所有AI` rewrite (`WKHoldToTalkManager.m:1320-1435`), so `matchString:` always finds exactly the items the caption added — `all` and `__ais__` included.
- ✅ **Input-bar UI self-corrects after restore** — the unconditional `sendButton.hidden = YES` + `resetInputHeight` at `:1243-1246` is undone by `pendingImageBar clear` → `onContentSizeChanged` → `handleTextViewContentDidChange` (`:1119`), by `WKGrowingTextView setText:`'s async `refreshHeight:` → `growingTextView:willChangeHeight:` (`:973`), and again by `toggleVoiceMode`'s keyboard branch (`:1464`). During a voice send the text view is hidden anyway, so there is no visible clipped-draft state.
- ✅ **No mutation-during-enumeration** — `removeName:` collects the victim inside the enumeration and removes it afterwards (`WKInputMentionCache.m:90`), so the new loop at `WKConversationContextImpl.m:1167-1172` is safe; `text ?: @""` covers a nil argument.
- ✅ **Protocol change is inert for conformers** — the new method lands inside the `@optional` section (`WKConversationContext.h:27`) and the only conformer, `WKConversationContextImpl`, implements it, so the `respondsToSelector:` guard at `:1261` always takes the new branch.
- ✅ **Failure path loses nothing** — `onFailure` reads the already-restored draft (`:1291`), merges caption + draft (`:1297`), restores the images and shows the merge HUD (`:1308`).
- ✅ **Sibling voice surfaces need no matching change** — the no-pending-image voice path forwards to `inputPanelSend:text:` (`:1546`, `:1563`) which never clears the box, and `WKVoicePanel.m:98` sends through `sendTextMessage:` without touching the input.

## Findings

No P0/P1 issues; four P2 items and one nit below.

### P2 — Removed mention items are never backfilled on send failure (`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKConversationInputPanel.m:1262`)

`removeMentionItemsMatchingText:` returns the evicted items and the header contract documents that return value as existing for failure backfill (`WKConversationContext.h:155`), but this sole call site discards it. When `sendRichTextMixedImageDatas:` fails, `onFailure` merges `captionRaw` back into the text view at `:1297`, so the recovered draft visibly contains the caption's `@name\u2004` markers while the cache no longer holds those items — re-sending that merged draft yields no mention entity and no `mentionedInfo`, i.e. a silent `@` failure. The same net loss existed on base (the unconditional `cleanMentionCache`), so this is a pre-existing gap rather than a regression, but the PR adds the API that was meant to close it and leaves it unwired.

```objc
// sketch
NSArray<WKInputMentionItem *> *consumed = [ctx removeMentionItemsMatchingText:captionRaw];
// ... capture `consumed` in the onFailure block and re-add before the text merge:
if (consumed.count > 0 && [self_ ctxRespondsToAddMentionItems]) { [ctx_ addMentionItems:consumed]; }
```

Fix direction: keep the returned array and re-add it via `addMentionItems:` inside the `onFailure` block so the merged draft's `@` stays functional.

### P2 — Eviction is keyed by display name rather than item identity (`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/WKConversationContextImpl.m:1168`)

`matchString:` → `removeName:` both resolve victims by `item.name`, and `addMentionItem:` appends without dedup (`WKInputMentionCache.m:70`). In the common collision case this self-heals: the voice path has just added its own duplicate via `addMentionItems:` (`:1553`), so the removal consumes that copy and the draft's item survives with the right uid. It breaks when two different channel members share a display name — `parseMentionMarkers:` builds its match table from both `memberRemark` and `memberName` (`WKHoldToTalkManager.m:1330-1345`) — because `removeName:` evicts the *first* name match, which can be the draft's item, leaving the other member's uid behind. Sending the restored draft later then binds that `@` to the wrong person. Narrow, and the cache has always been name-keyed (`entities:` resolves by name too), so not something I'd block on. Fix direction: pass the exact `mentions` instances the voice callback added down into `_commitPendingWithCaption:restoreText:` and remove those by identity instead of re-matching text.

### P2 — The new selective-removal contract has no regression test (`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/WKConversationContextImpl.m:1164`)

The behaviour this PR depends on is pure and cheap to cover, and this module already carries an active suite — `WKRichTextSendTests.m` locks the aggregation gate that decides this branch, `WKMentionUserCellTests.m` covers mention cells. Untested semantics that matter: names present in the text are removed *and returned*; names absent from the text (the retained draft's `@`) stay in the cache; a name repeated in the text removes one item per occurrence; nil text is safe. The `restoreText.length > 0` branch selection at `:1260` (including the whitespace-only draft case, which skips the full clean) is equally untested. Fix direction: add a cache/impl-level test for `removeMentionItemsMatchingText:` plus one branch-selection case.

### P2 — Selective removal has no fail-closed fallback (`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/Input/WKConversationInputPanel.m:1261`)

When `restoreText.length > 0` the branch only *attempts* `removeMentionItemsMatchingText:` and does nothing at all if the selector is unavailable, so a context that does not implement it would keep the caption's just-added mention items — including the `all` / `__ais__` sentinels — in the cache and leak them into the next message, which is precisely the hazard the neighbouring comment at `:1253-1259` warns about. This is unreachable today (`WKConversationContextImpl` is the protocol's only conformer and implements the method, so the guard always passes), so it is hardening rather than a live defect. Fix direction: add an `else` that falls back to `cleanMentionCache` so the cleanup fails closed and can never be silently skipped by a future conformer or a test double.

### Nit — Mention-token regex now exists in two public places (`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/WKInputMentionCache.h:35`)

Exposing `matchString:` is fine, but `WKStringUtil.m:16` (`+matchMention:`) already carries a byte-identical copy of the `@([^\u2004]+)\u2004` pattern and the same first/last-char slicing. Two copies of the mention grammar can drift if `WKInputAtEndChar` ever changes. Consider routing one through the other.

- **No false `@all` from retained items.** A whitespace-only draft makes `restoreText.length > 0` and skips the full clean, leaving caption-unrelated items in the cache. Every consumer re-gates on text containment (`WKInputMentionCache.m:32-56`), and the one unconditional branch (`item.name == "all"`) is unreachable because every producer sets `name` to `LLang(@"所有人")` or a member name (`WKConversationContextImpl.m:1118`, `WKHoldToTalkManager.m:1369-1411`). Retained items are inert until the text actually contains the marker.
- **Case folding.** `matchString:` is case-insensitive while `removeName:` compares exactly, but the caption is always rewritten to canonical member names by `parseMentionMarkers:` (even when the user edits the recognized text in the overlay), so the mismatch is not reachable on this path.
- **Re-entrancy.** All bail-outs (`inFlightPendingSend`, `images.count == 0`, `!ctx`) happen before the text mutation at `:1242`, so no path half-restores or drops the draft.
- **Repeat voice sends** re-snapshot the same draft and restore it again — no accumulation, no duplication.
- **Prior-round blocker is genuinely fixed.** The unconditional `cleanMentionCache` that wiped the restored draft's mentions is gone; the aggregate branch now splits on `restoreText.length` (`:1260-1266`) with the manual path still fully cleaned, and the new impl reuses the existing `matchString:`/`removeName:` primitives rather than adding a second matcher.
- **`matchString:` implementation confirmed present.** The header change at `WKInputMentionCache.h:35` only makes an existing method callable: the implementation is pre-existing at `WKInputMentionCache.m:106` (outside this diff) — pattern `@([^\u2004]+)\u2004` with first/last-char slicing down to the inner name, which is exactly what `removeName:` compares against. Cross-checked every caption producer (`WKHoldToTalkManager.m:1320-1435`): each item it adds is emitted into the text verbatim as `@name\u2004`, so the matcher cannot silently miss a caption mention.
- **Cross-session caveat, pre-existing.** `saveDraftOrKeepPosition` persists text only (`extra.draft`), so an `@` inside any draft is inert after leaving and re-entering the conversation. That is true for all drafts on base and is not introduced or amplified here — worth a separate issue if product cares.

## Verdict: APPROVED

The fix does what it claims and the data flow holds end to end: the typed draft is snapshotted before the text view is touched, restored instead of cleared, and picked up by the existing draft-save path, while the outgoing message keeps its own mention payload and the caption's items are evicted selectively. The four P2s are all about robustness around the new eviction API — failure-path backfill, identity-based removal, test coverage, and a fail-closed fallback for the new selector — and none of them makes a working path unusable or produces wrong data on a reachable common path, so they are not merge blockers.





---

## 数据流回溯（每个被消费数据 → 上游来源 → 是否真流到消费点）

| 被消费数据 | 上游来源 | 是否真流到消费点 |
|---|---|---|
| `restoreText`（`:1242` 写回 textView） | `self.textView.text` 于 `:1542` / `:1558` 快照（手动路径 `:687` 传 nil） | ✅ 流到。语音路径下 textView 未被 STT 污染（`didTranscribeText:` 从未被 manager 调用），快照=用户真实草稿 |
| 写回的草稿 → 持久化 | `inputText`(`:1040`) → `saveDraftOrKeepPosition`(`WKConversationView.m:356`，`viewWillDisappear` `:168` 调用) → `extra.draft` → `updateOrAddExtra` | ✅ 流到。回读 `recoveryDraft`(`:508`) → `inputSetText:` → `setText:` → `internalTextView` |
| `captionRaw` → 发出的消息 | 语音回调入参 `text`，由 `parseMentionMarkers:`(`WKHoldToTalkManager.m:1320`) 生成 | ✅ 流到 `extraText:`(`WKApp.m:2661`)；该函数不再读输入框，草稿不会混入消息 |
| `entities` / `mentionedInfo`(`:1251-1252`) | `ctx.entities:` / `ctx.mentionedInfo:` ← `mentionCache.items` + 文本包含匹配 | ✅ 在移除动作(`:1260`)之前求值，本条消息的 @ 完整 |
| `removeMentionItemsMatchingText:` 的移除目标 | `matchString:(captionRaw)` → `removeName:` | ✅ 语音 caption 中每个 mention 都伴随 canonical `@name\u2004` 被写入文本（含 `@All AIs`→`@所有AI` 改写），故 caption 自己加入的 item 必被命中移除；`all`/`__ais__` sentinel 不残留 |
| 被保留的草稿 mention item → 后续发送 | `mentionCache` 内残留 item | ✅ 可用。后续 `sendTextMessage:` 走 `entities:`/`allMentionUid:`，均按文本包含 `@name` 命中；文本已删则不命中，不会误 @ |
| `removeMentionItemsMatchingText:` 的返回值（头文件声明"用于发送失败时回填"） | 新 impl 返回 `removed` 数组 | ❌ **未流到任何消费点**：唯一调用点 `:1262` 丢弃返回值，失败合并(`:1297`)后文本里有 @、缓存里没有 item → 见 P2#1 |
| 失败路径 `currentText`(`:1291`) | 已被写回的草稿 | ✅ 流到合并分支，caption + 草稿都保留，图片经 `setImageDatas:` 回填，HUD 提示合并 |
| UI 状态（sendButton / 输入栏高度） | `:1243-1246` 硬置隐藏+最小高度 | ✅ 被三处自愈：`pendingImageBar clear`→`onContentSizeChanged`→`handleTextViewContentDidChange`(`:1119`)、`setText:` 内 async `refreshHeight:`→`growingTextView:willChangeHeight:`(`:973`)、`toggleVoiceMode` 键盘分支(`:1464`)。语音发送时 textView 本就 hidden，无可见裁剪态 |

## 盲点 checklist（C1–C6）

- **C1 双路径 parity — clear（附证据）**：`_commitPendingWithCaption:restoreText:` 全部 3 个调用点已迁移（`:687` nil / `:1543` draft / `:1559` draft），repo 内无旧单参 selector 残留。两个语音回调对称，均在调用前快照。姊妹语音面：无 pending 图路径只转发 `inputPanelSend:text:`（`:1546`/`:1563`，不清空输入框），`WKVoicePanel.m:98` 走 `sendTextMessage:`（不动输入框），无需同款修复。add↔remove 对称性：`addMentionItems:`(`:1553`) 与 `removeMentionItemsMatchingText:`(`:1262`) 成对，但**移除按 name 而非 identity** → P2#2；**返回值未回填** → P2#1。守卫自身覆盖：`respondsToSelector:`(`:1261`) 的唯一 conformer `WKConversationContextImpl` 已实现（协议 `@optional` 段，`WKConversationContext.h:27`），不存在守卫盲区导致静默跳过清理。
- **C2 control-flow ordering / 嵌套复用 — clear**：`entities:`/`mentionedInfo:` 求值严格早于移除（`:1251-1252` → `:1260`），顺序正确。复用点排查：同一方法被手动/语音两类 ordering 复用，手动路径 `restoreText=nil` → 走原 `cleanMentionCache`，零回归。所有早退（`inFlightPendingSend`、`images.count==0`、`!ctx`）都在文本写入之前，不存在半还原。非 canonical 形式试穿：大小写（`matchString:` 不敏感 vs `removeName:` 敏感）→ caption 恒为 canonical 名，不可达；unicode 结束符 `\u2004` 由生成端强制写入；空白-only 草稿（`length>0` 但无 @）→ 跳过整体清，但所有消费端二次做文本包含校验，残留 item 惰性无害；nil（`text ?: @""`）已防。枚举期变异：`removeName:` 先取后删（`WKInputMentionCache.m:90`），无 crash。
- **C3 授权边界 ≠ 能力边界 — N/A**：本 PR 不涉及权限/jail/tool 暴露/凭证。mention 广播（`@所有人`/`@所有AI`）属于通知面而非授权面，已按 C2/R3 校验 sentinel 不残留。
- **C4 授权生命周期 / 容器-成员状态级联 — N/A**：无鉴权、无成员/容器 status 查询改动。
- **C5 build/note 通过 ≠ 运行期路径正确 — clear（未依赖 build）**：本环境未编译、未跑测试（iOS/Xcode 不可用），结论全部来自运行期推演而非"能编过"。推演覆盖：写回文本→草稿持久化消费点、caption→发送 payload、移除→缓存后续消费、UI 状态三处自愈、失败回调合并。注释里的"无需改动失败合并逻辑"声明未被采信，已实读 `:1291-1310` 验证（并据此产出 P2#1）。
- **C6 治理/策略/安全文档自洽性 — N/A**：无文档/治理/策略文件改动（4 个文件均为 Objective-C 源码）。
- **R3 分布式/多实例副作用 — clear**：改动只在客户端单实例输入面板内。fan-out：`entities:`/`mentionedInfo:` 只作用于本条消息；未新增 emit/推送/计数。observer≠actor：无归因字段改动。无界面触发（scheduled/后台）：语音发送必在主线程（overlay 按钮），缓存变更同步完成。唯一对外副作用面 = mention 广播，已验证 caption 消费的 `all`/`__ais__` item 必被移除、草稿 item 因文本包含门控不会误触发广播。
- **安全 pass（STRIDE/OWASP）— 无发现**：Tampering/Repudiation — 发送内容与用户可见草稿一致性已核（消息只含 caption，草稿不回灌）；Information disclosure — 无日志/URL/token/PII 新增输出，草稿仅落本地 `extra.draft`（既有机制）；DoS — 无新增无界循环/分配，`matchString:` 正则作用于单条 caption；EoP — 无授权面改动。

## 跨轮 blocker 复检（R6）

本轮为该 PR 第 2 轮（head `295d54f5`，前一轮 head `1aa26b03`）。上一轮 blocker 依据本仓 git 提交说明 + PR 描述（未查询 GitHub 历史评论）：

| 上轮 blocker | 状态 | 证据 |
|---|---|---|
| 图文聚合分支无条件 `cleanMentionCache` 会把还原草稿依赖的 @ 缓存一并清空 → 草稿发出时"看得见 @ 但 entities/mentionedInfo 为空"，`@所有人`/`@所有AI` 广播与 bot 路由丢失 | **已修** | `:1260-1266` 按 `restoreText.length` 分流：非空走 `removeMentionItemsMatchingText:captionRaw`（只移除 caption 实际消费的项），为空仍走 `cleanMentionCache`（手动路径零回归）。新 impl `WKConversationContextImpl.m:1164-1175` 复用既有 `matchString:`(`WKInputMentionCache.m:106`)+`removeName:`(`:90`)，未新写匹配逻辑。`entities:`/`mentionedInfo:` 仍在移除之前求值(`:1251-1252`)，本条消息 @ 不受影响；草稿 @ 因 item 保留而在后续 `sendTextMessage:` 中正常生成 entity |

无未解决的上轮 P0/P1 残留。本轮新增 4 个 P2 + 1 个 Nit，均非阻塞（其中「新 selector 无 fail-closed 兜底」当前不可达，属加固建议；「matchString: 实现是否在」已用 `WKInputMentionCache.m:106` 既有实现证伪并转入已核验清单）。轮次 < 4，无需升级整体方案重审。

[Octo-Q] verdict: APPROVE — 核心修复的数据流已逐点回溯验证通过（草稿快照→写回→持久化消费点、caption→发送 payload、mention 选择性移除、失败合并、UI 状态自愈），上一轮 blocker 已真实修复且手动发送路径零回归；剩余 4 个 P2（失败路径未回填被移除的 mention item、移除按 name 而非 identity 在同名成员下可能驱逐草稿项、新缓存契约无回归测试、新 selector 无 fail-closed 兜底[当前不可达，加固]）与 1 个 Nit 均不满足 R1「让能工作的路径不可用 / 产生用户可见错误数据 / 比修复前更糟」，按 R4 映射为 APPROVE。

PR #95, opened 2026-09-15
feat(conversation): 子区来自群聊点击跳转到父群聊

swr614:
## What

子区聊天页顶部"来自: XX群聊"这行文字，点击后改为跳转到该子区所属的父群聊，而不是弹出子区详情；普通群聊/私聊场景下副标题（在线状态/成员数）的点击行为保持不变。

## Why

之前无论点"来自:"还是点后面的群聊名称，触发的都是同一个动作——打开子区详情页。这不符合预期：用户看到"来自: XX群聊"，直觉是想跳转去看那个群聊本身，而不是打开当前子区的设置/详情。

## How

- `WKConversationChannelHeader` 给副标题标签单独加了一个点击手势，新增 `onSubtitleTap` 回调；没有设置这个回调时，副标题点击行为退回到原来的 `onInfo`。
- `WKConversationVC` 只在解析子区所属父群信息的地方（即确认当前是子区场景时）才去设置 `onSubtitleTap`，并把它接到已有的 `pushConversation:` 方法上跳转到父群。普通群聊/私聊场景下这个回调始终不会被设置，副标题点击会按原有设计打开会话详情，不受影响。
- 标题栏其他区域（头像、群名、更多按钮）点击行为不变，仍打开子区详情。

## Checklist

- [ ] All tests pass locally

## Testing

- 进入一个子区聊天页，点击"来自: XX群聊"文字，跳转到父群聊天界面
- 点击标题栏头像/群名区域，仍打开子区详情
- 普通群聊/私聊场景下，点击副标题（在线状态/成员数）仍打开会话详情，不再是无响应

## Breaking Changes

N/A


Jerry-Xin (changes requested):
**Verdict: REQUEST_CHANGES** — the thread-to-parent-group jump itself is wired correctly and follows the app's existing navigation pattern, but the gesture hookup silently breaks subtitle taps in every non-thread conversation page.

### 🔴 Blocking

**B-1. Subtitle taps in normal group / private chats become a dead zone (regression vs base).**
- `WKConversationChannelHeader.m:53` — `subtitleLbl` is a subview of `infoBoxBtn`. On base, the label is non-interactive (no `userInteractionEnabled`, no gesture), so taps on the subtitle ("N个成员" / online status) hit-test through to `infoBoxBtn` → `infoPressed` → `onInfo` → conversation settings.
- Head adds `userInteractionEnabled = YES` plus a `UITapGestureRecognizer` on the label (`WKConversationChannelHeader.m:63-64`), so those taps are now routed to `subtitlePressed` (`:90`) instead.
- `subtitlePressed` falls back to `onInfo` only when `onSubtitleTap` is nil — but `WKConversationVC.m:406` sets `onSubtitleTap` **unconditionally in the lazy `channelHeader` getter**, i.e. for every conversation type. For non-thread conversations `parentGroupChannel` is nil, so the block silently does nothing, and the header's fallback branch is unreachable from its only consumer.
- Net effect: in every normal group chat (member-count subtitle) and private/customer-service chat (online-status subtitle), tapping the subtitle now does nothing instead of opening conversation info. This contradicts the PR description's own claim ("普通群聊/私聊场景…行为不变，仍然走原来的 onInfo") and its manual-testing checklist item.
- Fix options: install `onSubtitleTap` only for thread conversations (e.g. assign it in `setupThreadHeader`, or gate on `channel.channelType == WK_COMMUNITY_TOPIC`), or have the block invoke the existing settings path when `parentGroupChannel` is nil. The first option restores the header's intended fallback semantics.

### 🟡 Advisories

1. **Jump when the parent group is deleted/left/not loaded.** `WKConversationVC.m:309` stores `parentGroupChannel` unconditionally — including when `getChannelInfo:` returns nil and the subtitle falls back to the generic "来自: 群聊" (`:315`). A tap then pushes a conversation for a group the user may no longer be able to open. This matches the existing global-search group jump behavior (`WKGlobalSearchVM.m:694` pushes a synthesized `WKChannel` the same way), so it is consistent and non-blocking — but consider a no-op or toast when the parent info is still unavailable after the fetch.
2. **Accessibility / discoverability.** The subtitle is now interactive but stays a plain `UILabel`: no accessibility trait or hint, and no visual affordance distinguishing the tappable "来自: XX群聊" (jump) from member-count/online-status subtitles (not a jump). VoiceOver users get no signal that the element is actionable.
3. **PR scope/description.** The PR carries a second commit (`ebb85c7`, `WKSafeFilePreviewVC.m`) fixing the preview window under iPad Split View / Slide Over / Stage Manager. That change is verified correct here — `scene.coordinateSpace.bounds` is used symmetrically in show and dismiss, and the non-scene fallback preserves prior behavior — but it is unrelated to the conversation feature and is not mentioned in the PR description; consider documenting it (or splitting it) to keep review/rollback scope clean.
4. **Tests.** No automated coverage for the new gesture routing (the repo has no unit-test target for these classes, and the checklist acknowledges this). If test infrastructure is added, cover: thread subtitle tap → parent-group push, and normal-chat subtitle tap → conversation settings (the B-1 regression).

### Verification notes
- Full walk of all 4 changed files (+37/−9) at head `e050ff2b`; diff confirmed against merge-base `5b948cbe`.
- Base-vs-head comparison of `subtitleLbl` interactivity (base: no interaction flag/gesture on the label) and full tap-path trace: label gesture → `subtitlePressed` → unconditionally-set `onSubtitleTap` in the `WKConversationVC` getter. Grep confirms `WKConversationVC` is the only consumer of the header class, so the regression surface is every chat page it renders.
- Traced `pushConversation:` through the `WKApp` endpoint registry (`WKPOINT_CATEGORY_CONVERSATION_SHOW` → pushes a `WKConversationVC` for the channel) and cross-checked against the existing group-jump pattern in `WKGlobalSearchVM`.
- Confirmed the `{groupNo}____{shortId}` thread-channel format is the established convention across conversation list, search, forward, and summary code, and is group-origin only — there is no private-chat-origin thread variant that could misroute the jump. Threads whose channelId lacks the separator leave `parentGroupChannel` nil (no-op), which is safe.
- Memory management of both new blocks checked (`weakSelf` capture; header holds copied blocks; no retain cycle). Repeated `refreshTitle`/`setupThreadHeader` calls only reassign the same synthesized channel.
- `WKSafeFilePreviewVC` scene-bounds logic checked in both show and dismiss paths, including the scene-less fallbacks.
- No iOS build was run for this review; findings come from byte-level code walking and standard UIKit hit-testing/gesture semantics.

yujiawei (changes requested):
# Code Review — PR #95 (octo-ios)

Reviewed at head SHA `e050ff2ba0f4b14ab5c8acd9260c7ee432c49c6c` against `main`. The diff bundles two logically independent changes: (1) the stated feature — tapping the "来自: XX群聊" subtitle in a thread jumps to the parent group; and (2) an unrelated `WKSafeFilePreviewVC` iPad scene-bounds fix.

## Verdict: Changes Requested

There is one blocking issue: the change introduces a behavior regression for the subtitle tap in **normal (non-thread) conversations**.

---

## P1 — Subtitle tap fallback is dead code; normal chats lose their existing tap-to-info behavior

**Files:**
- `Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/WKConversationChannelHeader.m:63-64` (gesture is now always attached)
- `Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/WKConversationChannelHeader.m:90-95` (`subtitlePressed`)
- `Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/WKConversationVC.m:406-410` (`onSubtitleTap` set unconditionally)

**Problem.** In `WKConversationVC`'s `channelHeader` getter, `onSubtitleTap` is assigned **unconditionally** for every conversation type:

```objc
[_channelHeader setOnSubtitleTap:^{
    if(weakSelf.parentGroupChannel) {
        [[WKApp shared] pushConversation:weakSelf.parentGroupChannel];
    }
}];
```

So `self.onSubtitleTap` is never nil. In `subtitlePressed`:

```objc
-(void) subtitlePressed {
    if(self.onSubtitleTap) {
        self.onSubtitleTap();      // always taken
    } else if(self.onInfo) {
        self.onInfo();             // never reached
    }
}
```

the `else if(self.onInfo)` fallback is unreachable. The header property comment states the opposite intent — *"为空时副标题点击行为与 onInfo 一致"* (when unset, subtitle tap behaves like `onInfo`) — but the property is never left unset, so that contract is not honored.

**Regression.** Before this PR, `subtitleLbl` had `userInteractionEnabled` at its UILabel default (`NO`), so taps on the subtitle passed through to the enclosing `infoBoxBtn` and triggered `infoPressed` → `onInfo` → opened conversation settings/details. This PR sets `subtitleLbl.userInteractionEnabled = YES` and attaches a tap gesture (`WKConversationChannelHeader.m:63-64`), so those taps are now captured by the label. For a normal group or person chat the subtitle shows member count / online status, `parentGroupChannel` is nil, and the block is a no-op — so **tapping the subtitle in normal chats now does nothing, where it previously opened settings.**

**Suggested fix.** Only wire `onSubtitleTap` where a parent-group jump is actually intended (i.e. in the thread branch of `setupThreadHeader`), leaving it nil otherwise so the `onInfo` fallback engages. For example, move the `setOnSubtitleTap:` assignment into `setupThreadHeader` (guarded on a resolved parent channel), or clear it for non-thread channels. That restores the documented fallback and keeps the header component's contract intact.

---

## P2 — Non-blocking notes

- **Thread subtitle text may stay generic when parent info isn't cached** (`WKConversationVC.m:304-318`, pre-existing). When `getChannelInfo:` returns nil, the code shows the placeholder `来自: 群聊` and fires `fetchChannelInfo:`, but the VC does not appear to refresh the subtitle when the parent group's info later arrives (its channel-update handling is oriented at `self.channel`, not `parentChannel`). The jump itself still works because `parentGroupChannel` is set regardless. Worth confirming the subtitle updates after the async fetch resolves; this is existing behavior, not introduced here, so it does not block this PR.

- **Unrelated change bundled in** (`WKSafeFilePreviewVC.m:64-131`). The switch from `UIScreen.mainScreen.bounds` to the scene's `coordinateSpace.bounds` for iPad Split View / Slide Over / Stage Manager is a reasonable and self-consistent fix (show and dismiss paths match, with a sensible `UIScreen` fallback). However it is unrelated to the stated subtitle-jump feature. Prefer landing unrelated fixes in their own PR so each change has its own reviewable scope and revert boundary. Not blocking, but flagging the scope creep.

- **Missing tests.** The PR checklist leaves "Tests added/updated" and "All tests pass locally" unchecked. Given the tap-routing branch (thread vs non-thread) is exactly where the P1 regression lives, a small test/manual matrix covering both cases would have caught it.

---

## Verification performed

- Confirmed `pushConversation:` exists and routes via `WKPOINT_CATEGORY_CONVERSATION_SHOW`, accepting `WK_GROUP` channels (`WKApp.m:1157`, `WKApp.m:1186`) — the jump target is valid.
- Confirmed the `substringToIndex:` call in `setupThreadHeader` is guarded by `if(separatorRange.location != NSNotFound)`, so it is not a crash risk.
- Confirmed `subtitleLbl` sits inside `infoBoxBtn`, which is why enabling its interaction diverts taps away from the button's `infoPressed`.

Please address the P1 regression before merge.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #95 (octo-ios)

## Summary

The PR gives the thread ("子区") conversation header a dedicated subtitle tap: `WKConversationChannelHeader` turns on `userInteractionEnabled` for `subtitleLbl`, attaches a `UITapGestureRecognizer`, and adds an `onSubtitleTap` callback that falls back to `onInfo` when unset. `WKConversationVC` captures the parent group in a new `parentGroupChannel` property inside `setupThreadHeader` and wires the callback to `pushConversation:`, so tapping "来自: XX群聊" opens the parent group instead of the thread's settings. A second, unrelated commit (`ebb85c7c`) fixes the safe-file-preview window to size itself from the window scene's `coordinateSpace.bounds` instead of `UIScreen.mainScreen.bounds` for iPad Split View / Slide Over / Stage Manager.

The thread jump itself is wired correctly and the parent channel cannot go stale, but the gesture is installed for every conversation type while the data behind the callback only exists for threads. The result is a dead tap where a working one used to be, which is the one blocker below.

## Verification

Static analysis only at head `e050ff2b`; build and tests not executed in this environment (no Xcode/iOS toolchain in the review sandbox).

- ✅ **Thread path really reaches the jump** — `refreshTitle` gates on `channelType == WK_COMMUNITY_TOPIC` (`WKConversationVC.m:284`) before calling `setupThreadHeader`, which assigns `parentGroupChannel` (`:309`); `setSubtitleText:` re-shows the label that `setChannelInfo:` hides for topics (`WKConversationChannelHeader.m:236`, `:356`), so the target is both visible and hit-testable.
- ✅ **No stale parent group** — every `WKConversationVC` is allocated fresh before `.channel` is assigned (`WKApp.m:1180`, `:1189`, `WKConversationRouter.m:57`); the reuse branches only call `locateToMessageSeq:`, so `parentGroupChannel` cannot leak across channels.
- ✅ **No retain cycle / no duplicate recognizers** — the block captures `weakSelf`, and `setupUI` runs only from `initWithFrame:` (`WKConversationChannelHeader.m:38`), so repeated `refreshTitle` calls cannot stack gestures.
- ✅ **Base contract confirmed** — `git show pr-base:` on the header has no `userInteractionEnabled`, no recognizer, and no `onSubtitleTap`; the subtitle tap used to fall through to `infoBoxBtn`.
- ❌ **Non-thread subtitle tap regressed** — see P1.
- No test in `Modules/WuKongBase/Example/Tests` covers the header or subtitle routing (grep for `ChannelHeader` / `subtitleText` returns nothing).

## Findings

One P1 blocker; two P2 items.

### P1 — Subtitle tap goes dead in every non-thread conversation (`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/WKConversationVC.m:406`)

`onSubtitleTap` is installed unconditionally in the lazy `channelHeader` getter, for every channel type, while `parentGroupChannel` is assigned only in `setupThreadHeader` — which `refreshTitle` calls solely for `WK_COMMUNITY_TOPIC` (`:284`, `:309`). At the same time `WKConversationChannelHeader.m:63` makes `subtitleLbl` hit-testable, so taps that previously fell through to `infoBoxBtn` → `infoPressed` → `onInfo` are now captured by `subtitlePressed` (`:90`). In group chats (subtitle `N个成员`, `:366`) and private/customer-service chats (online-status subtitle, `:258`) the block body evaluates `parentGroupChannel == nil` and does nothing, and the documented `else if (self.onInfo)` fallback (`:93`) is unreachable in production because the only consumer always sets the callback. Tapping the header subtitle therefore silently stops opening conversation settings — the opposite of the PR description's "普通群聊/私聊场景…行为不变".

```objc
// WKConversationVC.m:406 — arm only where the data exists
if (self.channel.channelType == WK_COMMUNITY_TOPIC) {
    [_channelHeader setOnSubtitleTap:^{
        WKChannel *parent = weakSelf.parentGroupChannel;
        if (parent) { [[WKApp shared] pushConversation:parent]; }
        else if (weakSelf.channelHeader.onInfo) { weakSelf.channelHeader.onInfo(); }
    }];
}
```

Smallest fix: leave `onSubtitleTap` nil for non-thread channels (or fall back to the `onInfo` action inside the block), and add a regression test asserting `subtitlePressed` invokes `onInfo` when only `onInfo` is set.

### P2 — Jump can open a blank conversation for an unresolvable parent (`Modules/WuKongBase/WuKongBase/Classes/Sections/Conversation/WKConversationVC.m:309`)

`[WKChannel groupWithChannelID:]` unconditionally returns a freshly allocated `WKChannel` (`Modules/WuKongIMiOSSDK/WuKongIMSDK/Classes/model/WKChannel.m:26`), so `parentGroupChannel` is non-nil for any thread id containing `____`, including dissolved groups or groups the user has left. The `if (weakSelf.parentGroupChannel)` guard at `:407` can therefore never reject an invalid parent, and the tap pushes a conversation whose `channelInfo` fails to load — a blank chat where the old behavior opened thread settings. The affordance is also live while the subtitle still shows the placeholder `来自: 群聊` pending `fetchChannelInfo:` (`:315`). Assign `parentGroupChannel` only inside the `parentInfo.displayName.length > 0` branch.

### P2 — Preview geometry still comes from an arbitrarily chosen scene (`Modules/WuKongBase/WuKongBase/Classes/Sections/Messages/WKSafeFilePreviewVC.m:71`)

The sizing fix is correct, but `scene` is still the first `UIWindowScene` in `connectedScenes` (`:57`), which on iPad multi-window is not necessarily the foreground one — the preview window can be attached to, and now also sized by, a background scene. The selection loop is pre-existing; this hunk increases reliance on it by deriving bounds from the same object. Prefer a scene with `activationState == UISceneActivationStateForegroundActive` before falling back.

## 数据流回溯

- `weakSelf.parentGroupChannel` (consumed `WKConversationVC.m:407`) ← assigned only at `:309` ← reachable only via `refreshTitle:285` under `channelType == WK_COMMUNITY_TOPIC` **and** `self.channelInfo != nil` (`:278`). For every other channel type the value is nil at tap time ⇒ **data does not flow**, block no-ops (root of the P1).
- subtitle tap → recognizer (`WKConversationChannelHeader.m:64`) → `subtitlePressed` (`:90`) → `onSubtitleTap` is non-nil for all VCs (`WKConversationVC.m:406`) ⇒ the `else if (self.onInfo)` branch (`:93`) is **dead in production**.
- `[WKChannel groupWithChannelID:groupNo]` (`:308`) ← `WKChannel.m:26` always returns non-nil ⇒ the guard at `:407` cannot express "parent unknown".
- `parentInfo` (`:310`) ← `WKChannelManager` local cache, may be nil on first open ⇒ subtitle shows the placeholder (`:315`) while the tap is already armed.
- `pushConversation:` (`:408`) ← `WKApp.m:1157` → `WKPOINT_CONVERSATION_SHOW_DEFAULT` (`WKApp.m:1186`) → always `[WKConversationVC new]` + push; `WKNavigationManager.m:70` has no dedupe ⇒ a second live instance of the parent group.
- `sceneBounds` (`WKSafeFilePreviewVC.m:71`, `:112`) ← first `UIWindowScene` of `connectedScenes` (`:57`), unfiltered by activation state.
- `setChannelInfo:` hides `subtitleLbl` and returns early for topics (`WKConversationChannelHeader.m:236`); `channelAvatarUpdate:` (`:76`) re-enters it without re-running `setupThreadHeader` ⇒ an avatar-update notification re-hides "来自: 群名" (and now the new jump affordance) until the next `refreshTitle`. Hunks untouched by this PR ⇒ pre-existing, noted only.

## 盲点 checklist (C1–C6)

- **C1 双路径 parity — hit.** The "arm" side is global (`WKConversationChannelHeader.m:63`, `WKConversationVC.m:406`) while the "data" side is thread-only (`:309`) — asymmetric, hence the P1. Checked the symmetric counterpart too: no unregister/remove path is needed, the recognizer lives exactly as long as the label, and `setupUI` runs once per header, so there is no leak or double-add.
- **C2 control-flow ordering / 嵌套复用 — clear.** `refreshTitle` ordering (`:280` channelInfo → `:281` memberCount → `:285` setupThreadHeader) leaves the thread subtitle visible with the right text; `setMemberCount:` skips topic channels (`WKConversationChannelHeader.m:367`) so it cannot overwrite it; re-entry only reassigns the same parent channel.
- **C3 授权边界 ≠ 能力边界 — clear.** No new endpoint, tool, or credential. `pushConversation:` is the same app-wide entry already reachable from search (`WKGlobalSearchVM.m:694`) and group lists (`WKMyGroupListVM.m:39`); message fetch for the parent group is still authorized server-side. The missing local membership check is raised as a robustness P2, not an auth bypass.
- **C4 授权生命周期 / 容器-成员级联 — N/A.** The diff reads and writes no permission, status, or membership field.
- **C5 build/note 通过 ≠ 运行期路径正确 — hit as a review constraint.** Nothing was compiled here; the P1 is derived from UIKit hit-testing plus gesture/`UIControl` cancellation semantics and verified against the base file, not from a build. The PR checklist ticks "compiles with no new warnings" but leaves tests unchecked, and no existing test covers this routing.
- **C6 治理/策略/安全文档自洽性 — N/A.** No docs, policy, CI, or workflow files in the diff.

## 跨轮 blocker 复检 (R6)

N/A — first review round for head `e050ff2ba0f4b14ab5c8acd9260c7ee432c49c6c`; no prior review blockers exist on this PR to re-check.

## 建议

1. Arm `onSubtitleTap` only for `WK_COMMUNITY_TOPIC`, or fall back to the `onInfo` action inside the block when `parentGroupChannel` is nil (fixes the P1 and keeps the documented fallback in `WKConversationChannelHeader.h:31` true).
2. Assign `parentGroupChannel` only when `parentInfo.displayName.length > 0`, so an unknown/dissolved parent cannot open a blank conversation.
3. Before pushing, scan `_rootNavigationController.viewControllers` for a `WKConversationVC` whose channel equals `parentGroupChannel` and pop to it (`WKNavigationManager.m:99` already exposes `popToViewControllerClass:`); a thread is normally opened from its parent group, so the stack otherwise becomes Group → Thread → Group, and each extra tap adds another live instance of the same channel.
4. Filter the preview scene by `UISceneActivationStateForegroundActive`.
5. Regression tests: header with only `onInfo` set ⇒ `subtitlePressed` invokes `onInfo`; header with both set ⇒ `onSubtitleTap` wins.

## 额外发现 (non-blocking)

- Nit — the newly tappable subtitle has no affordance: it keeps `navBarSubtitleColor` (`WKConversationChannelHeader.m:400`) and the recognizer gives no pressed/highlight feedback, so nothing signals that "来自: XX群聊" is a link while the title beside it is not.
- Nit — the bundled iPad preview fix (`ebb85c7c`) is absent from the PR body's What/Why/How/Testing, so QA and release notes will not cover iPad non-fullscreen preview; mention it or split it out.
- Note (pre-existing) — `channelAvatarUpdate:` can re-hide the thread subtitle and thus the new jump target until the next `refreshTitle`; untouched by this PR.

## Verdict: CHANGES_REQUESTED

The thread jump is correctly sourced and cannot go stale, but the tap gesture is global while its data is thread-only, which removes a working "tap subtitle → conversation settings" path from every group and private chat. Restricting the callback to thread channels (or restoring the `onInfo` fallback inside the block) clears the blocker; the two P2s and the nits are non-blocking.

[Octo-Q] verdict: REQUEST_CHANGES — 一个 P1（非子区会话副标题点击变成死点击，比修复前更糟，R1/R2 判 new），按 R4 有 P1 即 REQUEST_CHANGES；两个 P2 与 nits 非阻塞。

yujiawei:
### Follow-up notes (non-blocking, supplements the requested changes above)

Two additional P2 robustness items on the same subtitle-jump path — neither changes the verdict, but worth folding into the same fix pass:

1. **The `if(parentGroupChannel)` guard can never reject an invalid parent** (`WKConversationVC.m:407`). `[WKChannel groupWithChannelID:]` unconditionally returns a freshly allocated non-nil channel (`WKChannel.m:26`), so for any thread id containing `____` the guard is always true — including dissolved groups or groups the user has left. Tapping then pushes a conversation whose `channelInfo` fails to load (blank chat), where the old behavior opened thread settings. Suggest assigning `parentGroupChannel` only inside the `parentInfo.displayName.length > 0` branch, so an unresolved parent leaves the affordance inert.

2. **Preview scene is chosen unfiltered by activation state** (`WKSafeFilePreviewVC.m:57`). The sizing fix is correct, but `scene` is still the first `UIWindowScene` in `connectedScenes`, which on iPad multi-window is not necessarily the foreground one — so the preview window can be attached to and sized by a background scene. Prefer a scene with `activationState == UISceneActivationStateForegroundActive` before falling back. Pre-existing selection loop, but this change increases reliance on it.

Nit: the newly tappable subtitle has no visual affordance (same `navBarSubtitleColor`, no highlight), so nothing signals "来自: XX群聊" is a link.

Jerry-Xin (approved):
APPROVE — round-2 fix confirmation. The delta commit `dcf2330` fixes the round-1 blocker B-1 with a minimal, correctly gated change, and a full fresh-eyes re-walk of the cumulative diff found no new defects.

**Round-1 item status**

| Item | Status | Evidence |
| --- | --- | --- |
| 🔴 B-1: `onSubtitleTap` installed unconditionally in the header lazy getter, making the header's `else if(onInfo)` fallback dead and turning the subtitle into a silent dead zone in all non-thread conversations (group member-count / private online-status), breaking the settings entry | FIXED | Installation removed from the lazy getter and moved into `setupThreadHeader` inside the `____`-separator branch (WKConversationVC.m:305-316). `setupThreadHeader` runs only when `channelType == WK_COMMUNITY_TOPIC` (WKConversationVC.m:284-285). In group / private / customer-service conversations `onSubtitleTap` stays nil, so `subtitlePressed` falls through `else if(onInfo)` (WKConversationChannelHeader.m:90-94) to the same settings action `infoBoxBtn` fires — base behavior restored even though the label gesture itself remains unconditional (Header.m:63-64). Thread-from-group jump intact: callback pushes the parsed parent group via the existing `pushConversation:` route (WKConversationVC.m:311-314). Thread-from-private cannot show or fire the jump: `parentGroupChannel` is assigned only inside the `{groupNo}____{shortId}` branch (WKConversationVC.m:304-309) and the callback is guarded by `if(weakSelf.parentGroupChannel)`. |
| 🟡 ours: jump fires even when the parent group was deleted/left | UNCHANGED (acceptable) | Guard checks object presence, not group membership; consistent with the pre-existing global-search jump behavior, which routes through the same `pushConversation:` path. Non-blocking, same grade as round 1. |
| 🟡 ours: no accessibility trait or visual affordance on the tappable subtitle | UNCHANGED (acceptable) | Gesture still on a plain `UILabel` (WKConversationChannelHeader.m:63-64). Advisory: button trait + larger hit area. |
| 🟡 ours: unrelated bundled commit `ebb85c7` (iPad Split View preview fix) not covered by the PR description | UNCHANGED (acceptable) | Still in the cumulative diff (WKSafeFilePreviewVC.m); description covers only the header change. Advisory: mention in description or split. |
| 🟡 ours + mochashanyao #5208101036: no tests for subtitle tap + routing | UNCHANGED (acceptable) | Repo still has no unit-test tree, so this cannot be added in this PR. Advisory only. |
| 🟡 yujiawei #5208069289: thread subtitle placeholder may persist when `parentGroupChannel` info is not yet cached | UNCHANGED (acceptable) | Re-verified at this head: the `channelInfoUpdate:` delegate refreshes only when the updated channel equals the VC's own thread channel (WKConversationVC.m:530-531), so completion of the parent's `fetchChannelInfo` (WKConversationVC.m:322) does not re-run `refreshTitle`; the placeholder persists until a member update or VC rebuild. Non-blocking, same grade as round 1. |

**Fix verified as new code**

- Install order: `setOnSubtitleTap:` runs before `subtitleText` is assigned in the same branch; the gesture added in `setupUI` only dispatches, so there is no ordering hazard between callback installation and text setting.
- Re-entrancy: `refreshTitle` can fire repeatedly (member updates, channel-info load); each pass replaces the previous block via the `copy` property with a fresh `weakSelf` capture — no block stacking, and the VC → header → block → weak-self chain cannot retain the VC (no retain cycle).
- Delta between round heads touches only WKConversationVC.m (7+/6-); WKConversationChannelHeader.h/.m and WKSafeFilePreviewVC.m are byte-identical to the round-1 head, whose mechanisms were already verified there.

No new findings introduced by the fix.

Credits: round-1 reviewers yujiawei (#5208069289) and mochashanyao (#5208101036) — blocker convergence and the advisories tracked above.

---

**Supplementary verification notes (appended after initial post; verdict unchanged):**

- **yujiawei's follow-up comment (`5677862978`) re-verified at this head.** P2-1 — the tap-time `if(parentGroupChannel)` guard can never reject an unresolvable parent: confirmed, `groupWithChannelID:` (`WKChannel.m:26`) allocates unconditionally; the suggestion to assign `parentGroupChannel` only inside the resolved-`displayName` branch is a reasonable hardening, but the exposure matches the pre-existing search-jump path, so it stays advisory (same mechanism as row 2 above). P2-2 — preview scene chosen without an `activationState` filter: confirmed the selection loop is merge-base bytes (pre-existing); the sizing fix only increases reliance on it. Advisory, severity unchanged. The nit (no visual affordance) duplicates the accessibility row in the table.
- **No double-fire under the label gesture.** `subtitleLbl` is a subview of `infoBoxBtn`; with the label interactive, hit-test returns the label and the tap recognizer's default `cancelsTouchesInView` cancels the button's tracking on recognition — exactly one action fires per tap, never both.
- **No unresponsive window in the lifecycle.** Before `setupThreadHeader` runs — and on all non-thread paths permanently — `onSubtitleTap` is nil, so the fallback reproduces base behavior; in threads, `setChannelInfo:` hides the subtitle for `WK_COMMUNITY_TOPIC` and the same method that installs the callback re-shows it via `setSubtitleText:`, so a visible thread subtitle always has the jump callback installed.
- **Cache-timing refinement to the yujiawei row:** `requestLoadChannelInfoIfNeed` forces `needFetch = true` for `WK_COMMUNITY_TOPIC` on every VC entry, so the thread-channel info re-arrival re-runs `refreshTitle` → `setupThreadHeader`; the placeholder typically self-heals within the session once the parallel parent `fetchChannelInfo:` lands, without requiring a VC rebuild. Advisory as graded.

yujiawei (approved):
# Code Review — PR #95 (octo-ios)

Reviewed independently at head SHA `dcf2330dca2f8eb473b1e404c92b64bc7c5849d4` against merge-base `5b948cbe` on `main`. This head includes the follow-up fix commit `dcf2330d` ("修复非子区场景副标题点击变死区的回归"), which resolves the subtitle-tap regression flagged in the earlier review round at `e050ff2b`. This review re-validates the full diff at the current head, not the previous one.

## Verdict: Approve

The stated feature is implemented correctly, and the earlier blocking regression is genuinely fixed. Remaining items are non-blocking (robustness / scope / a11y / tests).

---

## Feature correctness — the thread → parent-group jump

Traced the full tap path at the current head:

- `WKConversationChannelHeader` enables `subtitleLbl.userInteractionEnabled` and attaches a `UITapGestureRecognizer` → `subtitlePressed` (`WKConversationChannelHeader.m:63-64`, `:90-95`).
- `subtitlePressed` invokes `onSubtitleTap` when set, otherwise falls back to `onInfo`.
- `onSubtitleTap` is now assigned **only** inside `setupThreadHeader` (`WKConversationVC.m:311`), which `refreshTitle` calls solely for `WK_COMMUNITY_TOPIC` channels (`WKConversationVC.m:284`). It is the only `setOnSubtitleTap:` call in the file; the lazy `channelHeader` getter sets only `onInfo`.
- `pushConversation:` routes via `WKPOINT_CATEGORY_CONVERSATION_SHOW`, which accepts `WK_GROUP` channels (`WKApp.m:1157`, `:1186`) — the jump target is valid.

Tapping the "来自: XX群聊" subtitle in a thread pushes the parent group; avatar/title/more taps still open thread settings via `infoBoxBtn` → `onInfo`. Matches the PR description.

## Earlier blocking regression — now fixed ✅

The prior review round (at `e050ff2b`) blocked on `onSubtitleTap` being installed unconditionally in the lazy getter, which made the `onInfo` fallback dead code and turned the subtitle into a dead zone in every normal group/private chat. At the current head that assignment has been moved into the thread-only `setupThreadHeader` path. For non-thread conversations `onSubtitleTap` is never set, so `subtitlePressed` reaches the `onInfo` fallback and normal chats keep their tap-to-settings behavior. Base-vs-head comparison confirms this restores the header component's documented contract ("为空时副标题点击行为与 onInfo 一致"). This resolves the sole blocker.

---

## Non-blocking findings

**N-1 (robustness) — jump can open an unresolvable parent.** `[WKChannel groupWithChannelID:]` always returns a non-nil channel (`WKChannel.m:26`), so `parentGroupChannel` is non-nil for any thread id containing `____`, including dissolved groups or groups the user has left. The `if(weakSelf.parentGroupChannel)` guard (`WKConversationVC.m:312`) can therefore never reject an invalid parent, and the tap is also live while the subtitle still shows the placeholder `来自: 群聊` pending `fetchChannelInfo:`. This mirrors the existing global-search group-jump behavior, so it is consistent with the app — but consider assigning `parentGroupChannel` (or arming the tap) only once `parentInfo.displayName.length > 0`, and/or a no-op/toast when the parent is still unresolved.

**N-2 (scope) — unrelated fix bundled in.** The diff also changes `WKSafeFilePreviewVC` to size the preview window from `scene.coordinateSpace.bounds` instead of `UIScreen.mainScreen.bounds` for iPad Split View / Slide Over / Stage Manager. The change itself is correct — `sceneBounds` is used symmetrically in the show and dismiss paths with a sensible `UIScreen` fallback — but it is unrelated to the subtitle-jump feature and is not mentioned in the PR's What/Why/How/Testing. Prefer landing unrelated fixes in their own PR for cleaner review/revert scope, and note it so QA and release notes cover iPad non-fullscreen preview.

**N-3 (robustness, pre-existing) — preview scene selection.** The preview still picks the first `UIWindowScene` in `connectedScenes` (`WKSafeFilePreviewVC.m:57`), unfiltered by activation state. On iPad multi-window this may not be the foreground scene, and this hunk now also derives bounds from it. Pre-existing; consider preferring a scene with `activationState == UISceneActivationStateForegroundActive`.

**N-4 (a11y / discoverability, nit).** The subtitle is now interactive but stays a plain `UILabel` with no accessibility trait/hint and no visual affordance, so nothing signals that "来自: XX群聊" is tappable (and that member-count/online-status subtitles are not).

**N-5 (tests, nit).** No automated coverage for the tap routing; the repo has no unit-test target for these classes and the checklist leaves tests unchecked. If test infra is added, cover both branches: thread subtitle → parent-group push, and normal-chat subtitle → conversation settings (the branch where the earlier regression lived).

---

## Verification performed

- Full walk of all 4 changed files (+39/−10) at head `dcf2330d`; diff confirmed against merge-base `5b948cbe`.
- Confirmed `setOnSubtitleTap:` is set only in `setupThreadHeader` (grep: single occurrence) and that the lazy getter sets only `onInfo`; verified `refreshTitle` gates `setupThreadHeader` on `WK_COMMUNITY_TOPIC`.
- Confirmed base file had no `userInteractionEnabled`/gesture/`onSubtitleTap` on the subtitle, so the fallback path is what preserves normal-chat behavior.
- Traced `pushConversation:` through the endpoint registry and `groupWithChannelID:` always-non-nil semantics.
- Memory management checked: blocks capture `weakSelf`; `setupUI` runs once per header (no duplicate recognizers); no retain cycle.
- No iOS build was run (no Xcode toolchain in this environment); findings are from code walking plus standard UIKit hit-testing/gesture semantics.

PR #97, opened 2026-09-17
fix(profile): 好友资料页发消息不再闪现个人主页

swr614:
## What

好友资料页点击"发消息"进入聊天页时，闪现一帧"我的个人主页"。

## Why

`WKUserInfoVC.sendBtnPressed` 在 push 聊天页之前会先调用 `popToRootViewControllerAnimated:NO` 清空共享导航栈。App 内只有一条全局共享的 `UINavigationController`，栈底是 `WKMainTabController`（消息/上下文/我的三个 tab）。清空到根这一步会瞬间露出当前选中 tab 的内容——如果用户此前停留在"我的" tab 再进通讯录发消息，就会先看到自己的个人主页闪一帧，再被聊天页滑入盖住。

## How

把"清空历史"的时机从 push **之前**挪到 push **之后**：

1. 正常 `pushViewController:conversationVC animated:YES`，聊天页照常滑入，视觉体验不变。
2. push 完成后，取 `self.navigationController.viewControllers.firstObject`（栈底根 tab 控制器），用 `setViewControllers:@[root, conversationVC] animated:NO` 把整条历史直接替换为 `[根tab, 聊天页]` 这两层，去掉中间的通讯录页、资料页。
3. 因为这次替换时数组最后一个元素（当前显示页）仍是同一个 `conversationVC`，`UINavigationController` 不会触发任何可见过渡，所以不产生闪烁；从聊天页返回依然直接回到主页 tab，行为跟改之前一致。

## Checklist

- [ ] All tests pass locally

## Testing

1. 通讯录点击好友进入资料页，点"发消息"，确认不再闪现"我的"页面。
2. 从聊天页点返回，确认直接回到主页/消息列表 tab（不会退回资料页或通讯录）。

## Breaking Changes

N/A


yujiawei (approved):
# Code Review — PR #97 (octo-ios)

## Summary

This fixes a one-frame flash of the user's own profile ("我的") when tapping **发消息** from a friend's profile page. The root cause is accurate: `sendBtnPressed` called `popToRootViewControllerAnimated:NO` *before* pushing the chat page. Because the whole app shares a single `UINavigationController` whose stack bottom is `WKMainTabController`, clearing to root synchronously exposed the currently-selected tab's content for one frame before the chat page slid in.

The fix reorders the two operations: push the chat page first (so it animates over the current profile page — the root tab is never revealed), then silently rewrite the stack to `[root, conversationVC]` with `animated:NO`.

## Spec compliance

The change does exactly what the PR description states and nothing more:
- **Missing**: none — the flash is addressed and back-navigation from chat still returns to the root tab.
- **Extra**: none — 6 lines added, 1 removed, single file (`WKUserInfoVC.m`).
- **Deviation**: none — final navigation stack is `[root, conversationVC]`, identical to the pre-change behavior (`popToRoot` + `push`), so back-navigation semantics are preserved.

## Correctness

I verified the key assumptions:

1. **Shared navigation controller** — the app uses one global `UINavigationController` (`WKNavigationManager`'s static `_rootNavigationController`). Since `WKUserInfoVC` is pushed onto it, `self.navigationController` resolves to that same instance, so the push (via `WKNavigationManager`) and the trim (via `self.navigationController`) operate on the same stack.
2. **Synchronous stack update** — `pushViewController:animated:YES` updates `viewControllers` synchronously even though the animation runs asynchronously, so `nav.viewControllers.firstObject` correctly returns the root tab controller at the point of the trim.
3. **No visible transition** — because the last (top) element of the new stack is still the same `conversationVC` that is animating in, `setViewControllers:animated:NO` does not trigger a visible transition, so no flash is introduced by the trim itself.
4. **Guards** — `if (root && root != conversationVC)` safely no-ops if the nav controller is unavailable and prevents a degenerate `[conversationVC, conversationVC]` stack.

Behavior parity with the original code is preserved: back from the chat page still lands on the root tab, skipping the contacts/profile intermediate pages.

## Notes (non-blocking)

- **P2 — stack mutation during an in-flight push transition.** Calling `setViewControllers:animated:NO` synchronously right after an `animated:YES` push mutates the stack while the push transition is being set up. In practice this is a well-known and generally safe technique when the top view controller is unchanged (as here), but it can occasionally surface UIKit console warnings on some OS versions. Please confirm the manual test shows a clean push with no "unbalanced begin/end appearance transitions" warnings in the console. A fully transition-safe alternative would be to perform the trim in a `CATransaction` completion block, but that is optional given the top VC is unchanged.
- **P2 — reference consistency (nit).** The push goes through `[WKNavigationManager shared]` while the trim uses `self.navigationController` directly. They point to the same instance, so this is purely stylistic; routing both through `WKNavigationManager` (or adding a helper there) would keep navigation access uniform.
- **No automated test** — the PR checklist notes tests were not added. For a navigation-timing/visual fix this is understandable and the manual test steps in the description are appropriate; not a blocker.

## Verdict

No P0/P1 issues. The fix is minimal, correct, and preserves existing navigation semantics.

Jerry-Xin (approved):
**APPROVE** — round 1 at head ed4de80 (exactly 1 commit, 1 file, +6/−1). Diff is 1:1 with the stated intent; zero smuggling.

## Change summary

`sendBtnPressed` (Modules/WuKongBase/WuKongBase/Classes/Sections/Common/WKUserInfoVC.m:722-732) previously called `popToRootViewControllerAnimated:NO` BEFORE pushing the conversation page. Since the app hosts everything on one shared root navigation controller, the synchronous pop-to-root exposed the currently selected tab (e.g. the "Me" homepage) for a frame before the push covered it — the reported flash. The PR reorders this: push `WKConversationVC` animated first (:725), then in the same main-thread turn replace the stack with `[root, conversationVC]` non-animated (:727-731), silently trimming the intermediate pages (contacts / profile).

## Independent verification

**Executable simulator probe (iOS 26.4).** I built a minimal UIKit app replicating the exact pattern — animated push followed by same-tick `setViewControllers:animated:NO` trim — inside a faithful replica of `WKRootNavigationController` (hidden system bar, interactive pop gesture disabled in the `pushViewController:` override, re-enabled in `navigationController:didShowViewController:`). Instrumented with transition-coordinator completion, per-VC appearance-callback logs, per-frame center-pixel sampling, and final-stack dumps. Results (two independent runs, unfiltered console capture):

- The push animation ran to completion: coordinator completion fired with `cancelled=false` and the full 0.350s duration — no cancellation, no truncation, no corruption.
- Appearance callbacks were strictly balanced: the profile page received exactly one `viewWillDisappear`/`viewDidDisappear` pair at transition end; the conversation page exactly one `viewWillAppear`/`viewDidAppear` pair; the root received no spurious re-appear; ZERO UIKit console diagnostics (no "Unbalanced calls", no nested-transition warnings, no assertions).
- Final stack `[root, conversation]` — identical to what the old pop-then-push produced, so back navigation (WKBaseVC.m:107-109, `popViewControllerAnimated`) and the `gestureRecognizerShouldBegin` single-VC guard behave exactly as before.
- Notable UIKit semantics surfaced by the probe: a non-animated `setViewControllers:` issued while a push transition is pending is DEFERRED by the system — a mid-transition read of `viewControllers` still shows the untrimmed stack, and the trim lands atomically with the transition's `didShow` callbacks. The mutation is coalesced by UIKit itself, which is why the running animation is untouched.

**Navigation identity invariant (split-brain census).** The push at :725 goes through `WKNavigationManager` (which forwards to the single static root nav controller, WKNavigationManager.m:74-76), while the trim at :727 uses `self.navigationController`. These are the same instance for every reachable presentation of this page: all three instantiation sites of `WKUserInfoVC` (WKBlacklistVC.m:126-128; WKApp.m:1757-1761 in the WKPOINT_USER_INFO handler; WKApp.m:1799-1803 in scan-to-add-friend) push it onto the shared root nav, and the class is never presented modally or hosted in a second navigation controller. WKContactsModule.m:11 imports the header but never instantiates the class (stale import, pre-existing).

**Guard correctness.** `if (root && root != conversationVC)` (:729) makes the trim a no-op in the degenerate nil-nav / empty-stack cases — unreachable per the census, but the right defense, and it correctly leaves a lone pushed conversation page in place.

**No adjacent-surface regression.** Full census of `popToRootViewControllerAnimated` callers (WKUserInfoVC.m:829, :851, :871 for delete-friend / blacklist / unblacklist; WKConversationPersonSettingVC.m:219, :281, :301; WKConversationGroupSettingVC.m:801; WKThreadSettingVC.m:437, :460): all are terminal "go home after destructive action" pops, animated, with no subsequent push — none share the pop-then-push flash pattern and none are touched by this change. The only non-animated pop-then-push site in the repo was the one this PR removes. House precedent for the same mechanism exists at WKGlobalSearchV2VC.m:392-398 (non-animated stack replacement against `self.navigationController` in a live flow). `sendBtnPressed` is wired from exactly one button (WKUserInfoVC.m:612); the video-call path uses an invoke block with no navigation and is unaffected.

**Double-tap / re-entrancy.** During the transition the source page is covered and touches are dropped, and the root nav's `pushViewController:` override disables the interactive pop gesture until `didShow`, so the untrimmed stack is never user-interactable; even a hypothetical second tap would self-heal to `[root, newest conversation]`.

**Cross-check with the earlier approval on this head.** The prior review's P2 note asked for confirmation that the mid-push trim produces no "unbalanced begin/end appearance transitions" console warnings; the two unfiltered probe runs above provide exactly that confirmation (zero UIKit diagnostics, strictly balanced callbacks) on the current SDK.

**CI at head ed4de80:** `build` (xcodebuild build-for-testing with an explicit iOS 26 SDK floor assertion) SUCCESS; secret-scan, osv-scan, dependency-review, history ancestor check, pr-title-lint, labeler all green. The `check-sprint` failure is the org reusable administrative gate (sprint association), unrelated to these bytes.

## Adjudication of the advisory second opinion

An advisory automated analysis pass (raw archived) returned REQUEST_CHANGES with one Critical: the same-tick `setViewControllers:` supposedly "does not wait for push completion" and could cancel/corrupt the animation or unbalance appearance callbacks; it recommended deferring the trim into the transition coordinator's completion. OVERRIDDEN on executable probe evidence: the probe shows UIKit itself defers the non-animated replacement until the pending transition completes (full-duration animation, `cancelled=false`, balanced callbacks, zero console diagnostics). The recommended explicit coordinator-completion deferral would be redundant and strictly weaker: `didShowViewController` re-enables the interactive pop gesture at transition end, and the ordering between that delegate callback and a coordinator completion block is unspecified, so explicit deferral could open a window where the untrimmed stack is swipeable back to the profile page — a state the old code never exposed. Same-tick trim lands atomically with gesture re-enable. The advisory's second point (no regression test) is accepted as non-blocking and folded into the advisories below.

## Advisories (non-blocking)

1. No automated regression test pins the final-stack behavior. The repo's CI only compiles (build-for-testing; no test-run job), and there is no app-hosted harness for navigation flows — adding one is infrastructure work beyond this fix's scope. A testable seam would be a tiny helper that performs push+trim given a nav controller and destination, unit-testable with a plain `UINavigationController` as the probe demonstrated.
2. The probe validated current-SDK (iOS 26.4 simulator) behavior; deferred-application semantics on older iOS versions were not probed. Worst case on a hypothetical older runtime where the trim applies immediately is cosmetic (animation snaps to its end state with the conversation page on top) — it cannot re-expose the root tab, since the trimmed stack's top view is unchanged. Not blocking.
3. WKContactsModule.m:11 stale `WKUserInfoVC.h` import (pre-existing; opportunistic cleanup candidate).

## Verdict

Mechanism verified by executable probe on the current SDK; final stack and back-navigation semantics byte-identical to the old flow; flash root cause (root exposure between pop and push) structurally eliminated; minimal scope with no adjacent-surface regression. APPROVE.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #97 (octo-ios)

## Summary

`sendBtnPressed` on the friend profile page used to clear the shared nav stack with `popToRootViewControllerAnimated:NO` before pushing `WKConversationVC`, exposing the root page for a frame — visibly the 个人主页 when the user arrived from the 我的 tab. This PR reverses the order: push the conversation first, then rewrite the stack to `@[root, conversationVC]` with `setViewControllers:animated:NO`, so the intermediate pages (通讯录 / 资料页) are dropped while the chat page is already on top. The fix is directionally right and the resulting stack is identical to the old one; my findings are about how the two mutations are wired, not about the intent.

## Verification

Static analysis only at head `ed4de804`; build and tests not executed in this environment. Paths are relative to `Modules/WuKongBase/WuKongBase/Classes/`.

- ✅ **Resulting stack matches pre-PR behaviour** — old pop-then-push gave `[root, conversation]`; the new push-then-trim gives `[root, conversationVC]` (`Sections/Common/WKUserInfoVC.m:726`). `root` is `viewControllers.firstObject`, the same VC the removed pop targeted (`Sections/Common/WKNavigationManager.m:50`).
- ✅ **Back affordance intact** — the custom back button is installed when stack depth ≥ 2 at `viewDidLoad` (`Sections/Common/WKBaseVC.m:56`), true before and after the trim, and `backPressed` pops through the manager (`:107`) → home, as before. `WKConversationVC.m` never reads the stack (only a push at `:467`), so nothing at the destination depends on the removed intermediates.
- ✅ **Dropped pages leak nothing** — the profile page adds its channel delegate in `viewDidLoad` (`:123`) and removes it in `dealloc` (`:147`), and `WKChannelManager` stores delegates in a weak `NSHashTable` (`Modules/WuKongIMiOSSDK/WuKongIMSDK/Classes/manager/WKChannelManager.m:490`), so releasing it after the trim instead of before can neither dangle nor leak. Intermediates already received their disappear callbacks when the profile page was pushed over them, so no lifecycle work is skipped.
- ✅ **`conversationVC.channel` is populated at tap time** — traced upstream to `viewModel.channelInfo`, filled by `loadPersonChannelInfo:`; the send button sits in `footerHeader`, which stays `hidden = YES` until `refreshData` sees a non-nil `channelInfo` (`:589`, `:163`), so the tap cannot fire with a nil channel.
- ✅ **No sibling path left unfixed** — `sendBtnPressed` has one call site (the send button, `:612`), and the removed line was the repo's only non-animated pop-then-push. Remaining `popToRootViewControllerAnimated:` calls (`:829`, `:851`, `:871`, `Sections/ConversationSetting/WKConversationPersonSettingVC.m:219`, `WKConversationGroupSettingVC.m:801`, `Sections/Thread/WKThreadSettingVC.m:437`) are deliberate animated go-home flows.
- ✅ **No new reachability or authz surface** — the trim drops exactly the pages the removed pop dropped and keeps `firstObject`; the external-member gate and the channel source are untouched.

## Findings

No P0/P1 issues; two non-blocking P2 items and one nit.

### P2 — Stack trim targets a different nav controller than the push (`Sections/Common/WKUserInfoVC.m:727`)

The push above goes through `WKNavigationManager`, which always targets the global `_rootNavigationController` (`WKNavigationManager.m:70`), while the trim reads `self.navigationController` — two sources of truth for one stack, with no fallback when they disagree. Today all three creation sites push through the manager (`WKApp.m:1757`, `:1799`, `Sections/Me/WKBlacklistVC.m:126`), so they are the same object and the fix works. But the repo already builds modal navs elsewhere (`Sections/ConversationList/Search/GlobalSearchV2/WKGlobalSearchV2VC.m:345`, `Sections/Messages/WKSafeFilePreviewVC.m:55`); if this page is ever hosted in one, `setViewControllers:` would move the already-pushed `conversationVC` out of the root nav into it, corrupting both stacks, while a nil `nav` silently skips the trim so Back from the chat returns to the profile page instead of home — a regression against the removed pop. The comparable stack surgery here guards exactly that (`WKGlobalSearchV2VC.m:390`). Fix: do both operations on the same controller — push through `self.navigationController` too, or add a manager helper that trims `_rootNavigationController` to `@[root, vc]` — and keep an else-branch fallback.

### P2 — Second stack mutation lands mid-transition (`Sections/Common/WKUserInfoVC.m:730`)

`pushViewController:animated:YES` starts a transition and this `setViewControllers:animated:NO` replaces the stack while it is still in flight, in the same run loop. That matters because `WKRootNavigationController` disables the interactive pop gesture inside its own `pushViewController:` override (`Sections/Common/WKRootNavigationController.m:92`) and re-enables it only from `navigationController:didShowViewController:` (`:76`); if the replacement cancels or short-circuits that transition, the callback may never be delivered for the push, leaving swipe-back disabled on the chat page — and turning the slide into a hard cut — until a later transition re-enables it. The new inline comment asserts the trim is visually inert during the animation, but that is a runtime claim no build or static check establishes, and CI is compile-only (`xcodebuild build-for-testing`, `.github/workflows/ci.yml:66`), so it is worth confirming on a device on the iOS 26 target this repo already special-cases. Every other stack rewrite here is a single mutation (`WKNavigationManager.m:74`, `WKGlobalSearchV2VC.m:390`). Fix: build the array first and mutate once — UIKit renders that as a push and it never trips the gesture-disable override.

```objc
UINavigationController *nav = self.navigationController;
UIViewController *root = nav.viewControllers.firstObject;
if (root) {
    [nav setViewControllers:@[root, conversationVC] animated:YES];
}
```

### Nit — Dead pointer comparison (`Sections/Common/WKUserInfoVC.m:729`)

`root != conversationVC` can never be false: `conversationVC` is allocated two lines above (`:723`) and was just pushed, so it cannot also be `viewControllers.firstObject`. The `root &&` half is the meaningful guard. Drop the comparison, or use `nav.viewControllers.count > 1` if the intent is "only trim when there is history to trim".

- **Dual-path parity** — hit, cleared: pop-then-push was unique to this call site; the new two-source divergence is the first P2. No guard/lint tool is claimed by the diff.
- **Control-flow ordering / reuse** — hit: the second P2. One call site, so no nested or re-entrant invocation compounds it.
- **Authorization boundary vs capability** — N/A: no permission, jail, tool-exposure, or credential change; the button's gate is untouched.
- **Authorization lifecycle / container cascade** — N/A: no auth state read or written.
- **Build/note pass ≠ runtime path correct** — hit: pure UIKit ordering that a compile-only CI gate cannot validate; the inline comment is an assertion, not a proof.
- **Governance / policy docs** — N/A: no docs, policy, or workflow files changed.
- **Security pass (STRIDE / OWASP)** — N/A across lenses: no auth, parsing, serialization, path, network, credential, or data-exposure surface touched; the added text is a comment with no user data; no new logging.
- **Distributed / multi-instance side effects** — N/A: client-side navigation only, no notification, push, emit, counter, or cache change. No `UIApplicationSceneManifest` in `Octo/Info.plist`, so no multi-window path can diverge the two nav references today.
- **Cross-round blocker recheck** — N/A: first round for this head; one commit in range (`ed4de804`), no prior unresolved blockers.

## Verdict: APPROVED

A correctly scoped 6-line fix: the resulting stack is what the old pop-then-push produced, the dropped pages release cleanly through a weak delegate table, and no consumer of the stack depends on the removed intermediates. Nothing makes a working path unusable or produces wrong user-visible data, so there is no P1 to block on. Both P2s are hardening — one source of truth for the stack, and a single mutation instead of push-plus-trim — and the same one-line reshape addresses both; they can land here or in a follow-up. Worth confirming on device: swipe-back gesture and push animation on a chat page entered this way.

